PR #7, opened 2015-02-23
Update builds to gatling 2.1.4

cprice404:
This PR does the following:

* Fixes all of our main code to run with Gatling 2.1.4
* Moves the original gatling sims to a directory called 'orig', since it seems like we should be getting rid of them and migrating them over to the main source tree now
* Archives the old ssl pem scripts and adds a new shell script that can be used to retrieve the agent certs for use with a real-live master; this means you don't have to manually hack the master certs anymore.


cprice404:
@doug-rosser @mwbutcher I was doing some gatling stuff on Friday and decided to try to just go ahead and get everything updated to the latest gatling version.  I think this PR gets us there.  I'm not sure if we have any jenkins jobs pointing to the master branch at this point, and if so, this might break them.  If that's the case maybe I can help tweak things until we're good?  If not, would be good to go ahead and get this updated.

PR #8, opened 2015-03-04
Unique report timestamps

cprice404:
There are a few commits in this PR.  The first one gets rid of some 'resource' stuff in the recorded simulation.  My impression is that gatling tried to auto-detect cases where a browser might be able to load some resources for a web page in parallel; I should have unchecked that checkbox when I made the recording, because a puppet agent will never behave that way.  In the first commit I'm just manually fixing up the simulation to look like what it would have looked like without that checkbox checked, to ensure that gatling doesn't try to execute the requests for a single agent run in parallel.

The second commit just cleans up some comments in the build.sbt file.

The third commit is the meat of this PR; it changes the simulation so that we can inject dynamic values for the 'timestamp' field in the body report request.

This commit changes the CloseWait simulation to do the following:
    
    * Inject a user session variable called 'reportTimestamp' for each
      agent run, set to the current time.
    * Switch the report request to use a dynamic body (via ELFileBody)
      instead of a static body (RawFileBody), so that we can inject
      dynamic values into the report payload.
    * Modify the text file that contains the report body to use
      a variable interpolation, so that we get the value of the
      user session variable 'reportTimestamp' for the report,
      rather than a hard-coded timestamp.
    
Without these changes, PuppetDB would reject all the report submissions
after the very first one, because they violated PuppetDB's uniqueness
constraints for reports.


cprice404:
@mwbutcher would you mind reviewing this when you get a chance?

@camlow325 you might be interested in the third commit here, since it shows how to solve the problem with the uniqueness constraints in PuppetDB reports.  I think the example illustrated there should be pretty easy to generalize across other situations we may encounter when we want to do testing in PE environments.

cprice404:
@mwbutcher does this look OK to you?

PR #10, opened 2015-04-14
Add CatalogZero simulation for FOSS 3.7.5

cprice404:


simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS375CatalogZero.scala
```diff
@@ -0,0 +1,277 @@
...
+import io.gatling.http.Predef._
+import io.gatling.jdbc.Predef._
+
+import com.puppetlabs.gatling.runner.SimulationWithScenario
+import org.joda.time.LocalDateTime
+import org.joda.time.format.ISODateTimeFormat
+
+class FOSS375CatalogZero extends SimulationWithScenario {
+
+	val httpProtocol = http
```
camlow325: In some of the other sims, I've seen `httpProtocol`, `uri1`, and the `setup` call at the end of this commented out.  Do we want that here too?

camlow325:
The "${reportTimestamp}" change in the report request.txt file looks different here than what I've seen in some of the previous simulations.  I was expecting to see the "time" element appear at the top-level of the report document - same level as "host" and "configuration_version".  Instead, this one has the change on one of the log entries.  Maybe it doesn't matter as long as the content is "different" somehow?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS375CatalogZero.scala
cprice404: I don't *think* these matter, as I think that our subclass and driver override them anyway.  (I think it would be pretty easy to tell if they didn't, because the behavior of the sim would be very different... but it's been a long time since I looked into that code.)

cprice404:
I think you're right about the report timestamp thing, I think I hosed that... and that it just didn't matter for me because the only report processor in use by OSS by default is 'store', and it doesn't care if the reports are dupes.  What I have here would probably fail w/PuppetDB... will confirm tomorrow and fix it.

cprice404:
@camlow325 I added a commit that fixes that timestamp thing.

camlow325:
I'll just assume my comment that the need to comment out `httpProtocol`, `uri1`, and `setup` is irrelevant.

:+1:

cprice404:
Sorry, I missed that one when I was replying.

Commenting those out is not necessary, our driver code overrides them anyway.  Commenting them out might be worthwhile for clarity, but it also adds an extra step to the list of instructions for patching a raw recording... so I could go either way in terms of what we decide is our best practice going forward.  Happy to discuss further.

PR #14, opened 2015-04-23
(MAINT) update PE38 catzero to underscores

cprice404:
Three commits here:

1. fix some broken quotes in the PE3.7.2 catzero sim
2. clean up some of the code in the retrieve-certs shell script, and add another shell script that can be used if you're doing local runs with a local puppet environment
3. Update PE3.8 catzero sim to use underscores instead of dashes, to match up with current HEAD of test-catalogs repo.


simulation-runner/initialize-ssl-from-local-certs.sh
```diff
@@ -0,0 +1,62 @@
...
+mkdir -p ./target/ssl
+
+cp "${SSL_DIR}/certs/${CERTNAME}.pem" ./target/ssl/hostcert.pem
+cp "${SSL_DIR}/private_keys/${CERTNAME}.pem" ./target/ssl/hostkey.pem
+cp "${SSL_DIR}/certs/ca.pem" ./target/ssl/cacert.pem
+
+echo "Copied files.  Generating keystore file for gatling."
+
+cat ./target/ssl/hostcert.pem ./target/ssl/hostkey.pem > ./target/ssl/keystore.pem
+echo "puppet" | openssl pkcs12 -export -in ./target/ssl/keystore.pem -out ./target/ssl/keystore.p12 -name ${CERTNAME} -passout fd:0
```
camlow325: Might be nice to have the password defined as a local variable somewhere and referenced in all of the commands below rather than having the string "puppet' appear everywhere.

simulation-runner/retrieve-agent-ssl-certs.sh
```diff
@@ -73,9 +49,6 @@ scp root@${PE_AGENT}:${HOST_KEY} ./target/ssl/hostkey.pem
 scp root@${PE_AGENT}:${CA_CERT} ./target/ssl/cacert.pem
```
camlow325: Not related to these changes, but I got an error message initially when running this against a Puppet Server 2 install:

~~~~
Copying files from jb-centos7
bash: puppet: command not found
~~~~

I believe the Puppet bin directory isn't put into the root user's shell PATH by default with AIO - not sure if that's intentional?  In any event, if I added the following to my `/root/.bashrc` on my target agent host before running the script...

~~~~bash
export PATH=/opt/puppetlabs/puppet/bin:$PATH
~~~~

... then the script worked okay.

Maybe we should document this setup step if not automate it into the script at some point?

camlow325:
:+1: I didn't test the new sim files out locally.  The changes seem reasonable, though, so I assume they will work fine.  My other comments are pretty cosmetic so I figure they could be covered in a separate PR later if @cprice404 cares to do so.

simulation-runner/initialize-ssl-from-local-certs.sh
cprice404: that's legit

simulation-runner/retrieve-agent-ssl-certs.sh
cprice404: Yeah, let's add that to the script next time we run into it.

cprice404: (Assuming we confirm that's intentional in AIO)

PR #18, opened 2015-05-04
Server 566 and 580 apply manifests clone repo

jpinsonault:
I through both tickets into one PR because they're so small, and related.

If you have a centos-6 box with puppet installed, you can test it with

```bash
beaker --log-level debug --hosts dev/beaker/target_machine.yml --tests dev/beaker/
```
You'll have to put the hostname for your centos box into `target_machine.yml` first

### DIFF
--- a/dev/beaker/apply_manifests.rb
+++ b/dev/beaker/apply_manifests.rb
@@ -2,15 +2,18 @@ test_name "Apply manifests for jenkins/git, sbt, and jjb"
 
 manifests = Dir.glob("./dev/manifests/*.pp")
 
-hosts.each do |host|
-  manifests.each do |local_path|
-    manifest_name = File.basename(local_path)
-    remote_path = File.join("/tmp/", manifest_name)
+step "Make a temp dir for manifests"
+remote_temp_dir = create_tmpdir_on(dev_machine)
 
-    step "Send #{manifest_name} to host"
-    scp_to(host, local_path, remote_path)
+manifests.each do |local_path|
+  manifest_name = File.basename(local_path)
+  remote_path = File.join(remote_temp_dir, manifest_name)
 
-    step "Apply #{manifest_name} manifest"
-    on(host, "puppet apply #{remote_path}")
-  end
+  step "Send #{manifest_name} to #{dev_machine}"
+  scp_to(dev_machine, local_path, remote_path)
+
+  step "Apply #{manifest_name} manifest"
+  on(dev_machine, "puppet apply #{remote_path}")
+  
 end
+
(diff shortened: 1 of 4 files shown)


dev/beaker/apply_manifests.rb
```diff
@@ -0,0 +1,16 @@
+test_name "Apply manifests for jenkins/git, sbt, and jjb"
+
+manifests = Dir.glob("./dev/manifests/*.pp")
+
+hosts.each do |host|
+  manifests.each do |local_path|
+    manifest_name = File.basename(local_path)
+    remote_path = File.join("/tmp/", manifest_name)
```
cprice404: I think maybe there's a method on the `Host` class in beaker where you can ask it for a temp file path.  That way you don't have to assume `/tmp/` exists, and it works with Windows, etc.  Not that I really care about Windows, but might as well use the idiomatic approach to doing this if possible.

dev/beaker/clone_gatling-puppet-load-test.rb
```diff
@@ -0,0 +1,9 @@
+test_name "Clone the gatling-puppet-load-test repo on the target machine"
+
+repo = "gatling-puppet-load-test"
+repo_url = "https://github.com/puppetlabs/#{repo}.git"
+
+hosts.each do |host|
```
cprice404: Technically we don't want to do this on all of the hosts, right?  I know that for now we only have one machine in the hosts file, so I'll probably be fine merging this as-is, just wondering if you have any thoughts on what this might look like if we did introduce multiple hosts (or if this beaker script were to end up getting used as part of the production environment, rather than just the dev env).

cprice404:
+1 other than a few minor questions

dev/beaker/clone_gatling-puppet-load-test.rb
jpinsonault: Yes that's totally right and it should probably *not* work with multiple hosts, unless we can think of a reason it should be. I asked the QA folks and apparently you can define a custom role and use it as a unique identifier for the host. Something like
```yaml
roles:
      - dev_machine
```

```ruby
on(dev_machine, "echo hello")
```
This seems like the way to go, since it will be more explicit. Either that or something like `host = hosts[0]`

dev/beaker/apply_manifests.rb
jpinsonault: Good call, I'll add that in

dev/beaker/clone_gatling-puppet-load-test.rb
jpinsonault: But to answer your second question, it should work fine on multiple hosts, as far as I can tell. It would just create two development boxes with jenkins etc.

jpinsonault:
I went ahead and moved it away from using `hosts.each`
Switched it to using `create_tmpdir_on`

PR #19, opened 2015-05-05
(SERVER-568) Docs for setting up dev env

jpinsonault:


jeffmccune:
Looks good, only comment is to wrap at 80 chars please.

dev/README.md
```diff
@@ -0,0 +1,32 @@
...
+  * `puppet` installed and in the `PATH`
+
+This will be your jenkins/gatling box 
+
+To actually run gatling against a SUT, you'll need to set up another machine with some flavor of puppetserver/PE depending on what you're trying to test. Running tests isn't covered in this guide though.
+
+It's also important to note that if you want to work on the cobbler provisioning portion of the automation, it's much more involved. There will be a separate guide for that at some point.
+
+## Setting up the jenkins/gatling box
+To run the beaker script, you must first edit `beaker/target_machine.yml` and replace `target_hostname` with the hostname or IP of your system.
```
fpringvaldsen: I think it might be better to change this path to `dev/beaker/target_machine.yml`, since the beaker directory isn't at the root of the repo.

fpringvaldsen:
This looks good to me as well! I agree with @jeffmccune that we should wrap it at 80 characters.

Also, following this guide, I ran into an issue wherein puppet has the following error:

`Error: Could not find resource 'Class[Jenkins::Firewall]' for relationship from 'Class[Jenkins::Service]' on node`.

@jpinsonault Have you seen this error before?

dev/README.md
cprice404: +1.

Also, might change `target_hostname` in that yaml file to something that might serve as a reasonable default, e.g. `jenkins-gatling`, and then suggest that devs add an entry to their /etc/hosts file (maybe suggest modifying the yml file as a backup strategy, but suggest /etc/hosts first).  This way you're not encouraging people to make user-specific changes to files that live in VCS, because once they do that they have to be careful not to commit their changes.

dev/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+`cd` to the root of the repo, and run.
+```
+beaker --log-level debug --hosts dev/beaker/target_machine.yml --tests dev/beaker/
```
cprice404: This presumes that the user already has `beaker` set up and on the path.

I'd recommend, instead, adding a Gemfile to the `dev` directory, and documenting a set of commands to run along the lines of `bundle install`/`bundle exec beaker`.

dev/README.md
```diff
@@ -0,0 +1,32 @@
+# Setting up a Development Environment
+This describes how to set up a development environment suitable for working on the automation for `gatling-puppet-load-test`
+
+## What you get
+The beaker scripts in the `beaker` directory will set up a machine with a jenkins server, a copy of this repo (which contains the gatling executables), `sbt` for running gatling, and an installation of `jenkins-job-builder` configured to point at the jenkins server.
+
+## Requirements
+This guide assumes you have a centos-6 machine/VM with:
+  * ssh key authentication for the root user (for `beaker`)
+  * `puppet` installed and in the `PATH`
```
cprice404: Could we not handle this automatically via beaker?  Don't need to do this as part of this PR, just maybe something to create another ticket for.

cprice404: Also might add that having the time set properly on the VM is important.  (Or, create a ticket about adding ntpd to the beaker setup steps or something.)  I hit an issue where the PMT failed to install some puppet modules due to bad timestamp on the VM.

dev/README.md
cprice404: Also, I accidentally had a pretty old version of beaker when I ran this, and it didn't have the `create_tmpdir_on` method, so I got an error.  Hopefull Gemfile/bundle exec would fix this.

cprice404: The manifests don't appear to be idempotent.  My first run failed due to the `create_tempdir_on` method missing from my version of beaker; when I upgraded to a new `beaker` and tried to run again, I got an error about `git clone` trying to clone into a directory that already exists.

This doesn't need to be fixed for this PR but let's capture it in a ticket; basically just "running the beaker setup command multiple times shouldn't cause an error".

dev/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Setting up the jenkins/gatling box
+To run the beaker script, you must first edit `beaker/target_machine.yml` and replace `target_hostname` with the hostname or IP of your system.
+
+`cd` to the root of the repo, and run.
+```
+beaker --log-level debug --hosts dev/beaker/target_machine.yml --tests dev/beaker/
+```
+
+If everything goes well, the beaker output should show no errors.
```
cprice404: This was true for me after dealing with the issues above, but...

dev/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Jenkins should be available on port `8080` of your machine.
```
cprice404: This was not.  Nothing running on 8080 for me.  Here's a gist of my beaker output:

https://gist.github.com/cprice404/23668d82e276ae95000b

If I had to guess what went wrong, it looks to me like during the 'apply manifests' stage, it didn't apply enough things.

Not sure if this matters, but I changed the hostname to `jenkins-gatling` in the `target_machine.yml` file.

cprice404: Ah, I see what's wrong.  I was running my beaker command from inside the `dev` directory (because I created a `Gemfile` there to deal with installing beaker for me).  The `apply_manifests.rb` has a line in it where it does a glob on `./dev/manifests/*.pp`, which was basically a no-op for me since that path didn't exist.

It'd probably be better to list out the manifests directly in an array literal, so that it will fail if any are not found.  I wouldn't expect those to change often enough to where globbing them will save us a ton of effort.  (You can create a separate ticket for that, though.)

Also if you do put relative paths into the code I think you should probably have devs run their commands with `./dev` as their current working directory (because that will be the most sensible place to put a Gemfile), so paths should be relative from there.

dev/README.md
cprice404: This has bitten me like 4 times now (which, admittedly, is because I'm an idiot).  Would be really nice to have the setup scripts update the timestamp on the VM.

dev/README.md
cprice404: After resolving that ^^ issue, I hit the error that @fpringvaldsen mentioned:

```
Error: Could not find resource 'Class[Jenkins::Firewall]' for relationship from 'Class[Jenkins::Service]' on node centos6-64.localdomain
```

cprice404:
@jpinsonault looking good, here's what I think should be next steps and then we can get this merged and close out the current crop of tickets:

1. Add a `Gemfile` in `dev`, update docs to show how to use it to run beaker w/bundle exec
2. Fix glob paths in `apply_manifest.rb` so that they will work if you run from `dev` since that's where the `Gemfile` will be.
3. Figure out what's up with that `Jenkins::Firewall` issue that @fpringvaldsen and I hit.
4. Change `target_hostname` to something like `jenkins-gatling` and suggest users put an entry for it in `/etc/hosts`.

At that point I'd be happy to merge this PR and close out the current tickets.  A few new tickets that we should probably create:

1. Install Puppet on the jenkins box rather than having it as a prereq
2. Fix manifests to be idempotent (or at least, able to be run more than once w/o erroring).
3. Update timestamp on VM (by exec-ing ntpdate, installing ntpd, or whatever other means seems appropriate)

Looking good, thanks!

dev/README.md
jpinsonault: @fpringvaldsen Good idea
@cprice404 You're right, that makes more sense

jpinsonault:
@cprice404 Thanks for all the comments! I'll get started on your list.
@fpringvaldsen I don't think I ever ran into that firewall error, I'll investigate

jpinsonault:
@fpringvaldsen @cprice404 I addressed your feedback:
* Added Gemfile with beaker dependency
* apply_manifest.rb doesn't use glob
* All paths relative to `dev/`
* Added firewall module to fix error both of you were having (I think I installed it manually while messing around with the jenkins plugin and forgot it was needed)
* Suggests adding entry to hosts file in the readme
  * Renamed host is beaker config file
* Doesn't error if repo already exists
* Updates time with ntpdate before running

@cprice404 If that looks good the only ticket we'll have to make it for installing puppet

dev/README.md
```diff
@@ -0,0 +1,49 @@
...
+	--log-level debug \
+	--hosts beaker/target_machine.yml \
+	--tests beaker/
+```
+
+If everything goes well, the beaker output should show no errors.
+
+Jenkins should be available on port `8080` of your machine.
+
+`jenkins-jobs` and `sbt` should be available at the command line.
```
fpringvaldsen: Following this guide, beaker worked correctly and Jenkins is available on port 8080, and the copy of the repo lives in the right place, but `jenkins-jobs` and `sbt` don't seem to be available on the command line.

jpinsonault: Hmm, I'll investigate, do you have the beaker output?

jpinsonault:
@fpringvaldsen I made some changes to the jjb and sbt manifests to make them more general/actually work right. Tested it in the vmpooler.

Also added a beaker assert to make sure the errors in your puppet output are actually caught instead of ignored silently.

@cprice404 @fpringvaldsen  Also threw in SERVER-645, having beaker install puppet. This happens as the second step, after syncing the time

dev/manifests/setup_sbt.pp
```diff
@@ -1,10 +1,10 @@
...
+  # Command from the sbt getting started page
+  command => 'curl https://bintray.com/sbt/rpm/rpm | tee /etc/yum.repos.d/bintray-sbt-rpm.repo',
+  path    => '/usr/bin:/usr/sbin:/bin:/usr/local/bin',
 }
 
 package { 'sbt':
-  ensure => installed,
-  require => Yumrepo["sbt yum repo"],
+  ensure  => installed,
+  require => Exec["install_sbt_repo"],
```
nwolfe: Just to impart some Puppet knowledge, if both resources are in the same manifest file you can use the `->` operator instead of the long-winded `require => ...`.

For example:
```puppet
exec { 'install_sbt_repo': 
...
}
->
package { 'sbt':
...
}
```

dev/README.md
KevinCorcoran: it worked for me!

dev/README.md
KevinCorcoran: `jenkins-jobs` and `sbt` are there on my test VM.

dev/manifests/setup_sbt.pp
jpinsonault: Ahh, fancy. Thanks @nwolfe!

dev/beaker/10_install_puppet.rb
```diff
@@ -0,0 +1,6 @@
+test_name "Install puppet on target machine"
+
+puppet_version = "3.7.5"
```
KevinCorcoran: Any reason not to use a 4.x version of puppet here?

KevinCorcoran:
:tada: This looks good, @jpinsonault.  Just had the couple questions about puppet version, and the discussion we had offline about clarifying the docs about re: which machine things are expected to be run on.  The beaker stuff worked correctly for me the first time.

fpringvaldsen:
Everything is working for me now, and the most recent changes look good, so I'm :+1: once @KevinCorcoran's feedback is addressed.

dev/beaker/10_install_puppet.rb
jpinsonault: No, I honestly forgot I left it as 3.7. It worked and I said "cool :+1:" and kept going. I can change it to puppet 4 if you'd like.

KevinCorcoran: No strong opinion, it just seems like we might as well use the latest.

jpinsonault:
@KevinCorcoran Added a line mentioning that this doc is intended to be run from a separate machine, and beaker now installs the latest puppet

PR #22, opened 2015-05-25
(MAINT) Fix foss 2.1.0 sim, improve cert script

cprice404:

* Adds an explicit path to the puppet executable for the
  `retrieve-agent-ssl-certs.sh` script, so that it will work with
  puppet 4.0+ out of the box.  (Backward incompatible with Puppet 3,
  but we should be testing that more and more rarely in the future.)
* Fixes a few bugs/typos in the "Foss PuppetServer 2.1.0" sim

### DIFF
--- a/simulation-runner/config/scenarios/foss-puppetserver210-catalogzero-1.json
+++ b/simulation-runner/config/scenarios/foss-puppetserver210-catalogzero-1.json
@@ -1,5 +1,5 @@
 {
-    "run_description": "puppeserver 2.1.0 Catalog Zero - 1000 agents, 30 minute ramp, 2 times",
+    "run_description": "puppeserver 2.1.0 Catalog Zero - 1 agent, 30 minute ramp, 2 times",
     "nodes": [
         {
             "node_config": "foss-puppetserver210-catalogzero.json",
--- a/simulation-runner/config/scenarios/foss-puppetserver210-catalogzero-1.json
+++ b/simulation-runner/config/scenarios/foss-puppetserver210-catalogzero-1.json
@@ -1,0 +1,13 @@
+{
+    "run_description": "puppeserver 2.1.0 Catalog Zero - 1000 agents, 30 minute ramp, 2 times",
+    "nodes": [
+        {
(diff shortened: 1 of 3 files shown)


simulation-runner/config/scenarios/foss-puppetserver210-catalogzero-1.json
```diff
@@ -0,0 +1,13 @@
+{
+    "run_description": "puppeserver 2.1.0 Catalog Zero - 1000 agents, 30 minute ramp, 2 times",
```
nwolfe: Looks like the other scenarios are keeping the descriptions up to date with the configurations...we might as well correct this one too.  

Something like `1 agent, 2 times`?

nwolfe:
:+1: other than updating the scenario description

erikPrime:
I'm a little concerned about making the retrieve-agent-ssl-certs.sh incompatible with puppet 3, particularly given the news that we have a new release based on puppet 3 coming soon.  Could we keep the old one for a while yet and perhaps rename it to something obvious?

nwolfe:
> I'm a little concerned about making the retrieve-agent-ssl-certs.sh incompatible with puppet 3, particularly given the news that we have a new release based on puppet 3 coming soon.

Considering this script is called by a human, seems like we could make it support an optional argument that informs it where to find the `puppet` executable pretty easily too.

cprice404:
@nwolfe : fixed the description
@erikPrime if it's only going to work out-of-the-box with one of the two (which is the case right now), I'd prefer it to be the newer one.  Should be fairly easy to hack it by hand for any 3.x testing we need to do for now, right?

I'd also be happy for someone to file a PR that supports an option, like @nwolfe suggested, but that seems like a low-ish priority to me.

@jpinsonault I didn't realize this hadn't been merged yet... I would think that if you are doing a 2.1.0 run without this, you'd be having problems because of the invalid path to the request body file for the report (i.e. all of your reports should be failing on that run... shouldn't they?)

nwolfe:
:+1: going to merge this.

@erikPrime should be trivial to update the script if/when it's necessary.

PR #23, opened 2015-05-29
(MAINT) Use stahnma-epel module

nwolfe:
This commit removes the manual EPEL setup and instead just relies on the
stahnma-epel module. The result is less code, and more supported platforms.


nwolfe:
@jpinsonault If you're good with this change let's merge it in!

jpinsonault:
:+1:

PR #25, opened 2015-06-02
Server 329 script to install puppet on sut

jpinsonault:


jenkins-integration/beaker/install/foss/40_add_puppet_bin_to_path.rb
```diff
@@ -0,0 +1,6 @@
+test_name "Add puppet binaries to the $PATH"
+
+hosts.each do |host|
+  on(host, 'echo "export PATH=\$PATH:/opt/puppetlabs/puppet/bin" >> ~/.bashrc')
+  on(host, 'echo "export PATH=\$PATH:/opt/puppetlabs/server/bin" >> ~/.bashrc')
```
nwolfe: Do we symlink all the binaries into `/opt/puppetlabs/bin/`?  If so we could just add that path

jenkins-integration/scripts/foss_install.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
nwolfe: This file should be executable (the `pe_install.sh` is)

jenkins-integration/scripts/foss_install.sh
```diff
@@ -0,0 +1,17 @@
...
+  --helper beaker/helper.rb    \
+  --tests beaker/install/foss/ \
+  --preserve-hosts             \
+  --no-color                   \
+  --xml                        \
+  --ntp                        \
+  --root-keys                  \
+  --repo-proxy                 \
+  --debug                      \
+  --keyfile $SSH_KEYFILE
```
nwolfe: https://github.com/puppetlabs/puppet-server/blob/master/acceptance/scripts/generic/testrun-full.sh

This is a good pattern to follow for writing these scripts. We should do the `--keyfile` part here and hardcode it

jenkins-integration/beaker/install/foss/40_add_puppet_bin_to_path.rb
jpinsonault: Yes, we do. Good call

jenkins-integration/scripts/foss_install.sh
jpinsonault: I like it, I'll convert it over

erikPrime: We need some additional work here:
1.  We need to pass the name of an existing SUT in to the script.  Something like "perf-bl15"
2.  We need plug that host name into the host.yaml file, or into beaker somewhere.
3.  We need to pass --no-provision

harder stuff.
I think we should CONSIDER using a git submodule to pull in the helper.rb instead of using copy pasta.  BUT, maybe not because git submodule can be scary bad.

jpinsonault:
@erikPrime @nwolfe 
Addressed your feedback,
Updated gitignore for bundler
Converted beaker running scripts to a nicer format
Assume key file will be ~/.ssh/id_rsa-acceptance
Made foss_install.sh executable
Add /opt/puppetlabs/bin to path
Used cleaner on(hosts) instead of foreach loop where possible

Looking at @erikPrime's latest comment now

jenkins-integration/scripts/foss_install.sh
nwolfe: @erikPrime Good points.  These scripts are far from done, and we're re-building the automation around them.  I imagine we will continue to generalize them as necessary so they can be called by other scripts and various information passed in.

For now, we're running them by hand, but soon will be writing the scripts that configure all the various variables and then call these.

> I think we should CONSIDER using a git submodule to pull in the helper.rb instead of using copy pasta.

Our `helper.rb` will be unique - despite the name being the same in all the projects, they don't actually contain the same content.  I believe the naming is allude to the fact that it's the file you pass to the Beaker `--helper` argument.

jenkins-integration/beaker/install/foss/40_add_puppet_bin_to_path.rb
```diff
@@ -0,0 +1,5 @@
+test_name "Add puppet binaries to the $PATH"
+
+# hosts.each do |host|
+  on(hosts, 'echo "export PATH=\$PATH:/opt/puppetlabs/bin" >> ~/.bashrc')
+# end
```
nwolfe: We should get rid of this file if it's only meant for us humans if/when we log in to the system :smile:

jenkins-integration/scripts/foss_install.sh
```diff
@@ -0,0 +1,24 @@
...
+set -x
+
+export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
+
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-centos6-64ma}"
+export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-beaker/install/foss/}"
+export BEAKER_CONFIG="scripts/hosts.cfg"
+export BEAKER_KEYFILE="~/.ssh/id_rsa-acceptance"
+export BEAKER_HELPER="beaker/helper.rb"
+export BEAKER_INSTALL_TYPE="foss"
```
nwolfe: Hmm...I'm wondering if this is the correct value to use here.  I'm seeing `aio` used now.  Not sure if we should change this though, but if we run into pathing problems we should keep this in mind.

jpinsonault:
@nwolfe Per our discussion I removed genconfig/sqa_utils
Also removed the script that adds the bin folder to the path.

jenkins-integration/scripts/foss_install.sh
jpinsonault: That's a good question, I'm not sure what the difference is either

jenkins-integration/scripts/foss_install.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -x
+
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-centos6-64ma}"
+export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-beaker/install/foss/}"
+export BEAKER_KEYFILE="~/.ssh/id_rsa-acceptance"
+export BEAKER_HELPER="beaker/helper.rb"
+export BEAKER_INSTALL_TYPE="foss"
```
nwolfe: This Beaker `--type` argument determines file paths for a bunch of things, which in the past have been different between FOSS and PE installations.  In the new world, the paths for FOSS and PE have been unified under new locations, which are available with the `aio` argument.

I think this means we can always use `--type aio` for these scripts, regardless of FOSS or PE.  This would only break things if we try running against pre-AIO stuff, but I'm not sure that we're going to.

@jpinsonault If we run this with `"aio"` and it works, then I think we should use that.  I'll test this out in a minute.

jpinsonault: @nwolfe Ahh, ok, thanks for the info
Tested it out on my end and it works fine with aio

jpinsonault:
@nwolfe Went ahead and switched the install type to aio, since it works fine in my testing

jenkins-integration/scripts/foss_install.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -x
+
+export GENCONFIG_LAYOUT="${GENCONFIG_LAYOUT:-centos6-64ma}"
```
nwolfe: Oops, guess we can get rid of this in these scripts.

jpinsonault: D'oh! Thanks, pushed another commit

.gitignore
```diff
@@ -13,3 +13,7 @@ cache
 ivy2.tgz
 ivy-tmp
 proxy-recorder/user-files
+
+# Bundler
+.bundle
+vendor
```
nwolfe: I noticed that beaker will generate a `log` directory - might as well add that too

jpinsonault: Good call, I believe it creates a `junit` dir too, I'll add them both

jpinsonault:
Updated .gitignore with beaker output folders

nwolfe:
It works! :+1:

jenkins-integration/beaker/install/pe/10_install_pe.rb
```diff
@@ -0,0 +1,3 @@
+test_name 'Install PE'
+
+install_pe
```
cprice404: I presume this beaker method already has support for picking a specific version, configured via env vars or something?

cprice404:
I skimmed this; it looks good to me at a glance, and it seems like @nwolfe and @erikPrime have already been reviewing more thoroughly, so I'm +1 when they are!

jenkins-integration/beaker/install/pe/10_install_pe.rb
jpinsonault: Yeah, by environment variables or beaker host file config options. The readme in the `install` dir (in this PR) has some more info

cprice404: word

PR #28, opened 2015-06-10
(SERVER-713) Add directory environment support

nwolfe:
This commit adds an additional property to the node config JSON called
'environment' which specifies which directory environment the associated
modules will be installed to.

The default 'production' environment is used if the property is empty or
missing.

### DIFF
--- a/jenkins-integration/lib/puppet/gatling/config.rb
+++ b/jenkins-integration/lib/puppet/gatling/config.rb
@@ -1,5 +1,4 @@
 require 'json'
-require 'set'
 
 ## Assumptions:
 ## 1. CWD is "jenkins-integration"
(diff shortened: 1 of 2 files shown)


nwolfe:

I realized that I haven't added support for installing modules out of the test-catalogs repository, which you were pointing out on my basic module installation PR Chris. I'm going to do that now, before jumping into the hiera stuff.

jenkins-integration/lib/puppet/gatling/config.rb
```diff
@@ -18,7 +17,12 @@ def get_node_configs(scenario)
...
-  node_configs.reduce(Set.new) do |modules, node_config|
-    modules.merge(node_config['modules'])
-  end.to_a
+  result = Hash.new { |h, k| h[k] = Array.new }
+  node_configs.each do |node_config|
+    env = node_config['environment']
+    env = 'production' if (env.nil? || env.empty?)
+    result[env] += node_config['modules']
+  end
+  result.values.each &:uniq!
```
jpinsonault: That's some fancy ruby right there =) Could the same effect be achieved with something like 

```ruby
result = Hash.new { |h, k| h[k] = Set.new }
# ...
result[env] |= node_config['modules']
```

nwolfe: Glad you understand the `Hash.new { ... }` part, but it was easier to remove duplicates with arrays+uniq! then with Set+to_array since the former can update in-place and the latter can't, meaning I would need more code to iterate and convert to arrays (and I don't want to return Set values since they're not that friendly to work with).

jpinsonault: Makes sense, thanks for the explanation!

jpinsonault:
:+1: looks good to me

PR #29, opened 2015-06-12
(SERVER-762) Install modules from test-catalogs

nwolfe:
This commit adds support for installing modules from the local
test-catalogs repository as well as GitHub.

### DIFF
--- a/jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
+++ b/jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
@@ -1,0 +1,19 @@
+require 'fileutils'
+require 'tmpdir'
+
+test_name 'Clone test-catalogs repo onto master'
+
+# This is awful - we clone locally and then SCP the whole thing over.
+# See QENG-2556 for adding a proper way to do this in Beaker.
+tmp_repo = Dir.mktmpdir('test-catalogs')
+target = '/root/test-catalogs'
+
+teardown do
+  FileUtils.remove_dir(tmp_repo)
+end
+
+repo_url = 'git@github.com:puppetlabs/test-catalogs.git'
+%x(git clone #{repo_url} #{tmp_repo} --depth 1)
+
+on(master, "test -d #{target} && rm -rf #{target} || true")
+scp_to(master, tmp_repo, target)
--- a/jenkins-integration/beaker/install/foss/50_install_modules.rb
+++ b/jenkins-integration/beaker/install/foss/50_install_modules.rb
@@ -9,10 +9,18 @@ def install_librarian_puppet(host)
 end
 
 def generate_puppetfile(modules)
-  modules
-    .map { |mod| "mod '#{mod}'" }
-    .insert(0, "forge 'https://forgeapi.puppetlabs.com'")
-    .join("\n")
+  modules.map do |mod|
+    directive = "mod '#{mod['name']}'"
+    if mod['version']
+      directive += ", '#{mod['version']}'"
+    elsif mod['path']
+      directive += ", :path => '#{mod['path']}'"
+    elsif mod['git']
+      directive += ", :git => '#{mod['git']}'"
+      directive += ", :ref => '#{mod['ref']}'" if mod['ref']
+    end
+    directive
+  end.insert(0, "forge 'https://forgeapi.puppetlabs.com'").join("\n")
 end
 
 def run_librarian_puppet(host, environment, puppetfile)
@@ -31,6 +39,6 @@ def install_environment_modules(host, modules)
 end
 
 scenario_id = ENV['PUPPET_GATLING_SCENARIO']
-modules = get_modules(get_node_configs(get_scenario(scenario_id)))
+modules = scenario_modules(scenario_id)
 install_librarian_puppet(master)
 install_environment_modules(master, modules)
--- a/jenkins-integration/config/nodes/example-node1.json
+++ b/jenkins-integration/config/nodes/example-node1.json
@@ -1,0 +1,21 @@
+{
+    "certname": "example-node1",
+    "simulation_class": "com.puppetlabs.gatling.node_simulations.FOSSPuppetserver210CatalogZero",
+    "modules": [
+        {
+            "name": "catalog_zero1",
+            "path": "/root/test-catalogs/catalog_zero/modules/catalog_zero1"
+        },
+        {
+            "name": "nwolfe-simmons",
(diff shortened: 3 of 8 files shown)


nwolfe:
Ping @jpinsonault @cprice404 for review

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
```diff
@@ -0,0 +1,16 @@
+test_name 'Clone test-catalogs repo onto master'
+
+# This is awful - we clone locally and then SCP the whole thing over.
+# See QENG-2556 for adding a proper way to do this in Beaker.
+tmp_repo = "/tmp/test-catalogs-#{Process.pid}"
+target = '/root/test-catalogs'
+
+teardown do
+  %x(rm -rf #{tmp_repo})
```
jpinsonault: I learn learn something new every day

jpinsonault:
:+1: looks good to me

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
```diff
@@ -0,0 +1,16 @@
+test_name 'Clone test-catalogs repo onto master'
+
+# This is awful - we clone locally and then SCP the whole thing over.
+# See QENG-2556 for adding a proper way to do this in Beaker.
+tmp_repo = "/tmp/test-catalogs-#{Process.pid}"
```
cprice404: there are beaker util functions for asking it for a temp dir... but those may only work for the SUT?

failing that you still might want to do something like `mktemp -d` rather than hard-coding a location in `/tmp`.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
cprice404: `rm -rf ${var}`: that's not scary at all!  :)

Might be worth doing this through Ruby's File API just to make it a bit safer?  Or maybe at least guard this with `File.directory?(tmp_repo)`?

cprice404:
+1 other than maybe some small tweaks to how you are creating / cleaning up the temp dir.

Also we may want to replace those example node configs with some more real-world ones that can be run end-to-end, as soon as we have some... but having examples is definitely valuable so I think they're good for now.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
nwolfe: Sure, I can go through the Ruby file utils to do this, but...

How would that make it any safer? I still need to tell it to force remove the directory.  How does testing for a directory help?  I could see this being valuable if I was deleting a file and wanted to make sure to not delete a directory.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
nwolfe: Yeah Beaker only helps with that on the host nodes, not the local coordinator machine.

I actually don't want to make the directory, because the git clone will complain if it already exists.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
nwolfe: Updated to use `FileUtils.rm_rf`, which I agree is better than shelling out

nwolfe:
@cprice404 Feedback squashed in.

> Also we may want to replace those example node configs with some more real-world ones that can be run end-to-end, as soon as we have some... but having examples is definitely valuable so I think they're good for now.

Yes, and there are still some things left to "fill in" with these examples, like the simulation_class and the classes.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
cprice404: re: beaker: roger.

re: constructing the path to the temp dir... `mktemp -t` maybe, then?  The point is that you're letting the OS tell you where is the appropriate place to create temp files, rather than assuming it's always `/tmp`, which it may not be.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
cprice404: Yeah, I dunno.  I guess `File.directory?` by itself wouldn't be any safer.  Just trying to think of safeguards to prevent a `rm -rf /`.  It's not a huge deal since these are just test machines, but if this were production code I'd want to put in *some* kind of safeguard.  Anyway, I'm fine with it as-is for now.

cprice404:
I'd vote for asking the OS to help generate the temp path, but I don't consider it to be a huge deal.  +1, @jpinsonault feel free to merge whenever... only reason I'm not pushing the button now is to see if @nwolfe has any further thoughts on the `mktemp` thing.

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
rick: (Sorry for the :car::dash:-by) I support the mktemp recommendation.  Ruby stdlib has support for `mktemp`-style temp directories:  http://ruby-doc.org/stdlib-2.0.0/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir

nwolfe: Ahh, I was looking for the `mktemp -t` behavior but I didn't realize this is what the docs meant by "template".  I would prefer this over `mkdir /tmp/...`, although why stop there? Let's keep going and get rid of the shell-out and just `Dir.mktempdir` like @rick kindly reminded us :D

jenkins-integration/beaker/install/foss/40_clone_test_catalogs.rb
```diff
@@ -0,0 +1,19 @@
...
+
+test_name 'Clone test-catalogs repo onto master'
+
+# This is awful - we clone locally and then SCP the whole thing over.
+# See QENG-2556 for adding a proper way to do this in Beaker.
+tmp_repo = Dir.mktmpdir('test-catalogs')
+target = '/root/test-catalogs'
+
+teardown do
+  FileUtils.remove_dir(tmp_repo)
```
nwolfe: @cprice404 Updated to use Ruby utilities.

(I squashed the commits)

jpinsonault:
:+1:

PR #30, opened 2015-06-24
(SERVER-330) Automate launching gatling test

jpinsonault:
These scripts launch a single gatling run on the gatling driver and the memory monitoring script on the SUT.

Since it only launches a single gatling run, it's assuming that the length of the gatling run will be modified by changing the scenario to have more repetitions, as opposed to repeating a couple repetitions forever in a loop like we've been doing in our manual testing.

It launches gatling and the memory monitoring script in the background and exits immediately. I did this because I'm guessing we don't want to have jenkins jobs running for months at a time, hanging while gatling finishes.


jenkins-integration/beaker/gatling/watch_puppetserver_mem.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
jpinsonault: I grabbed this script off of one of the SUTs and cleaned it up quite a bit. But it's probably got room for improvement. 

Also not sure if this is a good place to store the script

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
```diff
@@ -0,0 +1,35 @@
...
+  abort "GATLING_SUT_HOSTNAME environment variable required"
+end
+
+repo_url = "https://github.com/puppetlabs/gatling-puppet-load-test.git"
+repo_path = "/root/gatling-puppet-load-test_#{simulation_id}"
+
+sbt_path = "#{repo_path}/simulation-runner"
+
+# Sets the environment variables and runs sbt
+sbt_command = "PUPPET_GATLING_SIMULATION_CONFIG=config/scenarios/#{gatling_scenario} PUPPET_GATLING_SIMULATION_ID=#{simulation_id} PUPPET_GATLING_MASTER_BASE_URL=#{sut_hostname} sbt run"
```
nwolfe: Long line, probably want to break this up with some `\`'s like:
```sh
sbt_command = 
 "PUPPET_GATLING_SIMULATION_CONFIG=config/scenarios/#{gatling_scenario} \
  PUPPET_GATLING_SIMULATION_ID=#{simulation_id} \
  PUPPET_GATLING_MASTER_BASE_URL=#{sut_hostname} \
  sbt run"
```

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
```diff
@@ -0,0 +1,35 @@
...
+if not simulation_id
+  abort "GATLING_SIMULATION_ID environment variable required"
+elsif not gatling_scenario
+  abort "GATLING_SCENARIO environment variable required"
+elsif not sut_hostname
+  abort "GATLING_SUT_HOSTNAME environment variable required"
+end
+
+repo_url = "https://github.com/puppetlabs/gatling-puppet-load-test.git"
+repo_path = "/root/gatling-puppet-load-test_#{simulation_id}"
```
nwolfe: How come each simulation gets its own repository? Because run results will be stored in the repo itself? 

That's fine for now, just making sure I understand

jenkins-integration/beaker/gatling/20_start_memory_watcher.rb
```diff
@@ -0,0 +1,27 @@
+test_name "Start memory watching script on SUT"
+
+simulation_id = ENV['GATLING_SIMULATION_ID']
+logger_refresh = ENV['MEMORY_WATCHER_REFRESH'] || 300
+
+if not simulation_id
+  abort "GATLING_SIMULATION_ID environment variable required"
+end
+
+remote_logger_dir = "/root/#{simulation_id}_memory_usage"
```
nwolfe: Since each simulation is getting its own gatling-puppet-load-test repo too, I might try to house all things related to that simulation in a single directory so it's easier to see all the things.

I'd probably consider creating an intermediate directory based on the simulation ID that we can put the gatling-puppet-load-test repo under, as well as this logging information.

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
jpinsonault: Good call

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
jpinsonault: Yeah the results get stored in the `simulation-runner/results` directory, and it seems easier to separate them out this way. It could be the case (probably is) that it wouldn't cause issues if they all ran from the same directory, but this is how we've been doing it manually so far

jenkins-integration/beaker/gatling/20_start_memory_watcher.rb
jpinsonault: Are you talking about the SUT or the gatling driver? I don't think we'll need `gatling-puppet-load-test` on the SUT (where this script is pointed)

jenkins-integration/beaker/README.md
```diff
@@ -43,3 +52,40 @@ When running the PE installer, two options can be set:
...
+    roles:
+    - agent
+    - sut
+  kf58a5eknd06u1s:
+    platform: el-6-x86_64
+    roles:
+    - agent
+    - master
+    - gatling
+```
```
cprice404: You wouldn't really have the master and gatling on the same node usually, would you?

cprice404: also would re-using 'master' suffice, instead of introducing 'sut'?

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
```diff
@@ -0,0 +1,35 @@
+test_name "Start gatling scenario"
+
+simulation_id = ENV['GATLING_SIMULATION_ID']
+gatling_scenario = ENV['GATLING_SCENARIO']
+sut_hostname = ENV['GATLING_SUT_HOSTNAME']
```
cprice404: could we get this from the beaker config instead of an env var?

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
```diff
@@ -0,0 +1,35 @@
...
+sbt_command = "PUPPET_GATLING_SIMULATION_CONFIG=config/scenarios/#{gatling_scenario} PUPPET_GATLING_SIMULATION_ID=#{simulation_id} PUPPET_GATLING_MASTER_BASE_URL=#{sut_hostname} sbt run"
+
+# To keep log/result files separate on the gatling machine
+step "Clone gatling-puppet-load-test into new folder for this scenario"
+on(gatling, "git clone #{repo_url} #{repo_path}")
+
+step "Start gatling with scenario on gatling machine"
+# This seems like an ugly but accepted way of starting background tasks over ssh
+nohup_cmd = "nohup bash -c 'cd #{sbt_path} && #{sbt_command}' &> #{sbt_path}/sbt.out < /dev/null &"
+on(gatling, nohup_cmd)
```
cprice404: So, with this approach, we wouldn't have any way of knowing when the run actually completed, correct?  In other words, the jenkins job would end immediately and would leave the rest of the stuff running indefinitely?

I'm thinking we really want to have the jenkins job keep running if we can, so that it can collect artifacts from the run when it completes.  I believe that if you look at the old code, we were doing this by using the jenkins `sbt` plugin:

https://wiki.jenkins-ci.org/display/JENKINS/sbt+plugin

Then we were using the `puppet-gatling-jenkins-plugin` after that, to collect the data and build graphs that would show up in jenkins itself.

So maybe we don't really need to have gatling represented in the beaker config at all?  gatling would end up being more of a jenkins-side requirement.  And for simple scenarios, you'd just run gatling directly on the master jenkins node, but in production, we could have some jenkins slaves that were especially for gatling runs, and restrict the gatling jobs to only run on those?  e.g. maybe we eventually set up one of those 32-core blades as a Jenkins slave?

cprice404: Sorry if I didn't give you enough initial guidance on that part :(

jenkins-integration/beaker/README.md
jpinsonault: Yes, you're right. It looks like I left `- master` in there accidentally. And yeah, using `master` instead of `sut` would make more sense I think

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
jpinsonault: It looks like you can. As simple as `sut_hostname = master.hostname` If you've set up a node with a `master` role. I'll switch to that

jenkins-integration/beaker/gatling/10_start_gatling_scenario.rb
jpinsonault: No this makes more sense. For some reason I had it in my head that it would be unreasonable to have a jenkins job running for 2-3 months, and that we'd do it a little more asynchronously. Start the gatling run, schedule a new job in the future or have the gatling machine itself tell jenkins it had finished. Something like that.

But your way simplifies things. So for this PR, should I create a JJB job that uses the sbt plugin then you think? I can just have it run sbt on the master jenkins node for now and we can update it later to use the specialized gatling slaves

cprice404: Yeah, for now I think we should focus on the use cases around shorter-term performance and regression testing.  So, assuming that most of these runs will be done, say, nightly or weekly, and will only run for a few hours.

We'll expand out from there to deal with the 'months-long' runs.  For those we may have to have a different pattern as I don't know how Jenkins will do with a run that goes for that long.  But the ability to have Jenkins collect artifacts and update graphs, etc., at the end of a run is a big enough win to where I think that should be our first step.

So, yeah, JJB seems like the way to go.  Maybe also updating the dev environment tooling to install the sbt plugin for jenkins?

jpinsonault: Ok, yeah that makes sense. Thanks @cprice404! Agreed on the dev environment too

jpinsonault:
@cprice404 I removed the beaker script and added a jjb file for starting sbt. It should at least be a decent starting point

jenkins-integration/jenkins-jobs/run_gatling_scenario.yml
```diff
@@ -0,0 +1,30 @@
...
+        description: "The gatling scenario file to run"
+    - string:
+        name: PUPPET_GATLING_SIMULATION_ID
+        description: "An descriptive identifier for this run"
+    - string:
+        name: PUPPET_GATLING_MASTER_BASE_URL
+        description: "The URL of the puppet master SUT"
+
+    builders:
+        - shell: sbt run
```
nwolfe: Do we need `shell` if we're `sbt:` below?

jpinsonault: d'oh! Forgot to remove that line

jpinsonault: fixed

dev/Gemfile.lock
```diff
@@ -0,0 +1,209 @@
+GEM
```
nwolfe: Is this file useful to you? Feel free to gitignore it if it's not. 

Looked like you and I have different versions of Bundler installed (which is totally fine), but that meant every time I ran bundler it would change this file with the Bundler version I was using. I haven't been working in `dev/` so not really a bother at this point, just saying.

jenkins-integration/jenkins-jobs/run_gatling_scenario.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- job-template:
+    name: run_gatling_scenario
+    parameters:
+    - string:
+        name: PUPPET_GATLING_SIMULATION_CONFIG
+        description: "The gatling scenario file to run"
+    - string:
+        name: PUPPET_GATLING_SIMULATION_ID
+        description: "An descriptive identifier for this run"
```
nwolfe: s/An/A

jenkins-integration/beaker/README.md
cprice404: Still possible to get rid of the `sut` role?

jenkins-integration/jenkins-jobs/run_gatling_scenario.yml
```diff
@@ -0,0 +1,29 @@
...
+    - string:
+        name: PUPPET_GATLING_MASTER_BASE_URL
+        description: "The URL of the puppet master SUT"
+
+    builders:
+        - sbt:
```
cprice404: Does the sbt plugin automatically take all of the job parameters and make them available to sbt as environment variables?

cprice404:
+1 other than the last couple of questions I added.  @nwolfe feel free to merge when you and @jpinsonault are satisfied.

jenkins-integration/jenkins-jobs/run_gatling_scenario.yml
jpinsonault: Jenkins makes them available as environment variables to all of the builders I think

cprice404: awesome

jpinsonault:
@cprice404 Got rid of the `sut` role and updated the docks to reflect that
@nwolfe Fixed typo and removed `Gemfile.lock`

jenkins-integration/beaker/gatling/watch_puppetserver_mem.sh
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+# Output PID
+echo $$ > watch_puppetserver_mem.pid
+
+# Dummy value
+origpid="firstpid"
+
+# What process to look for
+process_name="puppet-server"
```
nwolfe: The PE process is named `pe-puppetserver`, hmm

nwolfe: ...but the `ps -ef | grep -i java | grep -i $process_name | awk '{print $2}'` still seems to work as `puppet-server` matches the name of the jar...

I suppose this works for now.  I hate the idea of putting conditionals based on FOSS/PE anywhere; too bad there's no unified process name (yet?).

jpinsonault: Yeah I noted during manual testing that it still worked with PE, and left it. But we could make the  regex `(puppet-server|pe-puppetserver)`?

nwolfe: Hmm, I guess I'm inclined to just leave it for now? Yeah

nwolfe:
:+1:

PR #31, opened 2015-06-26
(SERVER-687) Classify agent nodes

nwolfe:
This commit adds support for classifying agent nodes with classes
specified in the node config JSON. Each node config will be a Puppet
node definition in the appropriate environment-specific site.pp file.

---

Note this is FOSS only, as the Node Classifier will be used for the PE implementation. This is work that still needs to be done, and will probably be a subsequent PR under the same JIRA ticket number.

### DIFF
--- a/jenkins-integration/lib/puppet/gatling/config.rb
+++ b/jenkins-integration/lib/puppet/gatling/config.rb
@@ -5,28 +5,36 @@ require 'json'
 ## 2. Scenario config JSON files in "./config/scenarios/*.json"
 ## 3. Node config JSON files in "./config/nodes/*.json"
 
-def get_scenario(scenario_id)
+def parse_scenario_file(scenario_id)
   JSON.parse(File.read(File.join('config', 'scenarios', scenario_id + '.json')))
 end
 
-def get_node_configs(scenario)
+def parse_node_config_files(scenario)
   scenario['nodes'].map do |node|
     config_path = File.join('config', 'nodes', node['node_config'] + '.json')
     JSON.parse(File.read(config_path))
   end
 end
 
-def get_modules(node_configs)
-  result = Hash.new { |h, k| h[k] = Array.new }
-  node_configs.each do |node_config|
-    env = node_config['environment']
-    env = 'production' if (env.nil? || env.empty?)
-    result[env] += node_config['modules']
-  end
-  result.values.each &:uniq!
-  result
+# Returns the list of node configs hashes in the given scenario.
+def node_configs(scenario_id)
+  parse_node_config_files(parse_scenario_file(scenario_id))
 end
 
-def scenario_modules(scenario_id)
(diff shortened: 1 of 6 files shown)


jenkins-integration/lib/puppet/gatling/config.rb
```diff
@@ -5,28 +5,36 @@
...
-  result
 end
 
-def scenario_modules(scenario_id)
-  get_modules(get_node_configs(get_scenario(scenario_id)))
+# Returns a hash from environments to modules; removes duplicate modules.
+def modules_per_environment(node_configs)
+  node_configs = group_by_environment(node_configs)
+  modules = node_configs.map do |env, configs|
+    [env, configs.map { |c| c['modules'] }.flatten.uniq]
```
jpinsonault: :boom: beautiful

jpinsonault:
:+1: Looks good to me

cprice404:
@nwolfe do you have any thoughts on how we can validate this?

e.g. in the old world, we would call the PE rake tasks to ask for a list of classes assigned to a node, so that we could make sure the catalog compilation was doing what it needed to be doing.  I think you can probably get away with something similar for PE when you get to that part, by querying the NC API.  I'm wondering if you have any ideas for OSS, though?

e.g., if the puppet.conf file somehow got configured to say that the base environmentpath was /tmp/ohcrap, then these site.pp files you're writing wouldn't take effect.  Is there any way we can guard against that?

If you do have any ideas on this, that doesn't necessarily need to block this PR.  If you want to put it into a different PR that's fine by me.  I'm +1 on this PR pending your thoughts on that.  @jpinsonault feel free to merge it w/o waiting for me depending on what @nwolfe says.

nwolfe:
To follow up...

bq. Note this is FOSS only, as the Node Classifier will be used for the PE implementation. This is work that still needs to be done, and will probably be a subsequent PR under the same JIRA ticket number.

I've been working on PE support for this using the Node Classifier.  This has taken me a bit longer than expected as things are silently not working, but in the process of debugging this I'll likely learn how we can implement the next step: validate our classification.

...which, as you pointed out @cprice404, is an absolute must in my mind.  I'm hoping in the FOSS implementation we can use the master's `resource_type(s)` HTTP API to do this validation, but haven't dug into that much yet.

As far as the implementation is going, I now need to share some of the beaker steps between FOSS and PE, specifically the "clone test-catalogs" and "install modules" files. In order to share them, I've created a third `shared/` directory next to `foss/` and `pe/`, and then the Beaker scripts now explicitly list out the steps (files) they want rather than just referring to a whole directory.

cprice404:
@nwolfe thanks for the update.  That all sounds good to me.  I'm fine with this being merged as-is if you think it's the right granularity for a building block.  If so just let @jpinsonault know and he can go ahead and merge it.

PR #32, opened 2015-07-02
(SERVER-687) Classify PE nodes

nwolfe:
This commit adds PE support for node classification via the Node
Classifier and the Puppet Labs 'scooter' gem. The Gemfile.lock is now
ignored because it will be constantly changing as we run FOSS and PE as
there are now gems that will only be used during PE runs.


jenkins-integration/beaker/install/pe/60_classify_nodes.rb
```diff
@@ -0,0 +1,26 @@
+require 'puppet/gatling/config'
+require 'scooter'
+
+test_name "Classify PE agents via Node Classifier"
+
+def classify_pe_nodes(classifier, nodes)
+  production_id = classifier.get_node_group_id_by_name('Production environment')
```
nwolfe: I feel like I'm hardcoding this to a UI label rather than some more concrete ID, but this is apparently the best way to refer to this group at the time.

jenkins-integration/Gemfile
```diff
@@ -1,3 +1,4 @@
 source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
-gem 'beaker', :github => 'puppetlabs/beaker', :branch => 'master'
+gem 'beaker', '2.15.1'
+gem 'scooter', :git => 'git@github.com:puppetlabs/scooter.git', :tag => '2.2'
```
nwolfe: :clap: :clap: :clap: 

Turns out we don't *need* to use the internal rubygems mirror and deal with Gemfile conditionals and remembering to set GEM_SOURCE when calling Beaker scripts.  

Bundler can pull gems out of private GitHub repositories just fine. The only thing we lose is the ability to do version ranges, although you can point it to a branch. Either way we don't need version ranges for these gems, we can just pin to a release.

nwolfe:
Ping @jpinsonault @cprice404 for review.

In the meantime I'm working on PR(s) for attempting to validate that the classification is correct prior to running the "agents" (gatling).

I'm thinking for PE this should be pretty easy via the Node Classifier.  For FOSS it's a bit more involved, but I'm thinking a combination of the master's HTTP endpoints (`environments`, `resource_type`) might work.

I don't think attempting to run the agent in some form is the right way here as we have no idea what the modules might do to the agent machine, or any good way to check that the agent received the correct classes.

cprice404:
+1

@nwolfe (for future PR): you could make a direct request to the `catalog` endpoint to help with the validation for OSS, and validate that the expected classes came back in the resulting catalog?

If PE case is easier, I think you should do that first, and we should probably consider the option of just moving forward and getting a PE job running end-to-end and coming back around for the OSS class validation later?

jenkins-integration/beaker/install/pe/60_classify_nodes.rb
```diff
@@ -0,0 +1,26 @@
...
+  nodes = group_by_environment(nodes)
+  nodes.each_pair do |env, node_configs|
+    node_configs.each do |config|
+      classifier.create_new_node_group_model(
```
rick: Fwiw, if you may run some or all of these tests repeatedly (i.e., run the suite, something fails, run parts of it again); or if there's a possibility this node group already exists for some reason, there's a `find_or_create_node_group_model` method available.  We've used it in a number of places in our scale testing setup code:

https://github.com/puppetlabs/pe_acceptance_tests/blob/4.0.x/setup/scale/install/20_classification/30_clamps.rb#L50-L52

nwolfe:
> You could make a direct request to the catalog endpoint to help with the validation for OSS, and validate that the expected classes came back in the resulting catalog?

I thought about this, but the GET request would need the facts for the agent in order to compile the correct catalog.  That's not too bad, I could just run `facter` on the agent...except their are no agents (it's just gatling sending HTTP requests), and I couldn't fake it by running `facter` on the master because that would result in different facts (hostname, etc).  The right/correct set of facts to use would the ones buried in the gatling simulation Scala class (yeah it might actually be in one of those text files, but same difference).

I was thinking that we could use the other endpoints to ask the master the following:
* What are the environments?
* What are the nodes defined in each of those environments?
* What are the classes defined in each of those nodes?

The last part might be hardest, although I think we could probably be reasonably sure that if it's still aware of the environments and classes (regardless of nodes) then things are probably okay. I would expect that it's more likely the paths would've changed rather than the contents of the `site.pp` files.

In other words, we probably don't need to validate absolutely everything and could instead be reasonably certain things are good to go with a few targeted checks.

cprice404:
> What are the nodes defined in each of those environments?

How would you ask it that?  AFAIK only PuppetDB would keep track of that sort of data, and then only after an agent had gone through a successful run.  Did you find some endpoint that I'm not aware of?  :)

I agree that we don't need to validate absolutely everything and am willing to make some compromises.  (And also defer this chunk of work to sometime down the road if it's too hairy.)

> I thought about this, but the GET request would need the facts for the agent in order to compile the correct catalog.

Would it?  If our mechanism for classifying the nodes is simply to add the classes to site.pp with an explicit `include foo` statement, then we know for sure that the logic for whether the class gets included is not conditional based on the values of the facts.  Right?  So, if you request the catalog for the correct certname, and send no facts at all, you should get a catalog back that includes all of the correct classes.  It won't necessarily be exactly the *same* catalog that would get compiled for the real node with its facts, but it should at least include the classes we care about so that we can check for them, right?

We should probably copy/paste all of these comments into the jira ticket and continue this conversation there so it's preserved for future reference, especially in case we decide to defer the implementation work on this.

nwolfe:
> > What are the nodes defined in each of those environments?

> How would you ask it that? AFAIK only PuppetDB would keep track of that sort of data, and then only after an agent had gone through a successful run. Did you find some endpoint that I'm not aware of? :)

The `resource_type` endpoint seems to give me what I expect. 
For example, to find all nodes in environment `exampleenv1`:
```sh
$ curl https://gatling-master.localdomain:8140/puppet/v3/resource_types/*\?environment=exampleenv1\&kind=node | pp
[
    {
        "kind": "node",
        "name": "example-node1"
    },
    {
        "kind": "node",
        "name": "example-node2"
    }
]

$ cat /etc/puppetlabs/code/environments/exampleenv1/manifests/site.pp
node 'example-node1' {
include catalog_zero1
include simmons
include python
}
node 'example-node2' {
include firewall
include tomcat
}
```

cprice404:
Wow.  I had no idea that it did that.  And so then how are you thinking you'd go from that to the list of classes?  There's a way to do that through 'resource_types' as well?

nwolfe:
> Would it? If our mechanism for classifying the nodes is simply to add the classes to site.pp with an explicit include foo statement, then we know for sure that the logic for whether the class gets included is not conditional based on the values of the facts. Right?

Whoa...I didn't know it was as simple as looking at the `site.pp` file and checking for fact references. If that's the case, then yes this would work. I'll check that out.

> And so then how are you thinking you'd go from that to the list of classes? There's a way to do that through 'resource_types' as well?

This is the missing piece - I haven't been able to get this information; the nodes->classes associations.
We could rearrange how we classify nodes in FOSS to make it possible/easier to get this information (such as specifying the classes for a given node in hiera), but that doesn't seem like the right thing to do.

I'm going to check out the `catalog` endpoint like you mentioned, that will be better if it works anyway, and less code/work.

PR #33, opened 2015-07-11
(SERVER-780) Validate node classification

nwolfe:
This commit adds a beaker script that will query the master's HTTP API
and validate that each node will receive the correct classes during the
catalog request.

This beaker script requires that each node already has an SSL
certificate and private key on the master. Unfortunately there is no
automation for putting the certificate and key on the master yet,
see SERVER-787.

This validation script will likely be executed as the very last thing
prior to launching Gatling.

### DIFF
--- a/jenkins-integration/Gemfile
+++ b/jenkins-integration/Gemfile
@@ -2,3 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 gem 'beaker', '2.15.1'
 gem 'scooter', :git => 'git@github.com:puppetlabs/scooter.git', :tag => '2.2'
+
+if File.exists? "#{__FILE__}.local"
+  eval(File.read("#{__FILE__}.local"), binding)
+end
--- a/jenkins-integration/beaker/install/shared/validate_classification.rb
+++ b/jenkins-integration/beaker/install/shared/validate_classification.rb
@@ -1,0 +1,39 @@
+require 'puppet/gatling/config'
+require 'json'
+
+test_name "Validate node classification"
+
+## ASSUMPTIONS:
+## - agent certs/keys are already on the master in the $ssldir (SERVER-787)
+
+def check_for_classes(catalog_response, config)
+  expected = config['classes']
+  actual = catalog_response['classes']
+  all_found = expected.all? { |klass| actual.include?(klass) }
+  if not all_found
+    abort "Node '#{config['certname']}' missing classes.\nExpected: #{expected}\nActual: #{actual}"
+  end
+end
+
+def build_catalog_query(host, node_config, server, cacert)
+  certname = node_config['certname']
+  cert = on(host, puppet("config print hostcert --certname #{certname}")).stdout.chomp
+  key = on(host, puppet("config print hostprivkey --certname #{certname}")).stdout.chomp
+  environment = node_environment(node_config)
+  url = "https://#{server}:8140/puppet/v3/catalog/#{certname}\?environment=#{environment}"
+  "curl --cert #{cert} --key #{key} --cacert #{cacert} #{url}"
+end
+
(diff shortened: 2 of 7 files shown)


jenkins-integration/config/nodes/example-node1.json
```diff
@@ -11,14 +11,14 @@
             "path": "/root/test-catalogs/catalog_zero/modules/catalog_zero1"
```
nwolfe: These modules are a little easier to work with during development as they have default values for class parameters.
Nothing is using these still, just easier for development.

nwolfe:

@cprice404 Querying the catalog endpoint turned out to be the best option. All we need now is the certs/keys for each agent on the master (SERVER-787).

jenkins-integration/Gemfile
```diff
@@ -2,3 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 gem 'beaker', '2.15.1'
 gem 'scooter', :git => 'git@github.com:puppetlabs/scooter.git', :tag => '2.2'
+
+if File.exists? "#{__FILE__}.local"
+  eval(File.read("#{__FILE__}.local"), binding)
+end
```
cprice404: ? this seems spooky

jenkins-integration/beaker/install/shared/validate_classification.rb
```diff
@@ -0,0 +1,40 @@
...
+    step "Checking classes for node '#{config['certname']}'"
+    catalog_query = build_catalog_query(host, config, server, cacert)
+    response = on(host, catalog_query).stdout.chomp
+    check_for_classes(JSON.parse(response), config)
+  end
+end
+
+scenario_id = ENV['PUPPET_GATLING_SCENARIO']
+nodes = node_configs(scenario_id)
+validate_classification(master, nodes)
```
cprice404: You put this in `shared`; I guess it works fine for both FOSS and PE?

cprice404:
+1

jenkins-integration/Gemfile
camlow325: FWIW, we do this in Puppet Server, too - https://github.com/puppetlabs/puppet-server/blob/puppet-server-2.1.1/Gemfile#L26-L28.  @jeffmccune probably could give more input here as to when this pattern was established in other Puppet projects.

nwolfe: Haha this is for loading the `Gemfile.local` if you have one.  Turns out Bundler doesn't have a global file akin to leiningen's `profiles.clj`, so this is how you support developer-specific gems.

jenkins-integration/beaker/install/shared/validate_classification.rb
nwolfe: Yes! Works for both FOSS and PE; just need the client's certificate.

Nothing is calling this script, but I expect the shell script that launches Gatling would probably call this validation first.

cprice404:
@jpinsonault would you mind reviewing/merging this today?

jenkins-integration/lib/puppet/gatling/config.rb
```diff
@@ -21,12 +21,18 @@ def node_configs(scenario_id)
...
+  (env.nil? || env.empty?) ? 'production' : env
+end
+
 # Group the list of node configs into a hash keyed by their environments.
 # A nil or empty environment will be interpreted as 'production'.
 def group_by_environment(node_configs)
   node_configs.group_by do |config|
-    env = config['environment']
-    (env.nil? || env.empty?) ? 'production' : env
+    node_environment(config)
```
jpinsonault: Good call, easier to read this way

jpinsonault:
Looks good to me :+1:

PR #34, opened 2015-07-17
(SERVER-714) Add Hiera support

nwolfe:
This commit adds support for optionally specifying a Hiera "tree" as
part of the scenario. Each hiera tree is made up a hiera.yaml
configuration file and multiple data directories.

Installation will copy the hiera.yaml file to the master, and copy the
data directories into place as specified by :datadir: properties in
hiera.yaml.

### DIFF
--- a/jenkins-integration/beaker/install/shared/70_install_hiera.rb
+++ b/jenkins-integration/beaker/install/shared/70_install_hiera.rb
@@ -1,0 +1,24 @@
+require 'puppet/gatling/config'
+
+test_name 'Install Hiera data'
+
+def install_hieraconfig(host, hiera)
+  configfile = hiera_configpath(hiera)
+  configpath = on(host, puppet('config print hiera_config')).stdout.chomp
+  scp_to(host, configfile, configpath)
+end
+
+def install_hieradata(host, hiera)
+  datadirs = hiera_datadirs(hiera)
+  datadirs.each do |localpath, hostpath|
+    targetpath = File.dirname(hostpath)
+    scp_to(host, localpath, targetpath)
+  end
+end
+
+scenario_id = ENV['PUPPET_GATLING_SCENARIO']
+hiera = parse_scenario_file(scenario_id)['hiera']
+if not hiera.nil?
+  install_hieraconfig(master, hiera)
+  install_hieradata(master, hiera)
+end
--- a/jenkins-integration/config/hieras/basic-example/hiera.yaml
+++ b/jenkins-integration/config/hieras/basic-example/hiera.yaml
@@ -1,0 +1,10 @@
+---
+:backends:
+  - yaml
+  - json
+:hierarchy:
+  - common
+:yaml:
+  :datadir: /etc/puppetlabs/code/hieradata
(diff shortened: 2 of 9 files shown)


nwolfe:
I plan to add documentation towards the end of the current development thrust; once things stabilize a bit and we move on to the next phase where these configuration files aren't changing as often.

jenkins-integration/beaker/install/shared/70_install_hiera.rb
```diff
@@ -0,0 +1,24 @@
...
+    targetpath = File.dirname(hostpath)
+    scp_to(host, localpath, targetpath)
+  end
+end
+
+scenario_id = ENV['PUPPET_GATLING_SCENARIO']
+hiera = parse_scenario_file(scenario_id)['hiera']
+if not hiera.nil?
+  install_hieraconfig(master, hiera)
+  install_hieradata(master, hiera)
```
nwolfe: Oops, need to restart the master service in order for it to pick up the hiera data.

nwolfe: Ugh...hope there's a general way I can do this without having to condition it on FOSS or PE service names...

nwolfe: Hmm, I guess the FOSS service isn't even started by default, so the master isn't running anyway.  In PE it is started though.  It might be useful in general to assume the services are not running and have an explicit step before running Gatling that starts the service, so individual setup scripts don't have to worry about that.

nwolfe:
@cprice404 Regarding your comment on the JIRA ticket...

> I think what we're going to end up wanting to do is baking some hiera data files into the 'test-catalogs' repo, to pair alongside various modules. Classes in a given module can then be modified to look up data from hiera as part of compilation.

I think we need the hiera data in this repo instead since it will need to be applied to all modules involved in the simulation, not just the ones from `test-catalogs`.

A huge benefit of Puppet and Hiera integration is that Puppet will automatically (implicitly) pull class parameters from Hiera without having to make an explicit `hiera(...)` function call in the module Puppet code.  This means we don't need to couple the data with the modules as tightly, although at some point we may consider extending hiera support into the `test-catalogs` repository for such purposes.

nwolfe:
Ping @jpinsonault for review

jenkins-integration/beaker/install/shared/70_install_hiera.rb
jpinsonault: Yeah that makes sense, doing a start/restart before running gatling

jpinsonault:
This looks good from my end

rick:
Just a cc: to @doug-rosser and @ericwilliamson, as in the scale testing world we also have an interest in loading hiera data. 

See also [QENG-2119](https://tickets.puppetlabs.com/browse/QENG-2119)

nwolfe:
PSA I'm going to merge this PR today or tomorrow before starting on the next set of work.

nwolfe:
Merging.

Feel free to question/comment still.

PR #35, opened 2015-07-28
(SERVER-809) Launch sbt from script

jpinsonault:
ping @nwolfe


simulation-runner/run_sbt.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
nwolfe: Man, so many similar files.

I think we should get rid of this one and just inline the call to `sbt run` in the `on(gatling, ...)` call in `20_run_sbt.rb`.

jpinsonault: Missed this comment, sorry. Yeah that sounds reasonable to me, I don't have a strong opinion

jpinsonault:
@nwolfe Inlined the sbt command

nwolfe:
:+1:

PR #36, opened 2015-08-05
(SERVER-689) First pass at multi-env scenario

nwolfe:
This commit includes new catalog_zero agent recordings against 2
environments. Naming of things in here can be improved, but I wanted to
at least get recording that's up to date with Puppet Server, Puppet
Agent, and the new environment support.

I imagine there will likely be some follow-up work to touch these things
up a bit once we start running it through the system more.


nwolfe:

A current deficiency we have is that N environments require N node configurations, since environments are tightly-coupled to the agent's that ran against them.  There's no reason this needs to be the case, it's just a result of the initial environment support implementation.

This discouraged me from doing what the ticket mentions:
> Ideally we'd have 10 or more environments with some code in them (could just use some stuff from PE-9704), and we'd have at least two different agent simulations running (with different environments).

10 environments requires at least 10 JSON node config files, even though we might only have 2 actual agent nodes to play.

I think we might consider changing our JSON configuration and introduce the concept of an environment configuration into the domain, which is decoupled from the agent/node configuration.

cprice404:
@nwolfe sorry, this is probably a dumb question but I'm way out of date on what's been going on.

How does this relate (if at all) to the goal we had about deploying the ops modules for a test?  Because they would include a ton of environments, and, yes, I thought we'd agreed that we would need to have some alternate mechanism besides the agent/node configuration file to get that code deployed.  Like a directory of 'r10k-prereqs' json files or something?

nwolfe:
> How does this relate (if at all) to the goal we had about deploying the ops modules for a test?

@cprice Ahh, sorry for the confusion.  I was working towards a walking skeleton, not a specific deployment that we wanted to test.  I saw this ticket about recording a multi-environment scenario in the Gating epic and thought that now was the time for it, considering that we have the initial environment support and need for a more real-world scenario to drive the rest of the implementation.

cprice404:
@nwolfe that's fine, it'll still be valuable eventually.

I'm not going to have time to review this for at least a week, so if you and @jpinsonault are happy with it, feel free to just merge it.

jpinsonault:
:+1: no complaints

PR #37, opened 2015-11-16
(PE-12968) Record PE 2015.2.3 baseline with catalog zero

haus:
This commit adds a recording of an agent run with the transformations
described in
https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/simulation-runner/README-GENERATING-AGENT-SIMULATIONS.md
applied to it. The run was between a 2015.2.3 master (bl14) and an agent
(vmpooler), both centos 6 x86_64.


haus:
I suppose I maybe should rename this to PE201523CatalogZero to be consistent with the other node_simulations. It will sort oddly in a file listing, which is unfortunate.

camlow325:
Other than that a number of "file_content" requests appeared in the output, this all looks good to me.  We expect agents to make "file_content" requests the first time they download files from the server but for follow-up runs where the agent had cached the content of the downloaded files, only "file_metadata" requests would be expected.  Since this capture needs to reflect the requests that an agent would make to the master in "steady-state", would be good to recapture the requests with no "file_content" requests being made.  (If the agent does make the "file_content" requests across repeated runs, might have a caching problem on the agent, not sure?).

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PE201523CatalogZero.scala
```diff
@@ -0,0 +1,380 @@
...
+		.exec(http("filemeta")
+			.get("/puppet/v3/file_metadata/modules/catalog_zero29/catalog_zero29_impl86.txt?environment=production&links=manage&checksum_type=md5&source_permissions=ignore")
+			.headers(headers_0))
+		.exec(http("filemeta")
+			.get("/puppet/v3/file_metadata/modules/catalog_zero30/catalog_zero30_impl34.txt?environment=production&links=manage&checksum_type=md5&source_permissions=ignore")
+			.headers(headers_0))
+		.exec(http("filemeta")
+			.get("/puppet/v3/file_metadata/modules/catalog_zero30/catalog_zero30_impl85.txt?environment=production&links=manage&checksum_type=md5&source_permissions=ignore")
+			.headers(headers_0))
+		.exec(http("filemeta")
```
camlow325: I think we've been renaming this one as "filemeta mco plugins".

camlow325:
:+1: with just the one minor note on the name of the request used for the mco plugins filemeta.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PE201523CatalogZero.scala
haus: ah yes, my script thing mislabeled that one

haus: updated

KevinCorcoran:
Thanks a lot for reviewing.  probably should just merge.  I looked at it a bit ... other things are blocked until.

camlow325:
:shipit:

PR #39, opened 2015-11-16
(PE-12968) Add node config and scenario configs for PE 2015.2.3

haus:
This commit adds a node config and some scenarios for the PE 2015.2.3
recording.


KevinCorcoran:
:see_no_evil:

simulation-runner/config/nodes/pe201523-catalogzero.json
```diff
@@ -0,0 +1,6 @@
+{
+    "certname": "w3h244u9nwii0y5.delivery.puppetlabs.net",
+    "simulation_class": "com.puppetlabs.gatling.node_simulations.PE201523CatalogZero",
```
nwolfe: I don't see this class in this PR - should this be one of the `WithFeeder` classes?

haus: Don't worry, already merged. It's here: https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PE201523CatalogZero.scala

nwolfe:
:+1:

PR #40, opened 2015-11-16
(maint) Add a simple script to automate part of the recording rewriting

haus:
Rewriting the request portions of the agent run into the various
mappings can be quite tedious, so this commit adds a script to automate
that step in the process.

### DIFF
--- a/proxy-recorder/process_gatling_recording.rb
+++ b/proxy-recorder/process_gatling_recording.rb
@@ -1,0 +1,163 @@
+#!/usr/bin/env ruby
+
+TERMINOLOGY = {
+  "catalog"                                                 => "catalog",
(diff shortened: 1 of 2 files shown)


haus:
Updated.

KevinCorcoran:
:+1: I used it.  It worked.  I liked it better when it had more 11's.

nwolfe:
:+1: This is awesome!  

We should probably mention this script exists in the corresponding README for these steps (https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/simulation-runner/README-GENERATING-AGENT-SIMULATIONS.md).

haus:
@nwolfe updated

nwolfe:
:metal:

PR #43, opened 2015-11-17
Update gatling feeder docs

haus:
This commit fixes a couple of typos and clarifies a few things. It also
removes a suggestion to use mkString on the request body file. This
should no longer be required with the current recommended post-recording
processing to move the request body into an ELFileBody.


fpringvaldsen:
:+1:

PR #45, opened 2015-12-09
(SERVER-852) Partial r10k support

nwolfe:
This commit adds the beginnings of configuring the master via an r10k
control repository. It goes so far as to run "r10k deploy" to map the
repository branches to Puppet environment directories, but does not do
anything after that like wire up any hiera.yaml configuration.

### DIFF
--- a/jenkins-integration/beaker/install/shared/r10k_deploy.rb
+++ b/jenkins-integration/beaker/install/shared/r10k_deploy.rb
@@ -1,0 +1,36 @@
+require 'puppet/gatling/config'
+
+test_name "Deploy r10k control repository"
+
+## TODO This probably needs more work to finish hooking up everything that
+##      comes in the control repository, like a hiera.yaml config file.
+##
+##      Also, r10k will manage the entire environment directory, which means
+##      previous gatling installation steps (e.g. 50_install_modules.rb) may
+##      be overridden. For example, any modules defined in the JSON node files
+##      that aren't defined in the r10k control repo will be removed.
+
+def install_r10k(host)
+  gem = '/opt/puppetlabs/puppet/bin/gem'
+  on(host, "#{gem} install r10k --no-document")
+end
+
+def create_r10k_config(host, r10k)
+  configfile = r10k_configpath(r10k)
+  configdir = '/etc/puppetlabs/r10k'
+  on(host, "mkdir -p #{configdir}")
+  scp_to(host, configfile, "#{configdir}/r10k.yaml")
+end
+
+def run_r10k_deploy(host)
+  r10k = '/opt/puppetlabs/puppet/bin/r10k'
+  on(host, "#{r10k} deploy environment --puppetfile --verbose")
+end
+
+scenario_id = ENV['PUPPET_GATLING_SCENARIO']
+r10k_config = parse_scenario_file(scenario_id)['r10k']
+if r10k_config
+  install_r10k(master)
+  create_r10k_config(master, r10k_config)
+  run_r10k_deploy(master)
+end
--- a/jenkins-integration/lib/puppet/gatling/config.rb
+++ b/jenkins-integration/lib/puppet/gatling/config.rb
@@ -7,6 +7,7 @@ require 'yaml'
 ## 3. Node config JSON files in "./config/nodes/*.json"
 ## 4. Hiera config YAML files in "./config/hieras/<hiera>/hiera.yaml"
 ## 5. Hiera data trees in "./config/hieras/<hiera>/<datadir(s)>/"
+## 6. r10k config YAML files in "./config/r10ks/<r10k>.yaml"
 
 def parse_scenario_file(scenario_id)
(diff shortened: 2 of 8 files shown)


jenkins-integration/beaker/install/shared/r10k_deploy.rb
```diff
@@ -0,0 +1,36 @@
...
+end
+
+def run_r10k_deploy(host)
+  r10k = '/opt/puppetlabs/puppet/bin/r10k'
+  on(host, "#{r10k} deploy environment --puppetfile --verbose")
+end
+
+scenario_id = ENV['PUPPET_GATLING_SCENARIO']
+r10k_config = parse_scenario_file(scenario_id)['r10k']
+if not r10k_config.nil?
```
haus: `if r10k_config` is probably more idiomatic ruby

jenkins-integration/lib/puppet/gatling/config.rb
```diff
@@ -65,3 +66,8 @@ def hiera_datadirs(hiera)
     [localpath, datadir]
   end
 end
+
+# Returns the path to the local r10k YAML configuration file
+def r10k_configpath(r10k)
+  File.join('config', 'r10ks', r10k + '.yaml')
```
haus: `r10ks` => `r10k`?

jenkins-integration/beaker/install/shared/r10k_deploy.rb
```diff
@@ -0,0 +1,36 @@
...
+  scp_to(host, configfile, "#{configdir}/r10k.yaml")
+end
+
+def run_r10k_deploy(host)
+  r10k = '/opt/puppetlabs/puppet/bin/r10k'
+  on(host, "#{r10k} deploy environment --puppetfile --verbose")
+end
+
+scenario_id = ENV['PUPPET_GATLING_SCENARIO']
+r10k_config = parse_scenario_file(scenario_id)['r10k']
```
haus: It looks like parse_scenario_file will blow up if scenario_id is nil. That might be worth checking before calling the method.

nwolfe: Yep, that's true in every single one of these install/ scripts.  There's room for improvement certainly.  There isn't that much guarding in these scripts because ultimately they'll be called by a higher-level script/Jenkins job where we'll probably do more of that.

jenkins-integration/lib/puppet/gatling/config.rb
nwolfe: `r10ks` is a directory full of r10k configuration files, following the existing pattern of `scenarios/`, `nodes/`, and `hieras/`

jenkins-integration/beaker/install/shared/r10k_deploy.rb
nwolfe: Ah yes.

jenkins-integration/lib/puppet/gatling/config.rb
haus: ah, cool

nwolfe:
Yeah at some point I need to document how the hiera and r10k support fits together between the JSON files and the `hieras/` and `r10ks/` directories.

So far I've just been including examples of each but it's not exactly obvious that arbitrary strings need to match filenames under specific directories.

jenkins-integration/scripts/foss_install.sh
```diff
@@ -1,6 +1,22 @@
...
+##
+## 2. Version of puppet-server package to install
+##    Can be specified with either:
+##    - ENV['PUPPETSERVER_BUILD_VERSION']=2.2.1
+##    - hosts.yaml#CONFIG#puppetserver_version=1.3.0
+
+if [ -z $BEAKER_CONFIG ]; then
+    echo "BEAKER_CONFIG hosts configuration required"
+    exit 1
+fi
```
cprice404: This doesn't seem related to the r10k stuff?

cprice404:
@nwolfe I know this isn't complete, but it seems like it's basically a no-op unless someone puts in r10k stuff, so it seems like maybe we should just go ahead and merge it?  Maybe we should add a comment / log message somewhere that explains that it's not fully functional yet?

jenkins-integration/scripts/foss_install.sh
nwolfe: It's not, but during development I run this file directly, and having this fail-fast is useful.

nwolfe:
> I know this isn't complete, but it seems like it's basically a no-op unless someone puts in r10k stuff, so it seems like maybe we should just go ahead and merge it?

Yep, that's what I was thinking.

The `r10k_deploy.rb` file is for the most part complete (at least in terms of handling the r10k-specifics), it's just not wired up in the foss/pe_install files yet.

There's still the outstanding question of how to integrate hiera data from the r10k control repository, which we'll be avoiding for the specific OPS deployment scenario.

I'll add some comments about r10k and address other feedback.

nwolfe:
Feedback addressed.  Ready for merge.

PR #46, opened 2015-12-16
[For Review Only] (SERVER-1050) Fix clumping issue

jpinsonault:
Putting this up for comment. I'm doing a long run with this code to see how successful it is, but I figured people might want to look at it in the mean time.

### DIFF
--- a/simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
+++ b/simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
@@ -16,6 +16,7 @@ import scala.concurrent.duration._
 class ConfigDrivenSimulation extends Simulation {
 
   val REPETITION_COUNTER: String = "repetitionCounter"
+  val SLEEP_DURATION: String = "nextSleepDuration"
 
   def addSleeps(chain:ChainBuilder, sleepDuration:Int, totalNumReps:Int): ChainBuilder = {
     // This is kind of a dirty hack. Here's the deal.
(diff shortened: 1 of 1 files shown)


simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
```diff
@@ -31,16 +32,40 @@ class ConfigDrivenSimulation extends Simulation {
...
       val repetitionCount = session(REPETITION_COUNTER).asOption[Int].getOrElse(0) + 1
-      println("Agent " + session.userId +
-        " completed " + repetitionCount + " of " + totalNumReps + " repetitions.")
-      session.set(REPETITION_COUNTER, repetitionCount)
+
+      // Calculate the difference between when the next start time should be,
+      // and now.
+      // This has the bonus that the start times will never drift from run to
+      // run. They will always be ((a multiple of sleepDuration) + the start time)
+      val timeUntilNextRun = (session.startDate + sleepDuration * 1000 * repetitionCount) - System.currentTimeMillis()
```
haus: is startDate baked into the session via gatling?

jpinsonault: Yeah, it's defined in the built in gatling `Session` class.

cprice404:
@camlow325 did you do some before/after with this PR that convinced you that it fixed the clumping?

@jpinsonault how long did your recent run take before it started crapping out?  Just so that we know how long we might need to wait for the new run to go, if we're gating this on it.

simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
```diff
@@ -31,16 +32,40 @@ class ConfigDrivenSimulation extends Simulation {
...
+    //
+    // In addition, simply sleeping for sleepDuration after a run causes issues
+    // long runs, where start/end times will drift, and agent runs will start to
+    // clump together. To solve this, the length of the pause is calculated to
+    // be the difference between (start time + (repetition * sleep duration))
+    // and the current time.
+    //
+    // That is, if the sleep duration is 30 seconds, and a run takes 5 seconds,
+    // the agent will sleep for 25 seconds
+    chain.exec((session) => {
```
cprice404: Any reason to remove the type hint from the arg list here?

cprice404:
+1 from me, looks good... nice find on `startDate`, @jpinsonault !

Maybe we wait to merge until we either get confirmation from @jpinsonault 's latest run, or from runs that @camlow325 may have done, but otherwise I'm +1.

simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
jpinsonault: Just consistency with the rest of the code. I could also add it into the other `exec` calls. Not sure what's more idiomatic

jpinsonault:
@cprice404 It crapped out after ~3 days

simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
cprice404: gotcha.  I prefer explicit in general so I'd probably add it to the other spots, but it's not a big enough deal to hold up the PR for or anything.

cprice404:
@jpinsonault ok thx, we can sit on this for a few more days then.

camlow325:
> @camlow325 did you do some before/after with this PR that convinced you that it fixed the clumping?

Yeah, these changes appeared to work great for the couple of Gatling runs that I used them with.  I used these as part of the work to validate a JRuby 9k port - see [SERVER-1064](https://tickets.puppetlabs.com/browse/SERVER-1064).

For the first few runs I did without the changes in this PR, I saw the typical signs of "clumping" - large spikes in requests per second followed by quick drops back down toward 0 late in the run.  With the changes in this PR included, though, I saw a fairly flat rate of requests per second throughout runs with exactly the same parameters against the same hardware.  Here's some output I captured from one of the runs with the changes in the PR applied:

http://int-resources/puppetserver%20gatling%20reports/server-1064-pe2015.3.0-jruby-9k-1000-8hr.tar.gz

These runs only went for 8 hours total, with the following Gatling simconfig:

~~~~json
{
    "run_description": "PE 2015.3.0 Catalog Zero - 1000 agents, 30 minute ramp, 16 times",
    "nodes": [
        {
            "node_config": "pe201530-catalogzero-with-feeder.json",
            "num_instances": 1000,
            "ramp_up_duration_seconds": 1800,
            "num_repetitions": 16,
            "sleep_duration_seconds": 1800
        }
    ]
}
~~~~

I can't speak to how sane the code changes are but that all looked good from my perspective.  Will defer to what @jpinsonault finds out after investigating his longer runs if we think that makes sense.

jpinsonault:
The reports directory filled up and puppetserver stopped responding to requests, but after 4.5 days it showed no signs of clumping

cprice404:
Calling this good for now.

PR #47, opened 2016-02-11
(SERVER-854) Dev/ improvements; start of OPS job

nwolfe:
This giant commit contains many fixes to the existing dev/ scripts so
they actually work properly again (they're still very specific to CentOS
6 though).

This also contains various adjustments to the new automation steps we
have in jenkins-integration, as this is the first time we're actually
trying to use them. Some reconciliation between new and old was
necessary, and there are certinaly more improvements that can be made.

Lastly, this commit contains the beginnings of the OPS deployment JJB
job. The JJB job uses the new jenkins-integration automation steps and
defines it's own specific step to clone the new
test-catalogs-large-files repository and untar the OPS environments into
code-staging.

The OPS job is not yet complete, and will likely need updates to the
ops-scenario.json and ops-node.json files as we continue development.


nwolfe:
I'm sorry.

jenkins-integration/scripts/pe_install.sh
```diff
@@ -1,34 +0,0 @@
-#!/bin/bash
```
nwolfe: I deleted these two `foss_install.sh` and `pe_install.sh` scripts because they weren't used, and after working on this OPS job I'm not sure that they will be useful as they're currently written since we need to use the ruby193 shim and define a very specific set of beaker --test files to execute.

We could probably re-create these files in some form after we've written a few more jobs and could see what/how to extract the installation.  For the time being, the code in `setup_master.sh` is a better template to follow for any new JJB jobs we define.

jenkins-integration/scripts/run_sbt.sh
```diff
@@ -3,14 +3,12 @@ set -x
 
 export BEAKER_TESTSUITE="${BEAKER_TESTSUITE:-beaker/gatling/20_run_sbt.rb}"
 export BEAKER_KEYFILE="~/.ssh/id_rsa-acceptance"
-export BEAKER_HELPER="beaker/helper.rb"
```
nwolfe: Don't use this snowflake helper file.  Put code into either `jenkins-integration/lib/puppet/gatling/config.rb` or create new namespaces in that directory, and then explicitly reference them using `require 'puppet/gatling/config'` like normal Ruby.

We've never used this helper.rb (and it was empty) so I deleted in and all references to it.

simulation-runner/config/scenarios/ops-scenario.json
```diff
@@ -0,0 +1,12 @@
+{
```
nwolfe: This file is likely incomplete and will need to be updated once we're able to get the scala simulation further in its execution.

We'll probably need to change the `num_instances`, etc values to something more interesting for our OPS job.

simulation-runner/config/nodes/ops-node.json
```diff
@@ -0,0 +1,12 @@
+{
```
nwolfe: This file is likely incomplete and will need to be updated once we're able to get the scala simulation further in its execution.

Particularly, I'm not sure if the `simulation_class` will work or if we need to make a new PE recording, or maybe use the `WithFeeder` version.  We may want to use a different catalog zero module too; I'm not sure that I got far enough to know if this is the correct one.

nwolfe:
Ping @cprice404. I've left some explanation comments both in the code and here in the PR.

Associated JIRAs are SERVER-1144 and SERVER-1145.

Sorry this is all one big commit -- everything was changing all the time and clean commits broke down pretty quick.  I'm happy to sit down and go through these changes in person.

cprice404:
@nwolfe I'm somewhat inclined to just merge this and then deal with any issues once the next person picks up on this work.  WDYT?

Also, thanks a ton for all the work writing up docs and notes and stuff!

nwolfe:
> I'm somewhat inclined to just merge this and then deal with any issues once the next person picks up on this work. WDYT?

@cprice404 Yep, that makes sense.  Happy to help out if/when issues come up.

PR #48, opened 2016-02-19
(MAINT) Update example to use test-catalogs

nwolfe:
Instead of my personal nwolfe/puppet-loadtest module that is old,
irrelevant, and replaced by puppetlabs/test-catalogs.


cprice404:
could you remove the '1's from these and just use `catalog_zero`?

nwolfe:
@cprice404 That's better. Updated.

PR #49, opened 2016-03-04
(MAINT) add docs re: tk auth and disks on perf hardware

cprice404:


fpringvaldsen:
:see_no_evil:

cprice404:
@nwolfe would you mind taking a quick gander at this

README_tk_auth.md
```diff
@@ -0,0 +1,49 @@
...
+            sort-order: 1
+            name: "allow all"
+        },
+```
+
+After this, it's a good idea to do an agent run with `puppet agent -t` to verify that PE doesn't think it needs to modify the `auth.conf` file.
+
+After verifying that, you will want to restart or HUP the server so that the new rules take effect.  You can test the rules by running a command like this:
+
+   curl -k https://localhost:8140/puppet/v3/catalog/<certname-of-this-server>\?environment=production
```
nwolfe: This isn't rendering correctly in Markdown. I think the `<certname-of-this-server>` problematic?  We can probably just change that part to `$(hostname)` like we have in other parts of the documentation.

nwolfe:
:+1:

cprice404:
@nwolfe added a commit to fix that, thx

nwolfe:
Hm, not sure why it's not rendering properly still.  Tabs vs spaces?  Not enough spaces?  Oh well, merging

cprice404:
seems to render ok for me?  who knows.  computers--

PR #53, opened 2016-03-16
Feature/master/non csv feeder

cprice404:
This commit introduces a feeder class in scala code, so that we don't need CSV files for all of the various numbers of feeders that we might want to use.

I think that it might break any existing sims that are explicitly using feeders via csv, but not 100% sure whether that is true, or whether I care :)

### DIFF
--- a/simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
+++ b/simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
@@ -3,7 +3,7 @@ package com.puppetlabs.gatling.runner
 import io.gatling.core.Predef._
 import io.gatling.http.Predef._
 
-import com.puppetlabs.gatling.config.PuppetGatlingConfig
+import com.puppetlabs.gatling.config.{NodeFeeder, PuppetGatlingConfig}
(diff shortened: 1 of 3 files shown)


simulation-runner/src/main/scala/com/puppetlabs/gatling/runner/ConfigDrivenSimulation.scala
```diff
@@ -101,7 +101,10 @@ class ConfigDrivenSimulation extends Simulation {
     val chainWithSleeps:ChainBuilder =
       addSleeps(chainWithFailFast, sleepDuration, numRepetitions)
 
+    val feeder = NodeFeeder("node", numInstances).circular
```
haus: `num_instances` from a scenario gets converted into `numInstances` here somewhere along the way?

haus: ah. found it. https://github.com/puppetlabs/gatling-puppet-load-test/blob/34f30e164580933d6f216d56f878ed2fb193c987/simulation-runner/src/main/scala/com/puppetlabs/gatling/config/PuppetGatlingConfig.scala#L23

haus:
this is awesome. :100:

jpinsonault:
Very cool :+1:

PR #55, opened 2016-04-02
(MAINT) Add a mechanism for generating reports via sbt

cprice404:


nathanielksmith:
:+1: but given that I have basically never looked at scala or done gatling, someone else should probably merge this.

PR #62, opened 2016-06-02
(MAINT) Add some catalog-zero sims for Couch

camlow325:
This PR adds a couple of catalog-zero agent simulations for Couch.  One of the simulations uses PSON as the wire serializaton format - which is the default used for most (all?) other simulations stored in this repo already.  This PR also includes a simulation taken with MSGPACK enabled as the wire serialization format for both the agent and master.


cprice404:
In the future, I think it's really valuable to have a commit that just contains the raw recorder output, followed by a second commit that overlays your changes.  It's useful for validating your modifications but also in case we ever wanted to revert to the raw recording and overlay a different set of changes.

camlow325:
Good call, thanks.  Will try to remember to do that next time.

PR #63, opened 2016-07-06
(SERVER-1376, PE-15655) Switch to jenkins pipeline

cprice404:

* Attempts to improve organization of a lot of the scripts / beaker code files on disk.
* Moves us from a JJB-based mechanism for managing jobs over to an approach that uses
  Jenkins' JobDSL plugin.
* Creates a seed job via JobDSL that can find perf test scenarios, defined by `Jenkinsfile`s,
  and create Jenkins pipeline jobs based on them.
* Defines some library code in `pipeline.groovy` that can be re-used to create different
  perf test jobs / scenarios.
* Defines two sample/template jobs: single-pass and multi-pass, which provide examples of
  how to run a single perf test job broken into lots of stages and a multi-pass job that
  runs two gatling sims serially.


jenkins-integration/beaker/install/pe/98_sync_codedir.rb
```diff
@@ -26,6 +26,30 @@
...
+# curl = <<EOS
+# curl -s -X POST -k
+# EOS
+
 # TODO: improve this
 # sleep for 10 minutes to make sure the file sync has completed
-sleep 600
+# step "Sleeping 600 to wait for file sync to complete; should change this to monitor the file sync status endpoint"
+# sleep 300
+# step "Done sleeping"
```
rlinehan: There's still some comments and TODOs in here - is that intended?

jenkins-integration/jenkins-jobs/common/scripts/job_steps/040_install_puppet_code.sh
```diff
@@ -1,20 +1,24 @@
...
 # - CatalogZero module added to production environment
-# - Node classified via NC to have catalog zero module
 #
 # Note that we only need the OPS environments installed, we don't need to
 # explicitly reference them or use any of their classes; simply having them
 # there will exhibit the slow performance behavior we're trying to highlight
 # with this job.
 
+# TODO TODO TODO: this should *NOT* be hard-coded in the job template,
+#  this is very specific to a given job.
```
rlinehan: what is the "this" here? (I'm not quite sure what this comment is referring to)

jenkins-integration/jenkins-jobs/job_bootstrap.groovy
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Paths
+
+// NOTE: these determine the default repo/branch that the seed job will
+// poll to look for Jenkinsfiles.  For production they should always
+// be set to the PL gplt repo's master branch.  For dev, you may want
+// to change them to point to your fork or your git daemon (though you can
+// accomplish the same by simply modifying the values in the Jenkins gui
+// on the 'configuration' screen for the seed job.
+
+def git_repo = 'https://github.com/puppetlabs/gatling-puppet-load-test.git'
```
rlinehan: There's a few places where you might end up wanting to change repo name/branch name/environment name for doing development. It seems like it might be helpful to have a comment (like `CHANGE ME FOR DEVELOPMENT` or something better named/worded) that you could easily grep for to find all these places?

jenkins-integration/jenkins-jobs/job_bootstrap.groovy
```diff
@@ -0,0 +1,71 @@
...
+scenarios_dir.eachFileRecurse (FileType.FILES) { file ->
+    if (file.name.equals("Jenkinsfile")) {
+        job_prefix = file.parentFile.name
+        relative_jenkinsfile = relativize(root_dir, file)
+        workflowJob(job_prefix) {
+            // TODO: this should be moved into the Jenkinsfile by use of
+            // the 'properties' step, see https://issues.jenkins-ci.org/browse/JENKINS-32780
+            parameters {
+                stringParam('SUT_HOST',
+                        'puppetserver-perf-sut54.delivery.puppetlabs.net',
```
rlinehan: Is this a thing that will eventually not be hardcoded?

rlinehan:
👍. I left a couple small comments; none of them are terribly important, and so @cprice404 if you don't get a chance to respond to them by tomorrow morning I'll merge this anyway.

rlinehan:
Oh, also I should have said - I ran this locally and it all looked good. Got the Jenkins instance up, ran the seed job and got the additional jobs, currently running the multi-pass scenario (it failed the first time for me because I had forgotten about running the `add_public_key` script).

jenkins-integration/beaker/install/pe/98_sync_codedir.rb
cprice404: intended-ish?  I'm still not 100% convinced that the code for file syncing is sufficient - was leaving some of this in case we needed to revert something, or find another way to deal with that race condition you hit, etc.

jenkins-integration/jenkins-jobs/common/scripts/job_steps/040_install_puppet_code.sh
cprice404: Referring to this file, really.  The current implementation of deploying the code is doing a very special / unusual thing that downloads an old snapshot of the ops modules from a tarball, because it can take over an hour to deploy it from r10k.  Ultimately this file needs to be generalized to support some more standard methods of deploying code.  I expect we'll need to do that in one of the next few tickets.

jenkins-integration/jenkins-jobs/job_bootstrap.groovy
cprice404: Yeah, or even figure out a way to consolidate those to a single data file or something.  Was thinking we could maybe get a little more mileage on it and see if a solution presents itself?

jenkins-integration/jenkins-jobs/job_bootstrap.groovy
cprice404: Possibly.  Ultimately what I want is to find some kind of Jenkins plugin for "locking" resources, and put our three SUT names into that, and have our jobs block unless one of the SUTs is available... and when one is, grab it.  In the meantime we may end up with certain jobs defaulting to certain SUTs, though.

cprice404:
I think all of the comments you raised are valid, thanks for the thorough review.  At the moment I'm feeling like enough is still in flux in this repo that most of those will keep bubbling up and we can address them in later PRs, but if there are any that you feel strongly about I'm happy to patch them up now.

rlinehan:
My multi-pass scenario job finally worked! I don't feel strongly about any of my comments, and agree that there's enough in flux to wait on them. So 🚢!

PR #65, opened 2016-08-04
(SERVER-1450) r10k deploy

cprice404:
This PR adds support for specifying some r10k metadata in your job description, and if you do so, it will cause the specified control repo to get deployed onto the SUT before running tests.


jenkins-integration/beaker/install/shared/r10k_deploy.rb
```diff
@@ -15,21 +18,65 @@ def install_r10k(host)
...
+
+########################
+
+step "Install git" do
+  on(master, puppet_resource("package git ensure=installed"))
+end
+
+step "Set up SSH key for github access" do
+  if !master.file_exist?("/root/.ssh/id_rsa")
+    result = curl_on master, "-o /root/.ssh/id_rsa 'http://int-resources.ops.puppetlabs.net/QE%20Shared%20Resources/travis_keys/id_rsa'"
```
camlow325: Wow, we have that on a share that doesn't even require 'https'?  Huh.  That's cool.

jenkins-integration/beaker/install/shared/r10k_deploy.rb
```diff
@@ -15,21 +18,65 @@ def install_r10k(host)
...
+
+    result = curl_on master, "-o /root/.ssh/id_rsa.pub 'http://int-resources.ops.puppetlabs.net/QE%20Shared%20Resources/travis_keys/id_rsa.pub'"
+    assert_equal 0, result.exit_code
+  end
+end
+
+step "add github to known hosts" do
+  # Create known_hosts file with GitHub host key to prevent
+  # "Host key verification failed" errors during clones
+  create_remote_file(master, "/root/.ssh/known_hosts", <<-EOS)
```
camlow325: Is this the only known host that the process needs?  Just wondering if we should append to whatever might already be there rather than just replacing it entirely?

jenkins-integration/beaker/install/shared/r10k_deploy.rb
```diff
@@ -15,21 +18,65 @@ def install_r10k(host)
...
+  github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==
```
cprice404: > Is this the only known host that the process needs? Just wondering if we should append to whatever might already be there rather than just replacing it entirely?

That's a good question.  Not sure how to do that in beaker.  Would be willing to change it now if @erikPrime or @kurtwall or someone has an idea of how to do that, or leave it alone until something breaks (though that could be confusing to debug)

kurtwall: I don't recommend appending to pre-existing files. It would leave (probably harmless) state behind for subsequent tests. The usual idiom is to make a backup of the target file, drop yours into place, then restore the backup in your teardown procedure.

camlow325:
I looked through the code in this PR and 🙈.  I didn't have time to fully test this out via the dev environment.  Since I'm out of the office tomorrow, I'd be okay with this being merged if it's going to end up blocking other work / making future rebasing too complicated.  Otherwise, I'll finish the dev environment tests on Monday and hopefully merge then.

camlow325:
I tried this out with the "dev" environment.  The "single-pass-scenario" worked for the most part.  The r10k code deploy (referencing the perf_control repo) - which is what I think this PR was mostly about - all worked fine.  In the "ops-scenario.json" file (and backing "ops-node.json") file, there are still references to "cat-zero".  These cause agent classification to fail because the actual classes installed onto the SUT are from the perf_control repo instead of catalog_zero.  @cprice404 mentioned that later PRs correct the scenario setup to use perf_control artifacts for classification.

Since everything else basically seemed to work, though, I'm going to go ahead and merge this one.

PR #66, opened 2016-08-05
(SERVER-1463) fix auth so gatling can connect to server

cprice404:
This commit adds back in some calls to some code that will set up permissive auth.conf on the server and then restart it, so that Gatling can connect without getting 403 errors.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/job-steps/020_install_pe.sh
+++ b/jenkins-integration/jenkins-jobs/common/scripts/job-steps/020_install_pe.sh
@@ -23,7 +23,9 @@ bundle exec beaker \
         --tests \
 beaker/install/shared/hack_hostname_into_etc_hosts.rb,\
 beaker/install/shared/disable_firewall.rb,\
-beaker/install/pe/10_install_pe.rb
+beaker/install/pe/10_install_pe.rb,\
+beaker/install/shared/configure_permissive_server_auth.rb,\
+beaker/install/pe/99_restart_server.rb
 
 # without this set +x, rvm will log 10 gigs of garbage
 set +x


jenkins-integration/jenkins-jobs/common/scripts/job-steps/020_install_pe.sh
```diff
@@ -21,9 +21,11 @@ bundle exec beaker \
...
         --no-color \
         --tests \
-jenkins-jobs/common/scripts/ops-deployment/hack_hostname_into_etc_hosts.rb,\
-jenkins-jobs/common/scripts/ops-deployment/disable_firewall.rb,\
-beaker/install/pe/10_install_pe.rb
+beaker/install/shared/hack_hostname_into_etc_hosts.rb,\
+beaker/install/shared/disable_firewall.rb,\
+beaker/install/pe/10_install_pe.rb,\
+beaker/install/shared/configure_permissive_server_auth.rb,\
+beaker/install/pe/99_restart_server.rb
```
camlow325: Not directly related to changes in this PR, but we may want to use the beaker `bounce_service` function here inside of "99_restart_server.rb", which uses `puppet resource <service> ensure=stopped|running` under the hood, instead of `service pe-puppetserver restart`.  I've heard that some OSes - maybe SLES? - don't support `service <service name> restart` whereas I think the `puppet resource <service> ensure=stopped|running` approach works more broadly.

cprice404: Sure, that's probably a good suggestion.  For now I don't have any plans for this to support any SUT OS's besides cent7 but we can definitely improve / broaden that in the future.

camlow325:
#65 has been merged so I think this can be rebased now.

camlow325:
I tried this out with a "dev" environment.  The auth.conf reconfiguration appears to work as expected.  I verified manually that the super-permissive auth.conf file was laid down and that, without needing to restart the puppetserver service, I could successfully do a curl request to the catalog endpoint with no client certificate.  I the "single-pass-scenario" job failed again as with #65 because catalog-zero classification is inappropriately attempted even though perf-control environment content is what is actually dropped in place for the scenario.  That should be addressed in a later PR, though, so I don't think it should block merge of this PR.

I can go ahead and merge this once it has been rebased on top of changes from #65.

cprice404:
@camlow325 rebased

camlow325:
:shipit:

PR #67, opened 2016-08-09
(SERVER-1464) support multiple pe versions

cprice404:
This PR provides initial support for specifying a specific version of PE in your job definition.  In order to support multiple PE versions it also adds in some logic for conditioning on what PE features are available (file sync, AIO paths, etc.)


cprice404:
rebased.

jenkins-integration/beaker/install/shared/r10k_deploy.rb
```diff
@@ -73,9 +80,22 @@ def run_r10k_deploy(host, r10k_config)
...
+    # exit code even when they've actually successfully deployed.  If you then
+    # simply re-run the same r10k command afterward, it'll get a zero exit code.
+    # So, we will do an initial deploy and then if we got any non-zero exit codes,
+    # we'll try those environments one more time.  Sorry.
+    results = run_r10k_deploy(master, bin, r10k_config[:environments],
+                              {:acceptable_exit_codes => [0, 1]})
+    failed_envs = results.select { |env,exit_code| exit_code == 1 }.keys
+    if failed_envs.size > 0
+      Beaker::Log.warn("R10K deploy failed on environments: #{failed_envs}; trying one more time.")
+      run_r10k_deploy(master, bin, failed_envs, {})
```
camlow325: If it fails after running a second time, should it cause the beaker script to blow up?

cprice404: I think it should, WDYT?

camlow325: Yeah, I think it should do.  Currently, though, it would just move on without anything in the job failing, right?

cprice404: No, it will fail because of the exit code from the remote process, right?

camlow325: Ah, yeah, I think you're right.  My bad...

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -12,31 +61,42 @@ def step000_provision_sut(SKIP_PROVISIONING, script_dir) {
...
 
-def step010_setup_beaker(script_dir) {
-    withEnv(["SUT_HOST=${SUT_HOST}"]) {
+def step010_setup_beaker(script_dir, server_version) {
+    if (server_version["type"] != "pe") {
+        error "Unsupported server type: ${server_version["type"]}"
+    }
+
+    withEnv(["SUT_HOST=${SUT_HOST}",
+             "pe_version=${server_version["pe_version"]}",
```
camlow325: In pe-p-s-e jobs, I've seen values set like:

* pe_family = 2016.3
* pe_version = 2016.3.0-rc3-11-gbbfedf2

Will that work with how it looks like these jobs currently appear to set those to the same value?

cprice404: It will work like that if/when we extend the data structure to support it, yes.

camlow325: Ok, will that mean that for now that we can't install a non-released (i.e., mergely) version of PE?  Sorry, I don't really understand how these variables are used.

cprice404: I believe that is correct but I was planning on digging into the beaker docs later to sort that out.  I'm fine with just released versions for this pass.

camlow325: Sounds good.

camlow325:
Per the changes in this PR, it appears that a server_version section would now be required in each of the entries passed into `pipeline.single_pipeline()` and `pipeline.multipass_pipeline`.  Those entries exist in the `Jenkinsfile` for the new `compare-3.3-to-couch` but do not in the `Jenkinsfile` for the previously existing `single-pass-scenario` or `multi-pass-scenario`.  I think it should be added to the other `Jenkinsfile` entries so that those jobs will continue to work.  Presumably, we'd just do a PE 2016.2.0 configuration for each of those?

cprice404:
I agree - would you be OK with doing that in a follow-up PR?

camlow325:
Overall 👍.

I tried this out in a "dev" environment with the new "compare-3.3-to-couch" job.  The job seemed to work fine up to the point where it tried to kick off the gatling run - on both PE 3.3 and PE 2016.2.0.  In both cases, the gatling job failed because PE had been setup to classify and play simulations for nodes for catalogs containing catalog_zero resources / classes whereas the actual resources / classes being deployed were from the perf_control repo.  I believe this will still be addressed in a later PR.

It would also be good to correct the `single-pass-scenario` and `multi-pass-scenario` to use the new `server_version` check but this can be done in a separate PR as well.

I'm going to go and ahead and merge this to unblock additional work.

PR #68, opened 2016-08-09
(SERVER-1473) Update docs re: feeders

cprice404:
This commit removes some old documentation and tooling around using
'feeders' to generate unique node names, and updates the remaining
docs with more up-to-date information about how to use feeders.


camlow325:
LGTM.  :shipit:

PR #69, opened 2016-08-10
(SERVER-1469) finish scala conversion script

cprice404:
This PR extends the script for post-processing gatling recordings to do the following:

* Complete *all* of the processing steps listed in the docs, rather than just a subset.  Theoretically this should mean that after taking a raw recording all you need to do is run this script and then you're done.
* Finds and modifies the request body text file in addition to the scala file
* Looks for a 'node config' JSON file and generates one if one doesn't exist.

### DIFF
--- a/proxy-recorder/process_gatling_recording.rb
+++ b/proxy-recorder/process_gatling_recording.rb
@@ -42,6 +42,21 @@ def validate_infile_path(infile)
   end
 end
 
+def find_certname(text)
+  matches = text.match(/\n\s*\.get\("\/puppet\/v3\/node\/([^\?]+)\?environment/)
+  unless matches
+    puts "Unable to find certname (from node request) in recording!"
+    exit 1
+  end
+
+  certname = matches[1]
+
+  puts "Found certname: '#{certname}'"
+  puts
+
+  certname
+end
+
 def find_report_request_info(text)
   matches = text.match(/\n\s*\.put\("\/puppet\/v3\/report[^"]+"\)\s*\n\s*\.headers\(([^\)]+)\)\s*\n\s*\.body\(RawFileBody\("([^"]+)"\)\)\)\s*\n/)
   unless matches
@@ -230,9 +245,10 @@ EOS
                     '"transaction_uuid":"${transactionUuid}"')
 end
 
-def step13_setup_node_feeder()
+def step13_setup_node_feeder(text, report_text, certname)
   puts "STEP 13: Set up ${node} var for feeder"
-  puts "\t(Not yet implemented)"
+  text.gsub!(certname, "${node}")
+  report_text.gsub!(certname, "${node}")
 end
 
 def main(infile, outfile)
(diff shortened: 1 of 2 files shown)


camlow325:
This one should probably be rebased since it's based on #68, which has already been merged.

cprice404:
rebased

proxy-recorder/process_gatling_recording.rb
```diff
@@ -20,17 +22,171 @@ def lookup_request_match(line, prefix)
...
-def rename_package(text)
+def step2_rename_package(text)
+  puts "STEP 2: Renaming package to `com.puppetlabs.gatling.node_simulations`"
   text.gsub!(/^\s*package (.*)/, "package com.puppetlabs.gatling.node_simulations")
 end
 
 # Step 3
-def add_new_imports(text)
+def step3_add_new_imports(text)
+  puts "STEP 3: Adding additional import statements (for SimulationWithScnario and date/time classes)"
```
camlow325: s/Scnario/Scenario

proxy-recorder/process_gatling_recording.rb
```diff
@@ -20,17 +22,171 @@ def lookup_request_match(line, prefix)
...
+def generate_node_config(simulation_class, certname)
+  node_configs_dir = File.join(get_simulation_runner_root_dir(), "config", "nodes")
+  node_config_filename = simulation_class.sub(/.*\.([^\.]+)$/, '\1') + ".json"
+  node_config_file = File.absolute_path(File.join(node_configs_dir, node_config_filename))
+  if File.exists?(node_config_file)
+    puts "ERROR: file already exists at path '#{node_config_file}'; exiting."
+    exit 1
+  end
+
+  data = prompt("Node config file name? [#{node_config_filename}]> ")
```
camlow325: Not a big deal but it might be nice to have a way to specify these as command line arguments to the tool and, if specified, use those and avoid prompting.  I generally like command line args over prompting because it's easier to repeat the command later from shell history, etc.

camlow325:
Not necessarily related to this PR, but might be nice to just make `process_gatling_recording.rb` executable so you don't have to run it as `ruby .../process_gatling_recording.rb`.  Not a big deal, though.

proxy-recorder/process_gatling_recording.rb
cprice404: Yeah, I agree that would be nice to add at some point.

camlow325:
👍 I ran this against an old raw gatling recording of a puppet agent run and the result produced by this script looked basically identical to the hand-edited one that was produced at the time.  I just spotted the one small typo.  I could merge this now, though, if you don't want to deal with that.

cprice404:
I'll fix the typo.  thanks

cprice404:
Added a commit that fixes the typo.

camlow325:
👍 :shipit:

PR #70, opened 2016-08-10
(SERVER-1465) recording for couch / perf medium catalog

cprice404:
These commits add a gatling scenario for PE Couch, for the 'medium' catalog from the perf control repo.


camlow325:
Looks like this will need a rebase.  Not sure if there are other PRs that should still land before it's worth doing one.

cprice404:
I think the scala-post-processing PR probably should come in before this one.

camlow325:
I just merged #69 so I think this can be rebased now.

cprice404:
rebased

camlow325:
👍 Tried this out on a "dev" vm and it worked great so :shipit:

PR #72, opened 2016-08-17
(MAINT) Add server_version sections to single/multi pass Jenkinsfiles

camlow325:
This commit adds a `server_version` section in the calls to the
pipeline() functions from the single-pass and multi-pass scenario
Jenkinsfiles.  This section is now required in that the beaker setup
step would otherwise fail if these sections are not included.

### DIFF
--- a/jenkins-integration/jenkins-jobs/scenarios/multi-pass-scenario/Jenkinsfile
+++ b/jenkins-integration/jenkins-jobs/scenarios/multi-pass-scenario/Jenkinsfile
@@ -4,12 +4,20 @@ node {
 }
 pipeline.multipass_pipeline([
         [job_name: 'firsttest',
+         server_version: [
+                 type: "pe",
+                 pe_version: "2016.2.0"
+         ],
          code_deploy: [
                  type: "r10k",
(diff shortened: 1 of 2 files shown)


jenkins-integration/jenkins-jobs/scenarios/multi-pass-scenario/Jenkinsfile
```diff
@@ -4,12 +4,20 @@ node {
 }
 pipeline.multipass_pipeline([
         [job_name: 'firsttest',
+         server_version: [
+                 type: "pe",
+                 pe_version: "2016.2.0"
```
camlow325: Running the same version of PE twice probably isn't that interesting, I guess.  Maybe this should be an earlier version - but later than PE 3.3 since we have a separate `compare-3.3-to-couch` one now?  @cprice404 - WDYT?

cprice404: This seems fine to me for now.  Once we have support for modifying settings before a run (HOCON, etc.), we could make one of these have 'environment_timeout' set to on and one set to off, or something along those lines.

PR #74, opened 2016-08-18
(SERVER-1452) Copy hiera config from source to target directory on SUT

camlow325:
This commit enables a job configuration to include a source location,
`hiera_config_source_file`, from which a hiera.yaml file will be copied
to the target location where Puppet expects it to be.  The datadir to be
used in the hiera.yaml file is also now configurable on a per-job basis
via the `hiera_config_datadir` key.


jenkins-integration/beaker/install/shared/45_install_hiera_config.rb
```diff
@@ -0,0 +1,39 @@
...
+test_name 'Install hiera config'
+
+def target_hiera_config_path(host)
+  return on(host, puppet('config print hiera_config')).stdout.chomp
+end
+
+def get_source_hiera_config(host, hiera_config)
+  source_hiera_config_path = hiera_config[:source_file]
+  if source_hiera_config_path.nil? || source_hiera_config_path.empty?
+    source_hiera_config_path = target_hiera_config_path(host)
```
cprice404: What does this do?

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -110,6 +110,19 @@ def step040_install_puppet_code(script_dir, code_deploy, server_era) {
     }
 }
 
+def step045_install_hiera_config(script_dir, code_deploy, server_era) {
+    hiera_config_source_file = (code_deploy["hiera_config_source_file"] == null) ?
+            "" : code_deploy["hiera_config_source_file"]
```
cprice404: Kinda seems like we could check for nil here and just skip the rest of this step from here if it's nil, but I feel like I must be missing something about what you're trying to do.

Also seems like you could check for the existence of the file here and if it doesn't exist, fail.  I've probably been inconsistent about doing checks like that from the groovy code vs. the shell code or beaker scripts, but my gut is that it'd be better to have as much of that in the groovy as possible?  WDYT?

Anyway, if there is some reason not to skip this step if that value is nil, then you can ignore all these comments and this is probably fine how it is.

cprice404:
+1 LGTM other than the questions about whether we need to do anything if the source path is null.

jenkins-integration/beaker/install/shared/45_install_hiera_config.rb
camlow325: The code as written in the PR would allow a job config to specify a datadir without specifying a full-blown hiera.yaml to apply first.  In that mode, we'd basically end up using the default hiera.yaml file laid down at package install time but possibly with the `datadir` key in it set to pick up hieradata from some other place.  The code above just defaults to pulling down the package-installed hiera.yaml file for the case that a job configuration would not provide its own.

I understand this is probably a very unlikely case that we'd need to support since we're usually going to have a custom hierarchy / custom hieradata or no hiera at all for whatever code configurations we're going to use.  If it bugs you too much, I suppose I could get rid the defaulting logic.  Not a big deal to me either way.

cprice404: It just seems pretty heavy-handed to me that, if I'm reading this code correctly, you're *always* going to overwrite the hiera.config file from the packages, even if the job doesn't specify any values for *either* of the two hiera settings.  It feels like there might be tests where we wanted to be 100% sure we were testing the default configuration and this seems like it makes it less obvious / certain that we will be.

In a vacuum I'd probably rather keep it simpler, but, I'm eager to keep moving forward, so I'll just merge this as-is for now.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
camlow325: > Kinda seems like we could check for nil here and just skip the rest of this step from here if it's nil, but I feel like I must be missing something about what you're trying to do.

I was allowing the environment variables to flow through as empty string if no corresponding keys for them were set in the `code_deploy` hash.  The code below this in the Ruby script layer would no-op copying in a hiera.yaml file and/or modifying the datadir in the hiera.yaml based on whether or not these environment variables have a non-empty value.  That's all these are here for.

> Also seems like you could check for the existence of the file here and if it doesn't exist, fail. I've probably been inconsistent about doing checks like that from the groovy code vs. the shell code or beaker scripts, but my gut is that it'd be better to have as much of that in the groovy as possible? WDYT?

I agree that checking for the existence of the file and failing would be a good thing to do.  It probably fails without those explicit checks anyway but not in a very understandable way.  I can add that in.

I think it might need to be done from the beaker code, though, since the files that I'd be checking for the existence of would be on the SUT and I don't think we have a straightforward way of accessing the SUT from where the Groovy code would be run, the Jenkins slave, right?

Also, I think I'd only want to check for the existence of `hiera_config_source_file` on the SUT and not the `hiera_config_datadir` since the `hiera_config_datadir` would often be an interpolated value and I don't really want any of this code to have to know anything about how to do the interpolation.  Sound reasonable?

jenkins-integration/beaker/install/shared/45_install_hiera_config.rb
camlow325: Yeah, you're right that it would overwrite the file with exactly the same content if the variables were both empty.  I thought about putting some more conditional logic in to skip doing anything for that case but had thought the code read a little more cleanly without doing that.  I (or somebody else I guess) could still add that back in if desirable later.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
cprice404: You're right about the fact that the file existence check would need to happen on beaker, good point.  I wouldn't worry about it for now though.

jenkins-integration/beaker/install/shared/45_install_hiera_config.rb
camlow325: I put up #75 to avoid running any of the custom hiera config steps if the job doesn't set them.  I did that at the groovy layer so it avoids touching the hiera configs and service restarting in that case.

PR #76, opened 2016-08-18
(SERVER-1471) Move scenario config out of job params

camlow325:
This commit moves the logic for specifying the
PUPPET_GATLING_SIMULATION_CONFIG setting from a Jenkins workflow job
parameter down into a job parameter at the level where the pipeline is
being constructed.  This was done since a multipass pipeline needs to
allow for the simulation config to be unique per "job" it runs.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -85,13 +85,12 @@ def step020_install_pe(SKIP_PE_INSTALL, script_dir, server_era) {
     }
 }
 
-def step025_collect_facter_data(job_name, script_dir) {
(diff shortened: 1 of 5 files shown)


camlow325:
I'm still testing this out with a compare-3.3-to-couch scenario run but wanted to get this up for early review.

cprice404:
LGTM happy to merge this when you tell me you're ready.

camlow325:
The compare-3.3-to-couch scenario passed on my dev Jenkins instance without any modifications to the job options so I think this can be merged whenever you're ready, @cprice404.

camlow325:
Added one other cosmetic update in 12bc271 to correct a bad reference to a couch sim in the commented out compare-3.3-to-couch job.

camlow325:
I rebased this on top of the changes that were merged for #77.  I also added one more commit, 4caead1, which extends the new approach for setting the gatling simulation config to the collect_facter_data step added in #77.

PR #77, opened 2016-08-18
(SERVER-1468) Add facter and scenario data

cprice404:
This commit writes facter data and some information about the
gatling scenario to a known location, so that the
puppet-gatling-jenkins-plugin can use it when post-processing
the gatling data after a run.

### DIFF
--- a/jenkins-integration/beaker/install/shared/collect_facter_and_scenario_data.rb
+++ b/jenkins-integration/beaker/install/shared/collect_facter_and_scenario_data.rb
@@ -1,0 +1,59 @@
+# Collect artifacts from machines
+# Brian Cain {brian.cain@puppetlabs.com}
+
+require 'puppet/gatling/config'
+require 'csv'
+require 'json'
+
+def get_facter_data
+  result = on master, "facter -p -j"
+  result.stdout
+end
+
+def save_data(facter_data, data_hash, simulation_id, config)
+  simulation_dir = File.join("..", "puppet-gatling", simulation_id)
+  unless File.exist?(simulation_dir)
+    FileUtils.mkdir_p simulation_dir
+  end
+
+  File.open(File.join(simulation_dir, 'facter-data.json'), 'w') { |file| file.write(facter_data) }
+  CSV.open(File.join(simulation_dir, 'important_data.csv'), 'w') { |csv| data_hash.to_a.each { |elem| csv << elem} }
+
+  File.open(File.join(simulation_dir, 'gatling_sim_data.csv'), 'w') do |file|
+    config["nodes"].each do |node|
+      file.write("#{File.basename(node["node_config"], ".json")},#{node["num_instances"]},#{node["num_repetitions"]}\n")
+    end
+  end
+  puts "Files now saved within #{simulation_dir}"
+end
+
+# disk, num cpus, speed of cpus, ram
+def get_data_hash(data)
+  facts_data = JSON.parse(data)
+  {
+      'processor0' => facts_data['processors']['models'][0],
+      'processorcount' => facts_data['processors']['count'],
+      'puppetversion' => facts_data['puppetversion'],
+      'blockdevice_sda_size' => facts_data['disks']['sda']['size'],
+      'memorysize' => facts_data['memory']['system']['total']
+  }
+end
+
+# Begin work
+
+puts "Gathering facter data and processing data..."
+config = parse_scenario_file(get_scenario_from_env())
+simulation_id = get_simulation_id_from_env()
+
+facter_data = get_facter_data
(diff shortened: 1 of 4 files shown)


cprice404:
@camlow325 this was resuscitated from some stuff brian/nate did a long time ago, and is a prereq for the jenkins plugin.  It probably doesn't work with 3.3 yet, sort of planning to just get stuff working with couch and then file additional PRs from there.

jenkins-integration/beaker/install/shared/collect_facter_and_scenario_data.rb
```diff
@@ -0,0 +1,59 @@
...
+      'processorcount' => facts_data['processors']['count'],
+      'puppetversion' => facts_data['puppetversion'],
+      'blockdevice_sda_size' => facts_data['disks']['sda']['size'],
+      'memorysize' => facts_data['memory']['system']['total']
+  }
+end
+
+# Begin work
+
+puts "Gathering facter data and processing data..."
```
camlow325: Is it better to do `puts` or `Beaker::Log.notify` for these tracing messages?  I've seen a few `Beaker::Log.notify` messages in other beaker scripts in the project but wasn't sure if we consider that the standard way these should be done or not.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -191,6 +199,9 @@ def single_pipeline(job) {
         stage '020-install-pe'
         step020_install_pe(SKIP_PE_INSTALL, SCRIPT_DIR, server_era)
 
+        stage '025-collect-facter-data'
```
camlow325: If we think all of these steps are going to end up being the same for a single as for an individual job instance in a multipass pipeline, at some point it would seem nice, if possible, to break the calls to these steps out into a helper method - maybe parameterized to indicate whether a `stage` line should be outputted before calling the underlying step helper function?

I keep forgetting to add a step to the `single_pipeline` when doing testing of a new function against the `multipass_pipeline` -- and vice-versa.  Having the common helper that the two share might help prevent us from these unintentionally drifting / breaking.

Maybe you can't do that in a `node` block here, though?

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -85,6 +85,14 @@ def step020_install_pe(SKIP_PE_INSTALL, script_dir, server_era) {
     }
 }
 
+def step025_collect_facter_data(job_name, script_dir) {
+    withEnv(["PUPPET_GATLING_SIMULATION_CONFIG=${PUPPET_GATLING_SIMULATION_CONFIG}",
```
camlow325: This will conflict with the approach in #76 which reworks how `PUPPET_GATLING_SIMULATION_CONFIG` is derived.  Not sure if you'd rather have this PR come in before that one - doesn't matter to me.

jenkins-integration/lib/puppet/gatling/config.rb
```diff
@@ -45,6 +45,15 @@ def get_scenario_from_env()
   scenario_file
 end
 
+# Returns the $PUPPET_GATLING_SIMULATION_ID variable or throws an error.
+def get_simulation_id_from_env()
+  simulation_id = ENV['PUPPET_GATLING_SIMULATION_ID']
+  if !simulation_id
```
camlow325: Should we fail if it's defined but empty?  Probably wouldn't hit that often.

camlow325:
🙈 single-pass-scenario passed against this PR.  I didn't see exactly how the new artifacts get used but I did see that a `/var/lib/jenkins/jobs/single-pass-scenario/workspace/puppet-gatling/singletest` directory on the Jenkins node contained three files with seemingly reasonable content: `facter-data.json`, `gatling_sim_data.csv`, and `important_data.csv`.

I'm going to go ahead and merge this to keep things moving.

jenkins-integration/beaker/install/shared/collect_facter_and_scenario_data.rb
cprice404: This is a good point, we should probably try to stick to the Beaker::Log one.  Will try to clean that up later.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
cprice404: That's a good idea - it hadn't occurred to me that the 'stage' part might be able to be inside of a conditional, but it's worth a try.

PR #79, opened 2016-08-23
(MAINT) Switch to puppet-gatling-jenkins-plugin

cprice404:
This commit switches us from calling the `gatlingArchive()` step
in the gatling jenkins plugin to calling the `puppetGatlingArchive()`
step in the puppet-gatling-jenkins-plugin.


camlow325:
:shipit:

PR #78, opened 2016-08-23
(SERVER-1492) Add support for installing OSS

cprice404:
This commit adds support for installing OSS puppetserver instead
of just PE.  It has a few limitations:

1. Currently only supports 'latest' puppetserver & puppet-agent
2. Have not yet recorded a new gatling sim for OSS, so the Jenkinsfile
   is just using the pe-couch one, so some filemeta requests will fail.

PR #80, opened 2016-08-23
(SERVER-1492) recording for oss 2.5.x, perf-medium catalog

cprice404:


camlow325:
Now that #78 has been merged, should we include a foss25xperfmedium scenario file that points to the other files in this PR?  Should we also update the oss-puppetserver-stable and oss-puppeserver-latest Jenkinsfiles to point to a new related scenario file as well?

camlow325:
👍 I locally ran this by putting a scenario file in place and having the oss-puppetserver-stable job point to a scenario file which uses this catalog recording.  This worked fine other than that some of the node requests failed - similar to the catalog request failures mentioned in #78.  I think we should compensate for that by having the setup perform an agent run to ensure that the yaml/node and yaml/facts directories exist before we kick off Gatling.  In separate PRs, I'll put up the work I did to use a foss scenario file and do the early agent run.  For now, I'm going to go ahead and merge this.

PR #81, opened 2016-08-24
(SERVER-1466) Configure custom JAVA_ARGS for pipeline jobs

camlow325:
This commit adds the ability for a pipeline job to supply a new
argument, `server_java_args`, which can be used to setup custom
JAVA_ARGS which, when set, replace the ones from the Puppet Server
sysconfig file.

### DIFF
--- a/jenkins-integration/beaker/install/shared/60_classify_nodes_via_site_pp.rb
+++ b/jenkins-integration/beaker/install/shared/60_classify_nodes_via_site_pp.rb
@@ -26,7 +26,7 @@ def classify_foss_nodes(host, nodes)
     # need to restart the server to make this change take effect.
     # TODO: should move the restart into a helper method or something, this
     # is duplicated in 99_restart_server.rb.
-    service_name = ENV['PUPPET_SERVER_SERVICE_NAME']
+    service_name = get_puppet_server_service_name_from_env()
 
     on(host, "systemctl restart #{service_name}")
 
--- a/jenkins-integration/beaker/install/shared/80_configure_java_args.rb
+++ b/jenkins-integration/beaker/install/shared/80_configure_java_args.rb
@@ -1,0 +1,38 @@
+require 'puppet/gatling/config'
+require 'inifile'
+
+test_name 'Configure java args'
+
+def service_config_name(service_name)
+  "/etc/sysconfig/#{service_name}"
+end
+
+def get_service_config(host, service_name)
+  local_service_config_dir = Dir.mktmpdir
+  scp_from(host, service_config_name(service_name), local_service_config_dir)
+  File.join(local_service_config_dir, service_name)
+end
+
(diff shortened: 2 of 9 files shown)


camlow325:
For simplicity, I was hoping in this pass to just use the same approach for setting the arguments for both PE and OSS - modifying the value for JAVA_ARGS in the sysconfig file directly.  I originally wanted to have the step to customize JAVA_ARGS happen in 030_customize_settings.rb.  This would be problematic for PE, though, since an agent run on the master could wipe out any manually modified settings in the sysconfig file.

The PE agent is disabled during PE installation by this code in beaker-pe:

https://github.com/puppetlabs/beaker-pe/blob/0.8.0/lib/beaker-pe/install/pe_utils.rb#L475-L477

However, this line in the `97_enable_file_sync.rb` later on does an agent run to apply file sync configuration changes:

https://github.com/puppetlabs/gatling-puppet-load-test/blob/afa54b776643b6ef5e3b475a4eed59e1ed9039ad/jenkins-integration/beaker/install/pe/97_enable_file_sync.rb#L7

For this PR, then, I added a new step after file sync for applying the JAVA_ARGS and pushed other previously existing steps back by `10` positions to accommodate for that.

camlow325:
I'm going to file a couple of follow on tickets for this work:

1) Consider having the JAVA_ARGS for a pipeline job include options which are munged into the existing set on disk rather than replacing them, as this PR would do.

In this PR, for example, if the JAVA_ARGS in the sysconfig file on disk had "-Dfoo=bar -Xms2g -Xmx2g" but the pipeline job specified "-Dbaz=bim -Xms8g -Xmx8g", the JAVA_ARGS setting in the file on disk would have "-Dbaz=bim -Xms8g -Xmx8g".  It may be desirable -- like is possible in JAVA_ARGS configuration through Hiera parameters and console classification -- to munge in the job specified values such that the resulting setting on disk would have "-Dfoo=bar -Dbaz=bim -Xms8g -Xmx8g".

2) Modify the approach for setting JAVA_ARGS to use PE mechanisms -- either Hiera config parameter or console classification -- vs. just modifying the value in the sysconfig file on disk directly (which would still be done for OSS.

This would allow for the beaker step to be moved back up to an earlier point in the process -- maybe `30_customize_settings.rb` -- and reduce the risk of a one-off agent run on the PE master wiping out the JAVA_ARGS.

camlow325:
I added a couple of follow up tickets for the items I mentioned above:

* [SERVER-1497](https://tickets.puppetlabs.com/browse/SERVER-1497) - Use PE mechanisms for managing JAVA_ARGS in Gatling Automation Sims
* [SERVER-1498](https://tickets.puppetlabs.com/browse/SERVER-1498) - Allow Custom JAVA_ARGS for Gatling Sims to be Additive

PR #82, opened 2016-08-24
(SERVER-1486) Support facter data collection on older versions

cprice404:
This commit adds some `server_era` data about facter versions;
it allows us to distinguish between versions of facter that support
structured facts vs. those that don't, when we are computing the
relevant data about the SUT hardware before running the simulations.


cprice404:
Rebased this, so hopefully now it doesn't have issues with 'SKIP_PE_INSTALL'.

camlow325:
👍 on the code changes so far.  Just doing a full run of this with the `compare-3.3-to-couch` scenario to see that it all works properly.  Assuming no issues, I'll go ahead and merge this.

camlow325:
The `compare-3.3-to-couch` scenario ran fine for me without the need for any manual modifications so I'm good merging this.  :shipit:

PR #84, opened 2016-08-25
(SERVER-1492) Do an agent run on FOSS SUT before Gatling run

camlow325:
This commit adds logic to perform an agent run on a FOSS SUT before
kicking off the Gatling run.  This is needed in order to ensure that all
of the necessary directories on the master have been created before
concurrent agent runs can be performed.  Without this step, some of the
first few agent runs performed via Gatling could fail due to PUP-6651.


camlow325:
This is based on work from #83.  That PR should be merged first and then I can rebase this one.

PR #85, opened 2016-08-25
(SERVER-1499) Add support for background scripts, metrics curl

cprice404:

* Adds support in the Jenkinsfile data structure for listing one
  or more scripts to execute in the background on the SUT during
  the run
* Adds logic to shut these scripts down after the run
* Adds a `curl-server-metrics` script that will log data from
  the metrics endpoint to a file every 5 minutes


cprice404:
Added initial support for archiving files; I think I probably need to fix it up a bit to support multi-pass scenarios, but it's probably mergeable if it gets reviewed before then.

cprice404:
Added another commit that makes the archive stuff work in multi-pass.

I think I'm all done on this one now.

jenkins-integration/beaker/install/shared/start_sut_background_scripts.rb
```diff
@@ -0,0 +1,28 @@
...
+  Beaker::Log.notify("Launching #{bg_scripts.count} background scripts")
+  master_tempdir = master.tmpdir('gplt_bg_scripts')
+  pids = {}
+  bg_scripts.each do |s|
+    script_filename = File.basename(s)
+    remote_path = File.join(master_tempdir, script_filename)
+
+    Beaker::Log.notify("Launching script '#{s}'")
+    scp_to(master, s, master_tempdir)
+    result = on(master, "#{remote_path} > #{master_tempdir}/stdout.log 2> #{master_tempdir}/stderr.log & echo $!")
```
camlow325: Nice - didn't know about `$!`.

jenkins-integration/beaker/install/shared/stop_sut_background_scripts.rb
```diff
@@ -0,0 +1,12 @@
+require 'json'
+
+step "Stop background scripts on SUT" do
+  pids = JSON.parse(File.read("bg_pids.json"))
+
+  Beaker::Log.notify("Stopping #{pids.count} background scripts")
+  Beaker::Log.notify(JSON.pretty_generate(pids))
+  pids.each do |k, v|
+    Beaker::Log.notify("Stopping script '#{k}' with pid '#{v}'")
+    on(master, "kill #{v}")
```
camlow325: Should we do some kind of a wait after each `kill` command to validate that the process is dead before moving on?  Maybe something like:

```ruby
on (master, "while kill -0 #{v} 2>/dev/null; do sleep 0.5; done")
```

jenkins-integration/beaker/install/shared/start_sut_background_scripts.rb
```diff
@@ -0,0 +1,28 @@
...
+    Beaker::Log.notify("GOT RESULT FROM SCRIPT EXECUTION: '#{result}'")
```
camlow325: Is including the `result` object here useful?  When I ran this on my dev Jenkins instance, it just output an object with object id, like:

```
GOT RESULT FROM SCRIPT EXECUTION: '#<Beaker::Result:0x00000004fe81a8>'
```

jenkins-integration/jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -x
+
+while true ; do
```
camlow325: For dev run purposes, we may want to nuke this file, if it exists, before starting the `curl` loop - just so that we don't end up with status info left over from a prior job.

jenkins-integration/jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh
```diff
@@ -0,0 +1,13 @@
...
+  curl -sS -w "\n" -k https://localhost:8140/status/v1/services?level=debug >> /var/log/puppetlabs/puppetserver/metrics.json
```
camlow325: Would it help to ensure that the output from a full job run could be "well-formed" json?  Like, could we just write the initial `metrics.json` file with a hash containing a key whose value is an array and append the output from each curl query as an element in the array (delimited with commas).  I guess closing it off when the run is over would be a little more challenging.  Could provide an extra job-level hook for running a custom cleanup script after the stop has been completed.  Or could just rewrite this script in a such a way where it could catch the default TERM signal from the kill which is done by the stop and close off the metrics.json file before exiting out.  Actually, that latter approach might be better anyway so we don't run the risk of getting a corrupted, half-written file if the stop is stopped in the middle of writing the output from a curl request.  Suppose that wouldn't be too likely given the frequency at which the curl requests would occur.

I guess if you're going to have something downstream which knows to interpret the non-well formed format that none of this would matter much.  Would just be convenient for manual inspection of the raw file that gets archived at the end of a run, I guess.  😄

camlow325:
It appears that if you run a job that doesn't have any background scripts configured that the job would now fail with these changes.  I tried running the oss-puppetserver-stable job (which doesn't have any scripts attached) from my dev Jenkins instance and it failed.  Here is some of the output from the end of the job log:

````
Entering stage 105-stop-bg-scripts
Proceeding
[Pipeline] echo
No background scripts configured, skipping.
[Pipeline] stage (110-collect-sut-artifacts)
Entering stage 110-collect-sut-artifacts
Proceeding
[Pipeline] echo
No SUT archive files configured, skipping.
[Pipeline] stage (900-collect-driver-artifacts)
Entering stage 900-collect-driver-artifacts
Proceeding
[Pipeline] puppetGatlingArchive
[PuppetGatling] - Starting deployment from the post-action ...
Not adding report 'oss-stable-1472070841126' because mtime '1472070849000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472077710430' because mtime '1472077725000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472081141742' because mtime '1472081155000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472140308472' because mtime '1472140323000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472140620151' because mtime '1472140634000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472140796587' because mtime '1472140811000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472142671968' because mtime '1472142686000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472142808846' because mtime '1472142823000' is older than the build start time '1472226572463'
Not adding report 'oss-stable-1472142972427' because mtime '1472142987000' is older than the build start time '1472226572463'
Adding report 'oss-stable-1472226636081' because mtime '1472227502000' is newer than the build start time '1472226572463'
Found 1 reports to archive.
[PuppetGatling] - Getting simulation configuration data...
[PuppetGatling] - Got it!
[PuppetGatling] - Here are the Gatling Simulation Results for oss-stable: FOSS25xPerfMedium, 20, 1
[PuppetGatling] - The stats javascript file is: /var/lib/jenkins/jobs/oss-puppetserver-stable/builds/16/simulations/oss-stable-1472226636081/js/stats.js
[PuppetGatling] - The hash map is below.
[PuppetGatling] - The values are printed as: [Total Requests, Successful Requests, Failed Requests, Mean Response Time]
[PuppetGatling] - Node group summary data for group 'FOSS25xPerfMedium': FOSS25xPerfMedium                                            [20,    20,    0,    4119]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium node                                       [20,    20,    0,    100]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium filemeta pluginfacts                       [20,    20,    0,    20]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium filemeta plugins                           [20,    20,    0,    87]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium catalog                                    [20,    20,    0,    2018]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium filemeta                                   [2460,    2460,    0,    12]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium report                                     [20,    20,    0,    407]
[PuppetGatling] - Generating simulation report data...
[PuppetGatling] - OS Data saved.
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
java.lang.ArrayIndexOutOfBoundsException: 1
	at com.puppetlabs.jenkins.plugins.puppetgatling.PuppetGatlingPublisher.generateSimulationReport(PuppetGatlingPublisher.java:307)
	at com.puppetlabs.jenkins.plugins.puppetgatling.PuppetGatlingPublisher.getBuildAction(PuppetGatlingPublisher.java:156)
	at com.puppetlabs.jenkins.plugins.puppetgatling.PuppetGatlingPublisher.perform(PuppetGatlingPublisher.java:106)
	at com.puppetlabs.jenkins.plugins.puppetgatling.steps.PuppetGatlingArchiverStepExecution.run(PuppetGatlingArchiverStepExecution.java:30)
	at com.puppetlabs.jenkins.plugins.puppetgatling.steps.PuppetGatlingArchiverStepExecution.run(PuppetGatlingArchiverStepExecution.java:11)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1$1.call(AbstractSynchronousNonBlockingStepExecution.java:49)
	at hudson.security.ACL.impersonate(ACL.java:213)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1.run(AbstractSynchronousNonBlockingStepExecution.java:47)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Finished: FAILURE
````

I suppose we'll probably want to do the curl status query for most jobs anyway so maybe not a big deal for now.

jenkins-integration/beaker/install/shared/start_sut_background_scripts.rb
cprice404: Yeah, not really.  I'll delete that line.  Honestly all of these were just for debugging but I thought that some of them might end up being useful in the future.

jenkins-integration/beaker/install/shared/stop_sut_background_scripts.rb
cprice404: Not necessarily opposed to this, but do you have a concrete example in mind of a case where it would be problematic to move on without that?  Also did you test that line that you pasted above?  Or just guessing?

jenkins-integration/jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh
cprice404: Yeah I thought about that, but figured it shouldn't matter at all for prod and couldn't think of a case where it would be problematic in dev, so I left it.  I guess I don't mind deleting it if you think it'd be better; just wanted to err on the side of not losing the file in case it contained something useful for debugging, but if the archiving code is working correctly then that shouldn't be an issue anymore.

jenkins-integration/beaker/install/shared/stop_sut_background_scripts.rb
camlow325: For our while looper script, hopefully it will just die fast enough it being alive to potentially keep writing stuff when the archiving kicks in later on won't be a problem.  I haven't practically run into this race condition yet, just seemed like it could happen.

I haven't run that bash line via beaker but I have run that bash line on one of the SUTs locally from the shell and it seemed to work well for looping until the process died.

If you want to just wait and see if we have a problem, I'm fine with that.

jenkins-integration/jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh
camlow325: For prod, I don't think it will be a problem.  For dev, I think it might be confusing if we end up graphing data from multiple runs together for one job instance.  I understand the concern about potentially losing data if we just wipe it out before we start.  Maybe a compromise would be to move the file at the start of a run, if it exists, to a timestamped backup - like `metrics.json.20160826...`? That way, each file would have a set of metrics for a single job run only.

jenkins-integration/jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh
cprice404: I've thought about this a little and my current opinion is that we should wait until we start trying to write tooling around this, and see what kinds of issues we actually encounter, before changing it.  I've used this exact curl command for a bunch of debugging in the past (e.g. graphing memory usage when we landed our HUP work) and haven't had any issues so far.

jenkins-integration/beaker/install/shared/stop_sut_background_scripts.rb
cprice404: I'll try adding the line.  If it doesn't work immediately I may revert it because I don't want to invest too much time in it for now, but if it does work, I agree it's cleaner.

jenkins-integration/jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh
camlow325: Fair enough.  I'm good waiting to see how this shakes out.

cprice404:
I don't want it to fail if you don't have any scripts configured.  I will double-check that.  I have a hard time imagining how that could cause the particular stack trace you pasted, so there might be something else going on, but it seems like it needs to be fixed if it's reliably reproducible.

camlow325:
Huh, does seem like the job failure I got isn't directly related to the background looper script.  I added that in and re-ran the foss job and it failed in a similar way, only with the archiving now happening.  No idea what causes that exception at the end.  Maybe I have a bad version of the puppetgatling plugin installed?

````
~/jobs/oss-puppetserver-stable/workspace
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] echo
Archiving SUT file: 'jenkins-integration/sut_archive_files/oss-stable/metrics.json'
[Pipeline] sh
[workspace] Running shell script
+ '[' '!' -f ./jenkins-integration/sut_archive_files/oss-stable/metrics.json ']'
[Pipeline] archive
[Pipeline] stage (900-collect-driver-artifacts)
Entering stage 900-collect-driver-artifacts
Proceeding
[Pipeline] puppetGatlingArchive
[PuppetGatling] - Starting deployment from the post-action ...
Not adding report 'oss-stable-1472070841126' because mtime '1472070849000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472077710430' because mtime '1472077725000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472081141742' because mtime '1472081155000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472140308472' because mtime '1472140323000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472140620151' because mtime '1472140634000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472140796587' because mtime '1472140811000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472142671968' because mtime '1472142686000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472142808846' because mtime '1472142823000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472142972427' because mtime '1472142987000' is older than the build start time '1472228371291'
Not adding report 'oss-stable-1472226636081' because mtime '1472227502000' is older than the build start time '1472228371291'
Adding report 'oss-stable-1472228437288' because mtime '1472229303000' is newer than the build start time '1472228371291'
Found 1 reports to archive.
[PuppetGatling] - Getting simulation configuration data...
[PuppetGatling] - Got it!
[PuppetGatling] - Here are the Gatling Simulation Results for oss-stable: FOSS25xPerfMedium, 20, 1
[PuppetGatling] - The stats javascript file is: /var/lib/jenkins/jobs/oss-puppetserver-stable/builds/17/simulations/oss-stable-1472228437288/js/stats.js
[PuppetGatling] - The hash map is below.
[PuppetGatling] - The values are printed as: [Total Requests, Successful Requests, Failed Requests, Mean Response Time]
[PuppetGatling] - Node group summary data for group 'FOSS25xPerfMedium': FOSS25xPerfMedium                                            [20,    20,    0,    4033]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium node                                       [20,    20,    0,    99]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium filemeta pluginfacts                       [20,    20,    0,    18]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium filemeta plugins                           [20,    20,    0,    84]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium catalog                                    [20,    20,    0,    1953]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium filemeta                                   [2460,    2460,    0,    12]
[PuppetGatling] - The hash map key, value is: FOSS25xPerfMedium, FOSS25xPerfMedium report                                     [20,    20,    0,    370]
[PuppetGatling] - Generating simulation report data...
[PuppetGatling] - OS Data saved.
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
java.lang.ArrayIndexOutOfBoundsException: 1
	at com.puppetlabs.jenkins.plugins.puppetgatling.PuppetGatlingPublisher.generateSimulationReport(PuppetGatlingPublisher.java:307)
	at com.puppetlabs.jenkins.plugins.puppetgatling.PuppetGatlingPublisher.getBuildAction(PuppetGatlingPublisher.java:156)
	at com.puppetlabs.jenkins.plugins.puppetgatling.PuppetGatlingPublisher.perform(PuppetGatlingPublisher.java:106)
	at com.puppetlabs.jenkins.plugins.puppetgatling.steps.PuppetGatlingArchiverStepExecution.run(PuppetGatlingArchiverStepExecution.java:30)
	at com.puppetlabs.jenkins.plugins.puppetgatling.steps.PuppetGatlingArchiverStepExecution.run(PuppetGatlingArchiverStepExecution.java:11)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1$1.call(AbstractSynchronousNonBlockingStepExecution.java:49)
	at hudson.security.ACL.impersonate(ACL.java:213)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1.run(AbstractSynchronousNonBlockingStepExecution.java:47)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Finished: FAILURE
````

cprice404:
@camlow325 added a commit that fixes all the stuff you mentioned except the facter issue with OSS, which I'll look at separately if you haven't already gotten it sorted.

camlow325:
The error I was seeing turned out to be due to the `facter_structured_facts` parameter for OSS `server_era` not being set to true, causing the SUT facts to be mishandled.  After I fixed that (and patched the puppetserver build to include a "status" service), the job succeeded for me and produced a `metrics.json` file with reasonable-looking content as a Jenkins build artifact.

I'll address both the structured facts and puppetserver status service issues separately but will go ahead and merge this.

camlow325:
Added #87 to cover the structured facts issue for OSS.

PR #89, opened 2016-08-29
(MAINT) Make SUT archive file path consistent

cprice404:
Prior to this commit, we were writing the facter data and other
CSV files into a directory called 'puppet-gatling' and a subdirectory
named after the job_name in the Jenkinsfile.  We were then
creating a similar, redundant directory structure to store the 'archive'
files in.  This commit just tweaks the logic for the 'archive' files
to re-use the existing directory structure under 'puppet-gatling'
and put the files near the facter data files.


camlow325:
👍

PR #95, opened 2016-09-12
(SERVER-1465) Add 2-hour scenarios for pe33 vs. couch

cprice404:
This commit adds scenarios for 2 hours, 500 agents for PE 3.3 vs. Couch.
The agent runtimes we've seen for PE3.3 so far are so long that I wanted
to start with a low-ish number of agents.  If the run goes well we could
try one with more agents.


camlow325:
👍

PR #99, opened 2016-09-26
(SERVER-1559) support for puppet.conf settings, couch w/o static catalogs

cprice404:
This PR adds support for specifying a map called `puppet_settings` in the Jenkinsfile.  If present, then before the run begins, we will modify the specified settings to the corresponding values and restart the server.

It also adds a Couch simulation with static catalogs disabled, for comparison with PE 3.3 and the Couch scenario with static catalogs enabled.


camlow325:
👍 with just the one minor note - which could be ignored for now if you'd like.

jenkins-integration/beaker/install/shared/30_customize_settings.rb
```diff
@@ -0,0 +1,25 @@
...
+sections = puppet_settings.keys
+sections.each do |s|
+  Beaker::Log.notify("Updating puppet settings for section '#{s}'")
+  puppet_settings[s].each do |setting, value|
+    Beaker::Log.notify("In section '#{s}', setting '#{setting}' to '#{value}'")
+    on(master, puppet("config set #{setting} #{value} --section #{s}"))
+  end
+end
+
+puppet_conf = on(master, "cat /etc/puppetlabs/puppet/puppet.conf").stdout.chomp
```
cprice404: that would be better but then I'd have to test it again :)  And I'm assuming we probably aren't going to be doing much (any?) additional testing on p3?

camlow325: Fair enough.  Can ignore it for now and look at it again when we might want to do some p3 testing.  Wasn't sure if you had any more plans for PE 3.3 testing in the works or what not.

camlow325:
:shipit:

PR #100, opened 2016-09-28
(SERVER-1559) Standalone static catalogs test

cprice404:
In the previous PRs for this ticket, I added a "no static catalogs"
setup to the existing pe33-vs-couch job.  This worked fine, but
because both couch configurations perform so much better than the PE3.3
run, it was hard to visualize the perf delta between the two couch
runs (with and without static catalogs).  This commit introduces
a new job that will strictly focus on couch, with and without
static catalogs.

The first few iterations of this are just going to be used to establish
a baseline for how many agents the server can handle, given a 12G heap
size.  This is intended to be run on the dev jenkins server, not the
production one.  Once we have some data about that, I'll file
a cleanup PR that sets up the final configurations for use in production,
and will run multiple scenarios with varying numbers of agents to
see how the load affects overall performance.


camlow325:
👍

PR #106, opened 2016-10-25
(SERVER-1467) Add support for modifying hocon files

jpinsonault:
Adds a groovy pipeline step with corresponding bash and beaker files, which lets you add/change/remove hocon settings on the SUT

Throwing this up here so people can look at it, but it needs at least a couple things:
- A way of installing the hocon gem on the SUT
- Hocon needs to be released before we can take advantage of the CLI tool

The settings can be provided by passing a `hocon_settings` key to the groovy pipeline functions.

For example, it might look like this:

``` groovy
hocon_settings: [
        [
                file: "/tmp/foo.conf",
                path: "foo.bar",
                action: "set",
                value: "blerp"
        ],
        [
                file: "/tmp/foo.conf",
                path: "foo.baz",
                action: "set",
                value: "hello"
        ],
        [
                file: "/tmp/foo.conf",
                path: "foo.baz",
                action: "unset",
        ]
]
```


jenkins-integration/beaker/install/shared/35_customize_hocon_settings.rb
```diff
@@ -0,0 +1,28 @@
...
+
+p hocon_settings
+hocon_settings.each do |setting|
+  case setting["action"]
+    when 'set'
+      set_hocon_setting(setting['file'], setting['path'], setting['value'])
+    when 'unset'
+      unset_hocon_setting(setting['file'], setting['path'])
+    else
+      raise Exception.new("Invalid hocon action '#{setting["action"]}'. Must be one of 'set', 'unset'")
```
cprice404: I'm guessing that the 99% case here will be to "set", so maybe we should allow people to omit the "action", and default to "set"?  We could log a message here explaining that if we wanted to.

cprice404:
This looks good to me.  Had the one comment about providing a default 'action'.

As for getting the hocon gem installed on the SUT, there are a few things to consider:

1. The puppet-agent package ships with the gem, but you'll need to know whether your call to `hocon` is going to be done using the puppet ruby or the system ruby.
2. If it was done with the puppet ruby then you would need to worry about whether or not the agent version had the correct version of the hocon gem packaged, so it might be better to try to intentionally use the system ruby?
3. If there are any of the beaker provisioning steps on the SUT that are doing 'puppet apply' commands, then maybe that'd be a good place to look into adding a 'gem' resource that would get the gem installed - but again, you'll have to take care to make sure that the gem package provider is using the correct ruby (system vs. puppet).

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -403,6 +417,7 @@ def multipass_pipeline(jobs) {
                     SCRIPT_DIR,
                     server_era)
             step030_customize_settings(SCRIPT_DIR, job['puppet_settings'])
+            step035_customize_hocon_settings(SCRIPT_DIR, job['hocon_settings'])
```
camlow325: @cprice404 Do you think there's a chance that we'll want to use this to configure some settings that in PE could get wiped out / reset by the PE module during an agent run?  If so, we might want to move this step past the file_sync and classify_nodes steps like we did with `step080_customize_settings`.

jenkins-integration/jenkins-jobs/common/scripts/job-steps/035_customize_hocon_settings.sh
```diff
@@ -0,0 +1,24 @@
...
+ssh-add ${HOME}/.ssh/id_rsa
+
+bundle exec beaker \
+        --config hosts.yaml \
+        --load-path lib \
+        --log-level debug \
+        --no-color \
+        --tests \
+beaker/install/shared/35_customize_hocon_settings.rb
+#beaker/install/shared/99_restart_server.rb
```
camlow325: I wonder if we should create some sort of rollup restart / reload server step before we do the run_gatling_sim step?  Seems pretty likely that we'll want to have the server restarted after reconfiguring some hocon settings.  Without a way to configure this, though, not sure there's a guarantee that the server will have been restarted and, therefore, using the new settings before the gatling run is kicked off.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
cprice404: Fair question - I'd be fine with moving it, or playing it by ear and just making sure we think about that as we're building tests that consume this functionality.  Either way.

jenkins-integration/jenkins-jobs/common/scripts/job-steps/035_customize_hocon_settings.sh
cprice404: Yeah we should either add a restart/reload step as part of the hocon changes, or move to a roll-up restart at the end.  I don't have a strong preference.

jenkins-integration/beaker/install/shared/35_customize_hocon_settings.rb
jpinsonault: I think defaulting to "set" is a good idea. I'm not sure if we need to log anything though, since there will be a beaker step in the log saying "Set X in file Y"

jenkins-integration/beaker/install/foss/80_run_agent_on_master.rb
```diff
@@ -5,5 +5,6 @@
   # ahead of the Gatling run avoids collisions in the priming that can happen
   # when multiple agent runs happen at the same time.  See PUP-6651 for more
   # information.
-  on(master, 'puppet agent -t --server `hostname`')
+  fqdn = fact_on(master, 'fqdn')
+  on(master, "puppet agent -t --server #{fqdn}")
```
jpinsonault: I think as long as `master` will spit out the fqdn that should work fine. I added this because puppet runs were failing, since `--server ``hostname`` ` was only returning the hostname and there was a name mismatch. Could be that I just misconfigured something though. 

Looks like `master` comes from the hosts.yaml, which comes from the SUT_HOST variable when you kick off the job. So that's probably good

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
jpinsonault: If it's a setting that will be overridden by a puppet run, I assume that means this is PE land and we could use scooter in a new pipeline step to configure those things through the NC

jenkins-integration/jenkins-jobs/common/scripts/job-steps/035_customize_hocon_settings.sh
jpinsonault: Looks like there's 5 or 6 steps that each restart the server, which seems kind of extreme. Maybe a rollup-restart pipeline step would be good

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
camlow325: Yeah, that's true.  It would mean that we'd might have to have conditional logic in the test setup that's different for OSS vs. PE for particular settings whereas if we moved when we run the step down to a later time - after the agent on the master has been disabled - we could eliminate the conditional setup.  That said, I don't have a strong opinion about this.  If we think this is ultimately the best spot for this step, I'm fine with leaving it how it is here.

jpinsonault: Ah, I forgot we disable the puppet agent, that makes sense. I'm not sure where would be the appropriate place to move this step to

jpinsonault:
@cprice404 

> If there are any of the beaker provisioning steps on the SUT that are doing 'puppet apply' commands, then maybe that'd be a good place to look into adding a 'gem' resource that would get the gem installed

I'm not seeing any, but maybe we could add a step for `install_system_gems` or something?

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
camlow325: I think the agent is disabled on the master as part of the PE install.  During the "file_sync" setup step, though, we run the agent in order to have the file sync configuration applied.  If we were to move the `customize_hocon_settings` step to just before or after `step080_customize_settings`, I think we'd get the potential benefit of being able to modify a HOCON config setting that wouldn't be overridden by an agent run.  Although, admittedly, that's a bit precarious since someone could always add a step later on which runs the agent again and undoes the HOCON configuration.  I'd have a slight preference for moving this step to around where 080 is - and maybe document somewhere that future steps which do agent runs should not happen after the settings* steps.  But, again, I'd be good with punting on all of this if neither @jpinsonault nor @cprice404 thinks it's worth doing anything more on this at this point.

cprice404:
> I'm not seeing any, but maybe we could add a step for install_system_gems or something?

Sounds good to me.

jenkins-integration/beaker/install/shared/23_install_system_gems.rb
```diff
@@ -0,0 +1,17 @@
+require 'json'
+
+test_name 'Install System Gems'
+
+def install_system_gem(host, gem_name)
+  step "Install '#{gem_name}' gem"
+  # Ensure it's the system ruby
+  on(host, "/usr/bin/env gem install #{gem_name}")
```
jpinsonault: @cprice404 I wasn't sure if this was necessary (or even makes sense) or if just calling `gem` by itself could be guaranteed to be the system ruby

jenkins-integration/beaker/install/shared/23_install_system_gems.rb
```diff
@@ -0,0 +1,17 @@
...
+end
+
+step 'Install gems using system ruby'
+
+gem_list = ENV['PUPPET_GATLING_SYSTEM_GEMS'].split(',')
```
jpinsonault: That's a great point. I like just serializing hashes, seems less janky

jenkins-integration/beaker/install/shared/23_install_system_gems.rb
cprice404: I *think* this is better than just calling `gem`, because the latter would depend on the PATH environment variable.  I don't *think* there will ever be a case where puppet would cause `/usr/bin/env` to return the puppet ruby, but I don't entirely know how `/usr/bin/env` is implemented.

I think what I would do is:
* Spin up a box.  make sure it has a system ruby and that that is the ruby you get via `/usr/bin/env`.
* Install puppet-agent on that box.  Log out and log back in and confirm that puppetlabs stuff has been added to PATH.
* Check `/usr/bin/env gem` again.  If that is still pointing to the system ruby then I think you're probably good with this approach.

jenkins-integration/beaker/install/shared/23_install_system_gems.rb
cprice404: I don't want to introduce gems into the `Jenkinsfile` layer until we have a forcing function.  I'd be fine with changing the data structure that we use here such that it would support (require?) specific gem versions to be specified in the beaker setup code, but it shouldn't bubble all the way up to the Jenkinsfile for now.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
cprice404: I don't have a strong opinion one way or the other, so, happy to go with @camlow325 's suggestion of moving this to 80.

jenkins-integration/beaker/install/shared/23_install_system_gems.rb
camlow325: Cool, yeah, I'm fine with deferring on exposing this up to the `Jenkinsfile` layer.  I'd be fine with saying that a version is required (not optional) for each gem listed if that works for y'all.

jpinsonault: @cprice404 what file were you thinking of putting the gem info in?

cprice404: @jpinsonault you could just hard-code them into the `pipeline.groovy` for now?  like, in that `install_system_gems` method?  I can't think of a reason why it'd be bad to have the hocon gem installed for all the runs for now.

camlow325: @jpinsonault, sorry for not bringing this up earlier.  One of the things that I hadn't thought through with this approach of installing system gems is that we could arguably be polluting the state of the SUT by installing gems onto it which are just for infrastructure / setup purposes - and not part of the actual simulation that we are actually running.  To that end, an alternative approach we might want to consider for this is:

1) Just add the hocon gem dependency to the appropriate `jenkins-integration` Gemfile(s) where we're already installing `beaker` as part of the Jenkins slave setup for the test run.  This should make the hocon CLI tool available on the Jenkins slave.

2) In the `set_hocon_setting` / `unset_hocon_setting` functions, scp the file from the master down to the Jenkins slave, run the hocon CLI tool to make modifications, and scp the modified file back to the master.

This is a bit more inefficient than modifying the file in place on the master but I don't think the overhead would be too big of a deal.

I talked through this alternative approach with @cprice404 and I think he's supportive of it, for the benefit of reducing the amount of pollution of SUT state.  What do you think?

jpinsonault:
Pushed another commit that uses the beaker hocon helpers. Before it can be merged, https://github.com/puppetlabs/puppetlabs-puppetserver_perf_driver/pull/12 has to be merged. It bumps the ruby verison to 2.2.5 in order to get the latest beaker with the hocon helpers

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -191,12 +191,13 @@ def step030_customize_settings(script_dir, puppet_settings) {
...
 
-def step035_customize_hocon_settings(script_dir, settings) {
+def step035_customize_hocon_settings(script_dir, settings, server_era) {
     if (settings == null) {
         echo "Skipping hocon settings customization; no overrides found in `hocon_settings`."
     } else {
         settings_json = JsonOutput.toJson(settings)
-        withEnv(["PUPPET_GATLING_HOCON_SETTINGS=${settings_json}"]) {
+        withEnv(["PUPPET_GATLING_HOCON_SETTINGS=${settings_json}",
+                 "PUPPET_SERVER_SERVICE_NAME=${server_era["service_name"]}"]) {
```
cprice404: What changed that made the SERVICE_NAME a requirement?

cprice404: What changed that made the SERVICE_NAME a requirement?

cprice404: What changed that made the SERVICE_NAME a requirement?

jenkins-integration/jenkins-jobs/common/scripts/job-steps/initialize_ruby_env.sh
```diff
@@ -4,10 +4,10 @@ echo ""
 echo "Setting Up Ruby Environment"
 echo ""
 source /usr/local/rvm/scripts/rvm
-rvm use 2.1.6 || exit 1
+rvm use 2.2.5 || exit 1
```
cprice404: This seems fine but now that we have this requirement, I think we'll need to do the following before it's safe to merge this:

1. Create a new release of the puppetlabs-puppetserver_perf_driver (which is really just an update to the CHANGELOG, plus the creation of a tag on github)
2. File a PR against puppetlabs-modules that bumps the Puppetfile up to that version.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
jpinsonault: As far as I can tell that's needed for the restart step inside of the customize_hocon_setings step

cprice404: OK, that makes  sense, it just looked like this change came in in your latest commit, and it seemed like it would have been needed prior to that.  Might have just been some screwiness  in the github diff though.  I agree that var is necessary for the restart step.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
camlow325: Did we decide against moving this to 80?  If so, I'm okay with that.  Just noticed it was still at 035 as of the latest commits.

jpinsonault:
@cprice404 @camlow325 Pushed a commit that moves the hocon settings step to 85

I think the only things left on this PR is a PR against puppetlabs-modules bumping the perf driver module to the latest version to pick up ruby 2.25
https://github.com/puppetlabs/puppetlabs-modules/pull/6563

jpinsonault:
The modules PR was merged

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -158,6 +158,19 @@ def step020_install_server(SKIP_SERVER_INSTALL, script_dir, server_era) {
...
+    if (gem_list == null) {
+        echo "Skipping installing system gems; no gem list given"
+    }
+    else{
+        gem_string = gem_list.join(",")
+        withEnv(["PUPPET_GATLING_SYSTEM_GEMS=${gem_string}"]) {
+            sh "${script_dir}/023_install_system_gems.sh"
+        }
+    }
+}
```
cprice404: Seems like we don't need this anymore for now?

jpinsonault:
@cprice404 removed the system gem step

PR #107, opened 2016-10-26
(SERVER-1605) Support configuration of maximum build history for job

cprice404:
The goal of this commit was to give us a mechanism for configuring the
maximum number of builds that would be retained in the history for a given
job on the Jenkins server.  The commit sets up a default value of 50
for all jobs, but also overrides this to 10 for our large 1-week job,
since history for that job will consume significantly more disk space.

It ended up that there was a solution to this particular issue that
is pretty general, and will allow users to tweak other metadata about
the jobs beyond just the build history, so that is the approach that
this commit takes.  It allows you to define a JobDSL.groovy file
alongside your Jenkinsfile, and then make calls to the JobDSL API
to modify the job.


cprice404:
@rlinehan thanks for reviewing!  Probably not necessary to spin up a full dev env to test it, it'll get plenty of incidental coverage.

PR #108, opened 2016-10-31
(SERVER-1603) Add check for cent7 repo, fallback when looking for build

cprice404:
Prior to this commit, when the framework was looking for the latest
OSS Puppet Server build (from either the stable or master) branch,
it would simply look for the latest build *directory* that existed
on the builds server.  However, it is possible (and not particularly
uncommon) for a build directory to exist but not have a cent7 repo/artifact
in it (due to either the build still being in progress, or due to
cent7 not being included in the target platform list for that build).

This commit makes the script a little smarter, so that it will validate
the existence of a cent7 repo inside of the directory on the build
server before choosing that build.  If the cent7 repo doesn't exist,
the script will fall back and try the next most recent build.


camlow325:
👍

PR #109, opened 2016-10-31
(SERVER-1549) Set OSS latest job to run automatically

cprice404:
This commit adds a `JobDSL.groovy` file for the "OSS latest" job,
and schedules that job to run automatically, once a week, on
Monday evenings.


cprice404:
Realized that this won't quite work as-is because the default values for the build parameters (SUT hostname, 'skip provisioning', are wrong.  Closing and will open a new one with fixes.

PR #110, opened 2016-11-01
(SERVER-1549) Set OSS latest job to run automatically

cprice404:
This commit adds a `JobDSL.groovy` file for the "OSS latest" job,
and schedules that job to run automatically, once a week, on
Monday evenings.

### DIFF
--- a/jenkins-integration/jenkins-jobs/job_bootstrap.groovy
+++ b/jenkins-integration/jenkins-jobs/job_bootstrap.groovy
@@ -1,6 +1,54 @@
 import groovy.io.FileType
 import java.nio.file.Paths
 
+class DSLHelper {
+    PrintStream out;
+
+    DSLHelper(out) {
+        this.out = out
+    }
+    def overrideParameterDefault(job, param_name, new_default_value) {
+        this.out.println("\tAttempting to override '${param_name}' default value to '${new_default_value}' for job '${job.name}'")
+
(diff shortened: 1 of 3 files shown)


jenkins-integration/jenkins-jobs/job_bootstrap.groovy
```diff
@@ -1,6 +1,54 @@
...
+        // NOTE: HACK.  for some reason, the 'configure' block may get called
+        //  several times.  See https://issues.jenkins-ci.org/browse/JENKINS-39417 .
+        // Worse, it seems like on the subsequent calls to the configure block,
+        // the state of the job is reset to still include the original parameter
+        // data.  It seems harmless to simply execute the overrides multiple times,
+        // but logging the messages about the overrides multiple times looks very
+        // confusing in the seed job output.  By closing over this local variable,
+        // we can make sure the log messages only show up once, which makes
+        // the seed job output a little less confusing.
+        def param_checked = false
```
camlow325: Will this end up suppressing messages that it shouldn't in the case where we're overriding more than one parameter through the helper?

camlow325: Ah, okay, got it.  Works for me, then.

jpinsonault:
As far as I understand groovy and jenkins (which isn't far), this seems reasonable

jpinsonault:
👍

PR #111, opened 2016-11-04
(MAINT) Add beaker-pe to Gemfile

cprice404:
We recently upgraded to the new 3.x series of beaker, which
no longer includes PE functionality by default.  This commit
adds an explicit dependency on the beaker-pe gem to regain
that functionality.


cprice404:
@jpinsonault do you want to look at this before we merge?

jpinsonault:
yes

jpinsonault:
👍

PR #112, opened 2016-11-11
(PE-18258) Add recurring test for pe-latest

cprice404:
This commit adds a new job, `pe-puppetserver-latest`, which installs
the latest flanders build and runs our normal 2-hour test against it.

The JobDSL.groovy file schedules this job to run weekly on Tuesday
evenings.


jpinsonault:
This looks good to me

PR #113, opened 2016-11-11
(SERVER-1562) A-B test for jruby's jit mode

jpinsonault:
Adds a job which runs two oss-stable pipelines: one with jit on, and one with jit off


jpinsonault:
Probably not a whole lot of utility is testing it yourself unless you want to change something. Seemed to run pretty seamlessly for me

PR #114, opened 2016-11-17
(SERVER-1662) Fix cron collisions, introduce server_config

cprice404:
This commit introduces a new config file, `server_config.json`.
In it we can list out hostnames of jenkins servers, and set
attributes for them.  The attributes will be made available to the
`JobDSL.groovy` scripts via a local variable called `serverConfig`.

This allows JobDSL configurations to do things like conditionally
modify a job based on whether it is being created in a dev environment
vs. a production environment.

### DIFF
--- a/jenkins-integration/jenkins-jobs/README_JOBDSL_SYNTAX.md
+++ b/jenkins-integration/jenkins-jobs/README_JOBDSL_SYNTAX.md
@@ -51,5 +51,13 @@ job.with {
   At the time of this writing, the only method is `overrideParameterDefault(job, param_name, new_default_value)`, which
   allows you to override the default value for a parameter such as `SUT_HOST`.
 
+* `serverConfig`: this is a map, containing some basic configuration information about the server that the job is
+  being created on.  The data for this is loaded from `./server_config.json`.  The keys that it will contain are:
+
+  * "hostname": the hostname of the server.
+  * "environment": an identifier that can be used to distinguish between different classes of servers; most common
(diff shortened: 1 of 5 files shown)


jpinsonault:
👍  looks good to me. Just curious though, what cron collisions were happening and how does this fix it?

cprice404:
@jpinsonault prior to this patch, *all* instances of the 'oss-puppetserver-latest' and 'pe-puppetserver-latest' jobs, on any Jenkins nodes, would have the default value for their SUT field set to point to puppetserver-perf-sut54, and they would also have the trigger scheduled to run the jobs automatically every Monday/Tuesday evening.

So we've got the Jenkins server running on puppetserver-perf-driver55, and the one on puppetserver-perf-driver68-dev, and then if you, Jeremy and I all had a dev jenkins server on a vmpooler instance, that's 5 jenkins servers who would all be trying to kick those jobs off via the cron trigger.  All 5 of those jobs would attempt to do a razor re-provisioning of puppetserver-perf-sut54, and also run 5 concurrent sets of all of the beaker setup stuff and if by some miracle nothing failed along the way, 5 concurrent gatling load tests against that one SUT.

No bueno.  :)

jpinsonault:
@cprice404 that's impressively bad

camlow325:
I spun up a new dev Jenkins node off of this code and it came up with "foo" SUT defaults for the oss-puppetserver-latest and pe-puppetserver-latest jobs.  Looks good.  👍

PR #115, opened 2016-11-18
(SERVER-1563) Add a scenario for measuring environment-timeout

camlow325:
This PR adds a new scenario for doing Couch-based runs, comparing
performance with 'environment_timeout = 0' vs. 'environment_timeout =
unlimited' across different agent loads, similiar to the loads used
with the pre-existing static-catalogs scenarios.

This PR also bumps the various couch jobs from 2016.2.0 to 2016.2.1.
This is being done in order to avoid an intermittent problem seen with
the PE classifier at installation time on 2016.2.0, PE-16367, which
was fixed for 2016.2.1.

This PR also moves the previous 030_customize_settings step down to step
75, now also renamed to 075_customize_puppet_settings.  This happens
after the final `puppet agent` run on the master and so should allow the
`puppet_settings` changes to persist on into the Gatling run for PE, not
potentially being undone by a PE agent run.

This PR also fixes a bug in the handling of the pe_dir field for
beaker-hostgenerator for scenarios that use non-latest PE build
configurations.


jenkins-integration/jenkins-jobs/scenarios/couch-environment-timeout/Jenkinsfile
```diff
@@ -0,0 +1,97 @@
...
+environment_timeout_unlimited_1250.put('gatling_simulation_config', '../simulation-runner/config/scenarios/pe-couch-medium-1250-2-hours.json')
+
+pipeline.multipass_pipeline([
+        environment_timeout_0_500,
+        environment_timeout_unlimited_500,
+        environment_timeout_0_1000,
+        environment_timeout_unlimited_1000,
+        environment_timeout_0_1250,
+        environment_timeout_unlimited_1250,
+])
```
jpinsonault: Was trying to figure out a way to shorten this file up a bit, but not sure it's worth it
something like 

```groovy
jobs = []
["500", "1000", "1250].each { -> numAgents
    ["0", "unlimited"].each { -> timeout
        // Put those variables into a copy of the base config
        jobs.add(customizedJob)
    }
}
```

You'd only need one base config, but it'd be harder to change things later if it didn't fit the pattern I guess. I don't really care, wdyt?

camlow325: I'd be open to doing something like that.

I copied this approach pretty much verbatim from the [static-catalogs Jenkinsfile](https://github.com/puppetlabs/gatling-puppet-load-test/blob/3f57c276b9d1cc9bdc0c1c0297974be914921a36/jenkins-integration/jenkins-jobs/scenarios/couch-static-catalogs/Jenkinsfile).  I think @cprice404 went with this approach as a compromise from having to repeat the entire config for each variation but also was leery about having Groovy code that was too sophisticated, like if Jenkins might have security constraints or interpreter problems or something?  That said, this could cut down on the repetition a bit and, if we like it better, could be something we model for all other jobs like this.

Guess I'm on the fence a bit about this, leaning a bit toward not doing it until we have at least one more variation of this kind of job.
@cprice404 - what do you think?

cprice404:
@jpinsonault if you think we should push on the looping stuff a bit more, holler, otherwise you can merge this.

PR #118, opened 2017-01-20
(MAINT) Run puppetserver oss and pe jobs on cron instead of scm poll

camlow325:
Previously, the oss and pe puppetserver latest jobs were configured to
run weekly on an scm poll.  This would cause Jenkins to only perform the
full Gatling job run if a commit in the gatling-puppet-load-test repo
had changed since the last job that was run.  For these jobs, we're
interested in running the jobs even if nothing in the GPLT repo has
changed - for example, if a new puppetserver nightly is available.

In this commit, the jobs have been changed to run on a cron rather than
an scm poll.  This should ensure that the jobs run every week even if
nothing in the GPLT repo itself has changed from week to week.


jpinsonault:
👍

PR #120, opened 2017-02-10
(MAINT) Switch oss and pe-puppetserver jobs from sut54 to sut56

camlow325:
This commit switches the target sut for the oss and pe-puppetserver
latest jobs from sut54 to sut56.  The hardware which corresponded to
sut54 appears to be out of commission, for now at least.  We may switch
this back later when the hardware is rehabilitated.


jpinsonault:
👍

PR #121, opened 2017-03-01
(MAINT) Switch oss and pe-puppetserver jobs from sut56 to sut54

camlow325:
Hardware for sut54 is working again but sut56 is offline.  sut54 was
showing more consistent results from run to run than sut56 was when it
was working.  Switching back to sut54 to see if stability returns.


jpinsonault:
👍

jpinsonault:
ah screw it http://i.imgur.com/9ovX6ka.gif

PR #124, opened 2017-05-02
(SERVER-1790) Fix oss-puppetserver-latest run failures

camlow325:
This PR contains two changes:

1) Enable tk-auth for latest/master OSS Puppet Server runs.

Previously, all OSS Puppet Server jobs assumed that the
use-legacy-auth-conf setting was set to 'true' and so only tried to
reconfigure the Ruby Puppet auth.conf file to enable the permissive
access needed to perform multi-node runs.  As of Puppet 5, though,
use-legacy-auth-conf will be set to 'false' by default.  This commit
updates the server config to assume that for any builds being tested
from the Puppet Server master branch, the tk-auth configuration file is
setup for permissive access.

2) Skip non-existent cent7 puppet-agent builds

Previously, the logic which tried to determine which 'latest'
puppet-agent release build to install would assume that a cent7 build
was available on the repo server if its parent directory existed.  This
will sometimes not be true - e.g., where some artifacts may be built up
over the course of a few days near a release.

The logic for determining the latest puppetserver release skips over any
build where the repo folder exists but the cent7 build itself does not.
This commit reworks the logic for determining the latest build to be
more common for both puppet-agent and puppetserver.  As a result,
non-existent puppet-agent builds will be skipped over like puppetserver
ones have been.

We hit this one because the puppet-agent 1.10.1 directory existed on the
repo server but the cent7 builds did not yet, leading to puppet-agent failing
to be installed.  With the changes on this PR, 1.10.1 would have been
skipped over, leading to 1.10.0 being installed instead.

### DIFF
--- a/jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
+++ b/jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
@@ -34,6 +34,11 @@ def get_latest_master_version(branch)
     branch = "master"
   end
 
+  # Scrape the puppetserver repo page for available puppetserver builds and
+  # filter down to only ones matching the specified branch.  The list of builds
+  # is ordered from most recent to oldest.  The resulting list is passed along
+  # to the get_cent7_repo routine, which returns a URL for the first build which
+  # has a cent7 repo in it.
   get_cent7_repo(
       response.lines.
           select { |l| l =~ /<td><a / }.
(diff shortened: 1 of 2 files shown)


jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
```diff
@@ -3,51 +3,53 @@
...
-    next unless l =~ /<td><a /
-    match = l.match(/^.*href="(\d+\.\d+\.\d+)\/\?C=M&amp;O=D".*$/)
-    next unless match
-    return match[1]
-  end
+  get_cent7_repo(
+      response.lines.
+          select { |l| l =~ /<td><a / }.
+          select { |l| l.match(/^.*href="(\d+\.\d+\.\d+)\/\?C=M&amp;O=D".*$/) },
+      "puppet-agent")
```
jpinsonault: I think I get what's happening here, but it might be nice to have a comment briefly explaining how this call to `get_cent7_repo` results in what we want

jpinsonault:
I'm 👍 , just had one minor comment

jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
camlow325: Makes sense.  I added a couple of comments in 514d949.

PR #117, opened 2016-12-15
(PDB-3209) Add a PDB metrics collection script

senior:
This commit adds a new background script that will collect all PDB JMX
metrics and dump them to a compressed file, similar to how metrics are
collected for puppetserver.

Note that this is currently a work in progress. It won't work right now due to https://github.com/puppetlabs/trapperkeeper-metrics/pull/39. Once that's merged, released, lein parent bumped and puppetdb is bumped to that new version and released, this should be able to run.

PR #125, opened 2017-05-05
(PDB-3209) Add a PDB metrics collection script

camlow325:
This PR adds a new background script that will collect all PDB JMX
metrics and dump them to a compressed file, similar to how metrics are
collected for puppetserver.


camlow325:
This is a replacement for #117.  Ping @rlinehan and @jpinsonault for review.

rlinehan:
This all looks good to me, code-wise. I'm currently testing it out locally/on some vmpooler instances.

jpinsonault:
This looks good to me

jenkins-integration/beaker/install/shared/install_deps.rb
```diff
@@ -0,0 +1,9 @@
+step "Configure epel" do
+  tmp_module_dir = master.tmpdir('configure_epel')
+  on(master, puppet('module', 'install', 'stahnma-epel', '--codedir', tmp_module_dir))
+  on(master, puppet('apply', '-e', "'include epel'"))
```
rlinehan: I keep getting 

``` 
kmzh191du28xs3v.delivery.puppetlabs.net 18:35:09$ puppet apply -e 'include epel'
    Error: Evaluation Error: Error while evaluating a Function Call, Could not find class ::epel for kmzh191du28xs3v.delivery.puppetlabs.net  at line 1:1 on node kmzh191du28xs3v.delivery.puppetlabs.net
  
  kmzh191du28xs3v.delivery.puppetlabs.net executed in 2.10 seconds
  Exited: 1
  Beaker::Host::CommandFailure: Host 'kmzh191du28xs3v.delivery.puppetlabs.net' exited with 1 running:
   puppet apply -e 'include epel'
  Last 10 lines of output were:
  	[1;31mError: Evaluation Error: Error while evaluating a Function Call, Could not find class ::epel for kmzh191du28xs3v.delivery.puppetlabs.net  at line 1:1 on node kmzh191du28xs3v.delivery.puppetlabs.net
```

when running this.

It works if I do `puppet apply -e 'include epel' --codedir tmp_module_dir`.

Why are we using a tmp module dir though? And should we be cleaning it up? I've done 4 or 5 installs on this box now, and I've got a number of `tmp/configure_epel.<whatever>` directories now.

camlow325: Oh, hm.  Yeah, not sure how I missed that.  Adding the "--codedir tmp_module_dir" would seem to make sense here - will do that.

I was using a tmp_module_dir mostly because I didn't want this to go into the normal production code dir that might be on the SUT - potentially conflicting with the code deployment that the specific test might do.  Does that sound reasonable?

I'd be okay with nuking the tmp_module_dir on completion.  Maybe I could just have it nuke the directory if the step was successful?  Might be useful to keep it around in the case of a failure for troubleshooting after the fact.

rlinehan: Nuking in the case of success makes sense, and yeah, I think it's probably reasonable to try to keep this from interfering with normal production code.

camlow325: @rlinehan Thanks again for finding this.  I added b031427.  I didn't run that back through a full Gatling job setup but I did use the beaker test in this repo to install puppet and then ran the install_deps.rb test.  The test passed for me, installing EPEL as expected and nuking the temp module directory when done.

rlinehan:
Tried this out again, it works now, looks good to me!

PR #126, opened 2017-05-25
(SERVER-1776) job to compare jruby 1.7 vs 9k

jpinsonault:
* Create job to compare oss-latest with jruby 1.7 and 9k
* Create a job for different numbers of agents to see how jruby 9k
affects performance at different loads

### DIFF
--- a/jenkins-integration/beaker/install/shared/80_configure_java_args.rb
+++ b/jenkins-integration/beaker/install/shared/80_configure_java_args.rb
@@ -2,27 +2,8 @@ require 'puppet/gatling/config'
 
 test_name 'Configure java args'
 
-def service_config_name(service_name)
-  "/etc/sysconfig/#{service_name}"
-end
-
-def set_bash_variable(host, filepath, variable_name, value)
-  manifest = <<-MANIFEST
-ini_setting { "#{variable_name}":
-  ensure  => present,
-  path    => "#{filepath}",
-  section => "",
-  setting => "#{variable_name}",
-  key_val_separator => "=",
-  value   => "#{value}",
-}
-  MANIFEST
-
-  on host, puppet('apply', '-e', "'#{manifest}'")
-end
-
 service_name = get_puppet_server_service_name_from_env()
 service_config = service_config_name(service_name)
 java_args = get_puppet_server_java_args_from_env()
 
-set_bash_variable(master, service_config, 'JAVA_ARGS', java_args)
+set_service_environment_variable(master, service_config, 'JAVA_ARGS', java_args)
--- a/jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
+++ b/jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
@@ -1,6 +1,6 @@
 require 'puppet/gatling/config'
 
-test_name 'Configure java args'
+test_name 'Configure jruby jar path'
 
 def service_config_name(service_name)
   "/etc/sysconfig/#{service_name}"
@@ -21,8 +21,9 @@ ini_setting { "#{variable_name}":
   on host, puppet('apply', '-e', "'#{manifest}'")
 end
 
+jruby_jar = ENV['PUPPET_GATLING_JRUBY_JAR']
+
 service_name = get_puppet_server_service_name_from_env()
 service_config = service_config_name(service_name)
-java_args = get_puppet_server_java_args_from_env()
 
-set_bash_variable(master, service_config, 'JAVA_ARGS', java_args)
+set_bash_variable(master, service_config, 'JRUBY_JAR', jruby_jar)
--- a/jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
+++ b/jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
@@ -2,28 +2,8 @@ require 'puppet/gatling/config'
 
 test_name 'Configure jruby jar path'
 
-def service_config_name(service_name)
-  "/etc/sysconfig/#{service_name}"
-end
-
-def set_bash_variable(host, filepath, variable_name, value)
-  manifest = <<-MANIFEST
-ini_setting { "#{variable_name}":
-  ensure  => present,
-  path    => "#{filepath}",
-  section => "",
-  setting => "#{variable_name}",
-  key_val_separator => "=",
-  value   => "#{value}",
-}
-  MANIFEST
-
-  on host, puppet('apply', '-e', "'#{manifest}'")
-end
-
 jruby_jar = ENV['PUPPET_GATLING_JRUBY_JAR']
-
 service_name = get_puppet_server_service_name_from_env()
 service_config = service_config_name(service_name)
 
-set_bash_variable(master, service_config, 'JRUBY_JAR', jruby_jar)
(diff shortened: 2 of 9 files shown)


jpinsonault:
Finished a run of the 1.7 vs 9k job:
screen shot 2017-05-25 at 2 14 17 pm

Running the other comparison job now on the dev driver

jenkins-integration/beaker/install/shared/80_configure_java_args.rb
```diff
@@ -1,38 +1,28 @@
...
-end
-
-def update_java_args_in_local_config(java_args, local_service_config_path)
-  # Replace the existing JAVA_ARGS in the config
-  ini_file = IniFile.load(local_service_config_path)
-  ini_file['global']['JAVA_ARGS'] = java_args
-  ini_file.save
+def set_bash_variable(host, filepath, variable_name, value)
+  manifest = <<-MANIFEST
+ini_setting { "#{variable_name}":
```
camlow325: One thing that's maybe a bit tricky about this is that I'm not sure for an arbitrary scenario that you're guaranteed to have had `puppetlabs-inifile` installed into the Puppet code directory.  This probably works when we use the `puppetlabs-puppetserver_perf_control` repo for a simulation because its Puppetfile includes `puppetlabs-inifile`.

It might be better to do something like what I did with configuring the `epel` repo [here](https://github.com/puppetlabs/gatling-puppet-load-test/blob/8c5c5a6b557e96a88df375510017e739abfbc2a7/jenkins-integration/beaker/install/shared/install_deps.rb#L2-L5), where we install the module to a temp location (which doesn't potentially pollute the Puppet code dir for the scenario under test) before doing the `puppet apply`.

jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
```diff
@@ -0,0 +1,29 @@
+require 'puppet/gatling/config'
+
+test_name 'Configure jruby jar path'
+
+def service_config_name(service_name)
```
camlow325: Seems like it would be nice to put the following methods into the ./lib/puppet/gatling/config.rb helper file so that the configure_java_args and configure_jruby_jar code could reuse most of the implementation:

* service_config_name
* set_bash_variable
* call to get_puppet_server_service_name_from_env
* call to service_config_name

Maybe that way, this file could just be shortened to:

```ruby
require 'puppet/gatling/config'

test_name 'Configure jruby jar path'

set_bash_variable(master, service_config, 'JRUBY_JAR', ENV['PUPPET_GATLING_JRUBY_JAR'])
```

And the java_args one could be shortened to something similar.  WDYT?

jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
```diff
@@ -0,0 +1,29 @@
...
+  "/etc/sysconfig/#{service_name}"
+end
+
+def set_bash_variable(host, filepath, variable_name, value)
```
camlow325: Maybe it would be better to name this function something like `set_service_environment_variable` - to indicate that the value you are setting is an environment variable that will be used by the different service commands?

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -284,6 +284,17 @@ def step080_customize_java_args(script_dir, server_java_args, server_era) {
     }
 }
 
+def step081_customize_jruby_jar(script_dir, jruby_jar, server_era) {
+    if ((jruby_jar == null) || (jruby_jar == "")) {
+        echo "Skipping jruby jar configuration because none specified in job"
```
camlow325: It might be good to go ahead and set the `JRUBY_JAR` variable to empty in the sysconfig file even if the configuration didn't specify a value for it.  For cases where you might be doing iterative work where you flip back and forth between `JRUBY_JAR` values on the same VM, it might be good to have the piece of mind that the value would be set from 9k back to the default in the case you didn't reinstall server components between test runs.  A value of `JRUBY_JAR=` in the sysconfig file is effectively the same as not having it there at all and for versions of puppetserver that don't look at the variable, having it there shouldn't be a problem.

jenkins-integration/beaker/install/shared/80_configure_java_args.rb
jpinsonault: crafty, that seems like a better idea

jenkins-integration/beaker/install/shared/81_configure_jruby_jar.rb
jpinsonault: Yeah I think that'd be good

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
jpinsonault: Yeah I think that'd be fine. I was initially thinking that it should be managed explicitly by the groovy job definitions, but yeah having it clear the variable between jobs is probably what a user would expect/want

jenkins-integration/beaker/install/shared/80_configure_java_args.rb
```diff
@@ -1,38 +1,28 @@
...
+  ensure  => present,
+  path    => "#{filepath}",
+  section => "",
+  setting => "#{variable_name}",
+  key_val_separator => "=",
+  value   => "#{value}",
+}
+  MANIFEST
```
camlow325: I tried using the following in my scenario `Jenkinsfile`:

```
server_java_args: "-Xms3g -Xmx3g",
```

That ended up being applied in the /etc/sysconfig/puppetserver file like this:

```
JAVA_ARGS=-Xms3g -Xmx3g
```

The puppetserver service seems to be started with this just fine - including `/usr/bin/java -Xms3g -Xmx3g...` in the command line.  When I run a puppetserver CLI command with that, though, I get this error:

```
[root@k2ziz2dkclswuvi ~]# /opt/puppetlabs/server/bin/puppetserver ruby --version
/etc/sysconfig/puppetserver: line 9: -Xmx3g: command not found
jruby 9.1.9.0 (2.3.3) 2017-05-15 28aa830 OpenJDK 64-Bit Server VM 25.131-b12 on 1.8.0_131-b12 +jit [linux-x86_64]
```

I think it would be good to figure out how to apply double-quotes around the setting value to avoid possible strange behavior in how the variables get used.

I think it would be

camlow325:
As we discussed IRL, the points on the graph for 1.7 and 9k for the mean agent run times both seem a lot higher than what I'm used to seeing.  One possibility for that might have been that this was captured after a relatively short run, not sure.  I see significantly higher numbers when doing just a single run for a small number of agents - where the startup costs for the first set of requests handled by JRuby instances are much higher than that for subsequent runs.  For longer runs - e.g., 2 hours or more - against our perf blade hardware, though, I would at least expect the mean agent run time for JRuby 1.7 to be more in the 4 - 6 second range instead of 1000 - 1200 seconds, like these.

jpinsonault:
Pushed some commits to address feedback. Running some tests with different heap sizes to see how that affects the times

camlow325:
The latest changes look good to me so far.  I'll do one more re-run with my gatling driver and an SUT just to confirm.

PR #129, opened 2017-06-15
(SERVER-1802) gatling recording with json puppet agent

jpinsonault:
This PR adds a recording of a puppet 5 agent using json for requests made to a puppetserver
It also includes a few new jobs:
* Jruby 9k
* Jruby 9k with json
* Jruby 1.7 with json


camlow325:
I think it might be good to name the new agent recording files as FOSS5x... instead of FOSS25x... - just to differentiate that they were captured from a Puppet 5 agent/server vs. the prior ones we captured originally with a Puppet Server 2.5.x server.  Might be better to rename the existing ones to FOSS2x now to make that distinction more clear but we could do that separately from this PR.

camlow325:
Hmm, the headers don't look right to me for the catalog request in the scala file.  It uses `headers_0`, which is defined as:

```scala
val headers_0 = Map("X-Puppet-Version" -> "5.0.0")
```

I was expecting it to look more like:

```scala
val headers_X = Map(
	"X-Puppet-Version" -> "5.0.0",
	"Accept" -> "application/json",
        "Accept-Encoding" -> "gzip;...",
	"Content-Type" -> "application/x-www-form-urlencoded")
```

I think the "Accept" and "Accept-Encoding" header values could make a significant difference in the run results.

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-jruby9k-json/Jenkinsfile
```diff
@@ -0,0 +1,30 @@
+node {
```
camlow325: We could just call this one latest-jruby9k, I think, and not work "json" into the name since we'll probably just stick with json always for Puppet 5-based runs.

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-json/JobDSL.groovy
```diff
@@ -0,0 +1,14 @@
+// NOTE: we only want to cron this job and use a 'real' SUT hostname if the job
+// is running on a production server.
+if (serverConfig["environment"] == "production") {
+    job.with {
+        triggers {
+            // This should run the job at a semi-random time between 9:00 and 10:59PM,
+            //  on Mondays.
+            cron('H H(21-22) * * 1')
```
camlow325: We probably don't want this to run at this time and with sut54, right, since this could conflict with the [existing oss-puppetserver-latest job](https://github.com/puppetlabs/gatling-puppet-load-test/blob/e5bad173cd35340c04debee485e5d367eb6fdf48/jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest/JobDSL.groovy#L8-L12)?

camlow325:
Thanks for including all of the different Jenkinsfile scenarios.  Here are some thoughts on the ones that might be most valuable to commit for this PR (some of which would involve some work beyond just the Jenkinsfiles):

1) oss-puppetserver-latest - Similar to what it is today but with the following changes:

- Start using the JSON agent recording instead of the FOSS25x legacy (PSON) recording.
- Continue using the implicit "latest" as the puppet `agent_version` but rework the helpers in the [30_install_dev_repos.rb](https://github.com/puppetlabs/gatling-puppet-load-test/blob/e5bad173cd35340c04debee485e5d367eb6fdf48/jenkins-integration/beaker/install/foss/30_install_dev_repos.rb) file to have "latest" mean the actual latest puppet-agent version which is currently available - latest SHA/merge/nightly build available on builds.puppetlabs.lan.  I think it will still be good to have the ability to pull the last released version of puppet-agent for the "oss-puppetserver-stable" job.  For that, I was thinking we could introduce a new "latest-release" designation that would be used for the `agent_version` in this case.

2) oss-puppetserver-stable - Same as today but switched to use the new "latest-release" designation for puppet-agent instead of the old "latest" designation.

3) oss-puppetserver-latest-jruby9k (just using the new JSON recording) - This would be identical to the one you called oss-puppetserver-latest-jruby9k-json and with it we could eliminate the one you called oss-puppetserver-latest-jruby9k.

4) oss-puppetserver-latest-json-vs-pson - Multipass pipeline job, JRuby 1.7 based, which runs one job with the FOSS25x legacy (PSON) recording and another job with the new JSON recording.

5) oss-puppetserver-latest-jruby9k-json-vs-pson - Same as the oss-puppetserver-latest-json-vs-pson job but using JRuby 9k instead of JRuby 1.7.

6) oss-puppetserver-latest-jruby9k-various-heaps - I think for this one that it would be fine to just have the one that has runs with the JSON recording but not worry about pulling in the one that uses the legacy/PSON recording.

7) oss-puppetserver-latest-jruby-1.7-vs-9k - I don't see this one currently on the PR but I think you have been experimenting with this one separately.  For this one, a multipass job with two jobs - one for JRuby 1.7, one for JRuby 9k - where puppet-agent is set to latest and the JSON recording is being used would sound reasonable.  (If you were thinking of doing this one as a separate PR, I'd be good with that.)

I think that among these OSS jobs, just having the oss-puppetserver-latest one continue to be the one scheduled job would be fine for now.

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-jruby9k-json/Jenkinsfile
jpinsonault: Makes sense

jpinsonault:
I think I agree with all your suggestions on jobs and renaming things

re: headers, it was only the report that was causing the server to respond with 400's due to lack of headers, so I only added the updated headers, `headers_1`, to the report request. Looking at the catalog request at least, it has the content type listed under a form param, `.formParam("facts_format", "application/json")`

Looks like the pre-json recording does not have that form parameter. Not sure if that's an important difference?

camlow325:
> Looking at the catalog request at least, it has the content type listed under a form param, .formParam("facts_format", "application/json")

The `Content-Type` header for the full request is different than the `facts_format` parameter.  The `facts_format` one just describes the format of the `facts` parameter which follows it.  It appears that that is getting set in the existing FOSS25x recording, [here](https://github.com/puppetlabs/gatling-puppet-load-test/blob/e5bad173cd35340c04debee485e5d367eb6fdf48/simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS25xPerfMedium.scala#L49).  It is expected that this would have a value of "pson" instead of "application/json" for a pre-Puppet 5 recording.

The `Content-Type` header for a catalog request should be set to `application/x-www-form-urlencoded`.  This is what I see both when I run a Puppet 4 agent and a Puppet 5 agent, so I'm not sure why we don't see that in the FOSS25x recording.  That could be due to us having taken the recording before the agent team tightened up some of the Content-Type / Accept header handling in later Puppet releases, not sure.  Or I guess this could be showing that Gatling's recorder is just ignoring some headers completely for some reason?

jpinsonault:
I pushed some commits that add/modify jobs. I believe that covers it all. Some use "latest" for the agent version and some use "latest-release". So that work will have to be done before this is merged

camlow325:
@jpinsonault I put up #131 with some changes to support the "latest" vs. "latest-release" distinction.  Seems like it would be good to let that one land before this one?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala
```diff
@@ -0,0 +1,500 @@
...
+// 		.baseURL("https://${node}:8140")
+
+	val reportBody = ELFileBody("FOSS5xPerfMedium_0127_request.txt")
+
+	val headers_0 = Map("X-Puppet-Version" -> "5.0.0")
+	val headers_1 = Map(
+		"X-Puppet-Version" -> "5.0.0",
+		"Accept" -> "application/json",
+		"Content-Type" -> "application/json",
+		"Connection" -> "close")
```
camlow325: I think it would be good to include `Accept-Encoding` header for this, as a real agent would, so that we can have the server exercise gzip encoding for the response bodies it produces.  Here's what my agent sends for this:

```
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
```

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala
```diff
@@ -0,0 +1,500 @@
...
+			.get("/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&source_permissions=use&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&checksum_type=md5")
+			.headers(headers_0))
+		.pause(736 milliseconds)
+		.exec(http("filemeta plugins")
+			.get("/puppet/v3/file_metadatas/plugins?environment=production&links=follow&recurse=true&source_permissions=ignore&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&checksum_type=md5")
+			.headers(headers_0))
+		.pause(1)
+		.exec(http("catalog")
+			.post("/puppet/v3/catalog/${node}?environment=production")
+			.headers(headers_0)
```
camlow325: Where this is using `headers_0` and `headers_0` doesn't include an `Accept` or `Accept-Encoding` header, I don't think the server will producing the gzipped JSON that we would expect it to be for runs we're doing.  Maybe we could create a new `headers_2` like the following for now and have the catalog request use that:

```scala
val headers_2 = Map(
		"X-Puppet-Version" -> "5.0.0",
		"Accept" -> "application/json",
                "Accept-Encoding" -> "gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
		"Content-Type" -> "application/x-www-form-urlencoded")

File: simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala [hunk 5]
```diff
@@ -0,0 +1,500 @@
...
...102f2ab6843e8f993b37\nSSHFP 4 2 27fec0970e79f7875be3c87a2be118c9b8cdb64f05f6ae5a8aeec7cb4b2bcebc","sshfp_rsa":"SSHFP 1 1 b7caf8ce4a295676009b49c012bae99f37bdfd1a\nSSHFP 1 2 1d645d15c047ac7a3e908f0bd7d6dc23b59f352182c126816236e9c055ac7ebf","sshrsakey":"AAAAB3NzaC1yc2EAAAADAQABAAABAQCn3YlHJ1y9vQxCijwF3MpKg4t3uMwFRFHWSeBxvI7Qh1Uo6rqPogn/5w5SaNnQrMISifadoOHz0Sg+5ZwLUQDpMxSbS6kx0BWr6ZLiDlk8nu58A3njCCSwmTJjMoqoRxQLoCFq6ECYs81Vu8UFUbtgujuSSEuoCGxYrI4j/zMIT3iZnhI69esYk/NuK7SxKfECUwXDRsCArVqaGNpqd56/uOHmjg+8xgSko1AypCGh+lx6Ghg5F7bXyQRsRFlXtvv2h9Fjv9lIwidSNneA14HfAMfcT07VE6sLnwUONHAD4X52iNxarAa/bAQxhqUVvlH/7RN5SaV/aDKZEURXiPEb","stage":"prod","staging_http_get":"curl","swapfree":"1000.00 MiB","swapfree_mb":999.99609375,"swapsize":"1000.00 MiB","swapsize_mb":999.99609375,"system_uptime":{"days":0,"hours":0,"seconds":1296,"uptime":"0:21 hours"},"timezone":"PDT","uptime":"0:21 hours","uptime_days":0,"uptime_hours":0,"uptime_seconds":1296,"uuid":"37323233-3135-5553-4533-34364C334B4A","virtual":"physical","whereami":"portland","clientcert":"${node}","clientversion":"5.0.0","clientnoop":false},"timestamp":"2017-06-14T17:17:27.391789175-07:00","expiration":"2017-06-14T17:47:27.392052206-07:00"}""")
```
camlow325: In the most recent Puppet 5 agent code, the fact payload now has an extra layer of URL encoding done, as mentioned in https://github.com/puppetlabs/puppet/pull/5999.  It would seem good to just re-record this with the most recent Puppet 5 agent code before applying the other request header edits from above.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala
```diff
@@ -0,0 +1,500 @@
...
+	val headers_1 = Map(
+		"X-Puppet-Version" -> "5.0.0",
+		"Accept" -> "application/json",
+		"Content-Type" -> "application/json",
+		"Connection" -> "close")
+// val uri1 = "https://${node}:8140/puppet/v3"
+
+	val chain_0 = exec(http("node")
+			.get("/puppet/v3/node/${node}?environment=production&transaction_uuid=122098b0-4ca4-4b70-ba22-a6fdf331b352&fail_on_404=true")
+			.headers(headers_0))
```
camlow325: Hmm, the `node` and `file_meta` requests should also be including `Accept: application/json` and `Accept-Encoding: gzip` like the `catalog` and `report` ones do.  I think we should add those into `headers_0`, like:

```scala
val headers_0 = Map(
		"X-Puppet-Version" -> "5.0.0",
		"Accept" -> "application/json",
                "Accept-Encoding" -> "gzip;q=1.0,deflate;q=0.6,identity;q=0.3")
```

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-jruby-1.7-vs-9k/Jenkinsfile
```diff
@@ -27,11 +30,14 @@ pipeline.multipass_pipeline([
         ],
         [
                 job_name: 'oss-latest-jruby-9k',
```
camlow325: Do you know if our Jenkins could have collisions if we use the same `job_name` across different pipelines?  This one's name I guess is technically different than the `oss-latest-jruby9k` one in the `oss-puppetserver-latest-jruby9k` scenario but the two are pretty similar.  If pipeline job names can't collide, I think this is fine.  If not, though, we might name this one a little differently to make it less likely to collide with another job_name, like `oss-latest-jruby-9k-for-1.7-comparison` or something.  naming--

camlow325:
I think it might also be nice to add a `foss5x-medium-3.json` scenario file, like the `foss25x-medium-3.json` one, for convenience in running a short job for VM testing, something like:

```json
{
    "run_description": "'medium' role from perf control repo",
    "nodes": [
        {
            "node_config": "FOSS5xPerfMedium.json",
            "num_instances": 3,
            "ramp_up_duration_seconds": 0,
            "num_repetitions": 1,
            "sleep_duration_seconds": 1
        }
    ]
}
```

camlow325:
My tests with a couple of these jobs on a dev gatling driver seemed to work well, so I'm good with this pending some updates to the Gatling agent recordings for the header issues and maybe adding a short run FOSS5x scenario file.

jpinsonault:
Rebased onto https://github.com/puppetlabs/gatling-puppet-load-test/pull/131

Updated recording using latest agent builds. Looks like the facts encoding has been updated

Created the updated headers

Added new short 5.x scenario

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-jruby-1.7-vs-9k/Jenkinsfile
jpinsonault: I wouldn't be surprised if they would conflict, but I'm not sure. Updated the names either way

camlow325:
Looks good to me.  I'm going to do one more short run with the new agent recording on a dev VM to make sure it looks sane.  @adrienthebo or @justinstoller - one of you want to take a quick look at this as well?

simulation-runner/user-files/bodies/FOSS5xPerfMedium_0127_request.txt
```diff
@@ -0,0 +1 @@
...sers","role::by_size::medium","role","by_size","medium","node","__node_regexp__og8iwz6f52a9ig8."],"time":"2017-06-26T18:09:58.742080001+00:00","failed":false,"changed":false,"out_of_sync":false,"skipped":false,"change_count":0,"out_of_sync_count":0,"events":[],"corrective_change":false},"Ssh_authorized_key[citizen3@citizen3]":{"title":"citizen3@citizen3","file":"/etc/puppetlabs/code/environments/production/dist/userwrapper/manifests/user.pp","line":31,"resource":"Ssh_authorized_key[citizen3@citizen3]","resource_type":"Ssh_authorized_key","containment_path":["Stage[main]","Profile::Users","Userwrapper::User[citizen3]","Ssh_authorized_key[citizen3@citizen3]"],"evaluation_time":0.000299282,"tags":["ssh_authorized_key","userwrapper::user","userwrapper","user","citizen3","class","profile::users","profile","users","role::by_size::medium","role","by_size","medium","node","__node_regexp__og8iwz6f52a9ig8."],"time":"2017-06-26T18:09:58.743165314+00:00","failed":false,"changed":false,"out_of_sync":false,"skipped":false,"change_count":0,"out_of_sync_count":0,"events":[],"corrective_change":false},"User[citizen4]":{"title":"citizen4","file":"/etc/puppetlabs/code/environments/production/dist/use
```
justinstoller: wtf

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-jruby9k-json-vs-pson/Jenkinsfile
```diff
@@ -0,0 +1,57 @@
...
+                        "./jenkins-jobs/common/scripts/background/curl-server-metrics-loop.sh"
+                ],
+                archive_sut_files: [
+                        "/var/log/puppetlabs/puppetserver/metrics.json"
+                ],
+                jruby_jar: "/opt/puppetlabs/server/apps/puppetserver/jruby-9k.jar",
+        ],
+        [
+                job_name: 'oss-latest-jruby9k-pson',
+                gatling_simulation_config: '../simulation-runner/config/scenarios/foss25x-medium-1250-2-hours.json',
```
justinstoller: Sorry for the dumb question, but why does this use a `foss25` simulation?

camlow325: Belated answer on this one...  The main thing we're looking for in this comparison is how the use of PSON serialization compares to the use of JSON serialization in agent->master communication.  By default, a Puppet 5 agent will try use to use JSON.  We could do a re-recording of a Puppet 5 agent where the preferred serialization format is set to PSON, but effectively the existing FOSS 25x already gets us PSON by default.  FOSS25x was captured from a Puppet 4 agent run.  Since Puppet 5 is backward compatible with Puppet 4 agents on the wire, I think just reusing the existing FOSS25x recording is reasonable here.

justinstoller: That makes sense, and re-reading the discussion I eventually came to understand it similarly. I think the biggest dissonance for me was the 2x/5x describing the server that it was meant to test against, rather than the agent version whose requests it was emulating (thinking about it again there was no way we would have a 25x agent anyways).

Slightly orthogonally, I know the puppet apis we call are versioned (v3, right?), would it be valid to call this whole agent/server transaction versioned along with those apis?

camlow325: Versioning these solely per the Puppet APIs, i.e., /puppet/v3/catalog, may not work here.  For example, for the PSON->JSON migration, /puppet/v3 is used for both but only the use of HTTP headers, etc., changed.

Might be a bit more verbose, but we could name these files with both versions of the master and agent - e.g., foss-agent-4x-master-5x-...  We could also qualify the names if necessary, like foss-agent-4x-master-5x-pson...  I still like being able to name the files such that the intention can be derived without a separate keyring, if possible.  Open to any specific suggestions you have on naming, though, since we definitely haven't had a very consistent / understandable strategy around these in the past.

justinstoller: MMm... all good, I don't think it's probably worth the effort of re-articulating a new naming scheme for the files. I guess I'm just learning the correlations of the naming scheme to the what axes we care to test.

PR #130, opened 2017-06-21
(SERVER-1843) Update keypair used for github access

camlow325:
This commit updates the process used to setup the keypair for the
gatling driver when accessing github.


camlow325:
I tested this PR with the Jenkins dev driver setup and SUT in vmpooler instances with the OSS puppetserver stable job.  The job ran properly with the new credentials.

We should hold off on merging this until https://github.com/puppetlabs/puppetlabs-modules/pull/7225 lands, since the related PR controls razor/SUT credential setup that we use for the "production" runs on hardware.

camlow325:
The related puppetlabs-modules PR has been merged so this one should also be ready for merge now.

PR #131, opened 2017-06-22
(SERVER-1862) Introduce latest vs latest-release agent_version config

camlow325:
This commit changes the agent_version parameter for OSS jobs to use a
distinction between 'latest' (latest successful puppet-agent build from
the master branch) vs. 'latest-release' (latest release of puppet-agent).
The oss-puppetserver-stable job now uses the 'latest-release' value for
puppet-agent whereas the other oss-puppetserver jobs continue to
implicitly use the default value of 'latest' (via not specifying a
discrete value for agent_version).

### DIFF
--- a/jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
+++ b/jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
@@ -46,6 +46,25 @@ def get_latest_master_version(branch)
 end
 
 def get_latest_agent_version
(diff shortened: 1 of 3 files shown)


jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
```diff
@@ -46,6 +46,25 @@ def get_latest_master_version(branch)
 end
 
 def get_latest_agent_version
+  url = "https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent%20suite%20pipelines/job/platform_puppet-agent_intn-van-promote_suite-daily-promotion-master/lastSuccessfulBuild/api/json"
```
camlow325: This is the same approach we would be using in OSS puppetserver to determine the latest puppet-agent version per https://github.com/puppetlabs/puppetserver/pull/1432.

PR #132, opened 2017-06-26
(SERVER-1861) Upgrade gatling to 2.2.5

jpinsonault:
This PR updates gatling to the latest release version and updates the recordings to be compatible

The only breaking change I encountered was that gatling 2.2.0 renamed `ELFileBody` to `ElFileBody`

I've tested this with the oss-puppetserver-latest/stable jobs and they seem to be running normally. I'll make sure the jobs don't have any unexpected performance changes once they finish though.

The gatling recorder app also seems to be working fine


jpinsonault:
Should also mention that I pushed this to the dev driver and ran it, so if there was a java version problem (gatling 2.2.5 only supports java 8 I believe), I think it would have shown up

camlow325:
@doug-rosser Heads up that we're probably going to merge this soon.  I expect you may need to update your old recordings like we have in order to avoid having them break.  Not sure where you maintain those...

jpinsonault:
Did a recording and played it back with gatling 2.2.5. It seems to work fine. The only issue is that the post processor adds in a call to the old ELFileBody and I had to change that. Will have to update the post processing script

camlow325:
> so if there was a java version problem (gatling 2.2.5 only supports java 8 I believe), I think it would have shown up

Hmm, it looks like our production dev driver, puppetserver-perf-driver55, only has Java 7 on it, not Java 8.  Also, a dev driver that I setup using the scripts in the repo only had Java 7 on it too, not Java 8.  puppetserver-perf-driver68-dev.delivery.puppetlabs.net, however, does have both Java 7 and Java 8 on it, with Java 8 being the default.  Not sure why that is, but we may need to look into our JDK setup for the dev and production nodes to ensure that they'll pull in Java 8.

camlow325:
On my dev driver, I also get the error that we usually see when trying to run code built for Java 8 running on Java 7:

```
Running com.puppetlabs.gatling.runner.PuppetGatlingRunner 
Exception in thread "main" java.lang.UnsupportedClassVersionError: io/gatling/core/scenario/Simulation : Unsupported major.minor version 52.0
```

So it does seem like sbt doesn't automagically pull in Java 8 as part of running the newer Gatling jars.

doug-rosser:
This is actually a really good time to make a major change as most of my existing recordings are meant to be thrown away.

camlow325:
@jpinsonault Did you intentionally add all of the "blah" commits to the PR?

jpinsonault:
ha. No I did not. Thank you

jpinsonault:
That's what happens when you forget to change branches. Rebased to remove those commits

camlow325:
Per my earlier comment, we separated out the work for configuring the Gatling driver with Java 8 into a new ticket, [SERVER-1871](https://tickets.puppetlabs.com/browse/SERVER-1871).  A [PR]( https://github.com/puppetlabs/puppetlabs-puppetserver_perf_driver/pull/17) is up for review in the perf_driver repo to add that to the perf driver configuration.  A related PR will need to be submitted to the puppetlabs-modules repo in order to bump the perf driver config up to a version which includes the changes. 

I think we should hold off on merging this one until the Java 8 changes have been merged and running in puppetlabs-modules.

jpinsonault:
Pushed some commits to update the post processing script. The new version of gatling puts common header items in the httpProtocol object with some helper methods. The script now pulls the headers out of those helper methods and creates a new scala map, and each header variable in the recording is merged on top of this new base header.

I have a simulation running right now using the latest recording (from this PR) after it was processed by the updated script. It's running fine so far, will update when finished

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala
```diff
@@ -12,26 +12,28 @@ import io.gatling.http.Predef._
...
+		"Accept-Encoding" -> "gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
+		"User-Agent" -> "Puppet/5.0.0 Ruby/2.4.1-p111 (x86_64-linux)")
 
-	val headers_127 = Map(
+	val headers_0 = baseHeaders ++ Map("X-Puppet-Version" -> "5.0.0")
+
+	val headers_127 = baseHeaders ++ Map(
 		"Content-Type" -> "application/json",
-		"X-Puppet-Version" -> "5.0.0",
-		"Connection" -> "close")
```
camlow325: Hmm, where did the `Connection` header go for this one?  Seems like we'd still need that.

jpinsonault: Bah, looks like the regex to do that was messed up by my changes. Nice catch

jpinsonault:
Updated the script and recording to get the Connection header back

jpinsonault:
Ran a gatling sim with the latest version of the recording and there were no problems that I could see.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala
```diff
@@ -13,48 +13,44 @@ import io.gatling.http.Predef._
...
+		.pause(355 milliseconds)
 		.exec(http("filemeta plugins")
 			.get("/puppet/v3/file_metadatas/plugins?environment=production&links=follow&recurse=true&source_permissions=ignore&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&checksum_type=md5")
 			.headers(headers_0))
-		.pause(1)
+		.pause(2)
 		.exec(http("catalog")
 			.post("/puppet/v3/catalog/${node}?environment=production")
-			.headers(headers_catalog)
+			.headers(headers_0)
```
camlow325: It looks like the `Content-Type` of `application/x-www-form-urlencoded` was lost for the catalog request.  Is that header present in the gatling proxy's raw recording?

proxy-recorder/process_gatling_recording.rb
```diff
@@ -221,55 +237,110 @@ def step5_update_extends(text)
...
+
+  http_protocol_start = lines.find_index {|line| line.match(/^.*val httpProtocol/)}
+
+  if http_protocol_start.nil?
+    puts "Can't find headers after httpProtocol definition"
+    exit 1
+  end
+
+  # +2 to skip .baseURL call
+  index = http_protocol_start + 2
```
camlow325: Maybe it would be better to just skip the first line rather than the first two - e.g., in the event that the `.baseURL` call were to start appearing somewhere below the second line in the future, like:

```scala
 val httpProtocol = http
 	.acceptHeader("application/json, text/pson")
 	.baseURL("https://puppetserver-perf-sut57.delivery.puppetlabs.net:8140")
 	...
```

If we did that and changed the logic to only add an entry into the `found_header_data` hash if the `header_name` is actually found in the `HELPER_TO_HEADER` hash, then I think we'd get the same result?

proxy-recorder/process_gatling_recording.rb
```diff
@@ -221,55 +237,110 @@ def step5_update_extends(text)
...
       retline = line
     end
     # This end check won't work with long chains
-    end_comment = true if line.match(/\..*\)/)
+
     retline
   end
   out_text.flatten.join
+
+  # puts text.gsub(multiline_http_protocol_regex, )
```
camlow325: Extraneous comment?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xPerfMedium.scala
camlow325: It's probably fine if it's not in the original recording.  I suppose Gatling has some logic that knows to insert `Content-Type: application/x-www-form-urlencoded` when `.formParam` method calls are chained into it - just curious.

camlow325: I confirmed when playing the recording attached to this PR via gatling that a `Content-Type: application/x-www-form-urlencoded` header was sent in the request.  So I think we're good here.

jpinsonault: Awesome, thanks for looking into that

proxy-recorder/process_gatling_recording.rb
jpinsonault: Yeah, I think that'd be more robust

jpinsonault:
Pushed a commit that makes the header extraction more robust

PR #133, opened 2017-07-01
(SERVER-1875) Add puppetserver 2.x and 5.0.x and remove stable jobs

camlow325:
This PR adds oss puppetserver jobs for the 2.x and 5.0.x branches.
In order to allow for the jobs to specify a version pattern which
matches the artifacts that each branch builds, this commit also
introduces support for interpreting the content of the version string as
a regular expression whenever it contains the text "SNAPSHOT" in it.  In
this case, the latest build matching the regular expression should be
installed.

This PR removes the oss puppetserver stable job.  The 'stable'
branch no longer exists for puppetserver.


jenkins-integration/beaker/install/foss/30_install_dev_repos.rb
```diff
@@ -42,7 +38,7 @@ def get_latest_master_version(branch)
   get_cent7_repo(
       response.lines.
           select { |l| l =~ /<td><a / }.
-          select { |l| l =~ /#{branch}/}, "puppetserver")
+          select { |l| l =~ /">.*#{version}.*\/<\/a>/}, "puppetserver")
```
justinstoller: jeebus

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-5.0.x/Jenkinsfile
```diff
@@ -0,0 +1,26 @@
+node {
+    checkout scm
+    pipeline = load 'jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy'
+}
+
+pipeline.single_pipeline([
+        job_name: 'oss-5.0.x',
```
justinstoller: Will this move to 5.1.x pretty quickly (perhaps in a month)? Would we want to make this a 5.x and change the regex to be `/5\.\d\.\d/`?

camlow325: Yeah, that's a good point.  I suppose we'll probably only have one "5.x" vs. one "master" branch active at a time so maybe just making this one a "5.x" instead of a "5.0.x" would be better.  I'll work up that change.

PR #135, opened 2017-10-20
(QA-3234) Move scale setup from pe_acceptance_tests to perf repo

samwoods1:
This is the first step in consolidating all of the perf related code into one repo.  I moved required Beaker setup and helpers here.
Also renamed jenkins-integration directory to setup.  Subsequent check-ins will remove jenkins related content and replace with beaker scripts.
Renamed existing install directory to old_install so we can see if there is anything worth keeping in there, if not it can be delted later.

### DIFF
--- a/rakefile
+++ b/rakefile
@@ -1,0 +1,97 @@
+require 'rototiller'
+
+task :default => :performance
+
+desc 'Run Performance setup'
+rototiller_task :performance do |t|
+  t.add_env({:name => 'ENVIRONMENT_TYPE', :message => 'Either gatling or clamps', :default => 'gatling'})
+
+  if ENV['ENVIRONMENT_TYPE'] == 'clamps'
+    default_tests = 'setup/install'
+  else
+    default_tests = 'setup/install/00_pre_install/,setup/install/10_install/00_install_pe.rb,setup/install/30_post_install/'
+  end
+
+  # env vars needed for Beaker
+  t.add_command do |command|
+    command.name = "bundle exec beaker"
+    command.add_env({:name => 'BEAKER_EXECUTABLE'})
+
+    command.add_option do |option|
+      option.name = '--hosts'
+      option.message = 'The configuration file that Beaker will use'
+      option.add_argument do |arg|
+        arg.add_env({:name => 'BEAKER_HOSTS'})
+      end
+    end
+    command.add_option do |option|
+      option.name = '--keyfile'
+      option.message = 'The SSH key used to access a SUT'
+      option.add_argument do |arg|
+        arg.name = "#{ENV['HOME']}/.ssh/id_rsa-acceptance"
+        arg.add_env({:name => 'BEAKER_KEYFILE'})
+      end
+    end
+    command.add_option do |option|
+      option.name = '--log-level'
+      option.message = 'The log level under which you want beaker to run'
+      option.add_argument do |arg|
+        arg.name = 'debug'
+        arg.add_env({:name => 'BEAKER_LOG_LEVEL'})
+      end
+    end
+    unless ENV['BEAKER_PRE_SUITE'] == ''
+      command.add_option do |option|
+        option.name = '--pre-suite'
+        option.message = 'Beaker pre-suite'
+        option.add_argument do |arg|
+          arg.name = default_tests
+          arg.add_env({:name => 'BEAKER_PRE_SUITE'})
+        end
+      end
+    end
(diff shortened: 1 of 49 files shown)


samwoods1:
The two helper libraries are the only place where I really don't like copying everything.  What I would like to see in the future, and something that @er0ck, @RandellP and I discussed a bit would be to move those helpers into beaker helper libraries that can be consumed by anyone. That way we don't have to keep unrelated code (like perf stuff) in pe_acceptance_tests, but can use those libraries in any repo we want.

samwoods1:
I reverted some of the previous changes.  I had moved around some existing files, I decided to keep them where they are for now, and just add the new stuff in this PR.  We can clean up the other stuff later, after we have proved that it is no longer needed.

er0ck:
can you amend your commit message with the (maint) header?

rakefile
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end
+    command.add_option do |option|
+      option.name = '--pre-suite'
+      option.message = 'Beaker pre-suite'
+      option.add_argument do |arg|
+        arg.name = default_tests
+        arg.add_env({:name => 'BEAKER_PRE_SUITE'})
+      end
+    end unless ENV['BEAKER_PRE_SUITE'] == ''
```
RandellP: end unless?  What does that do?

rakefile
```diff
@@ -0,0 +1,89 @@
...
+    command.add_option do |option|
+      option.name = '--helper'
+      option.message = 'Setup helper required for installing PE'
+      option.add_argument do |arg|
+        arg.name = 'setup/helpers/classification_helper.rb,setup/helpers/ldap_helper.rb'
+        arg.add_env({:name => 'HELPER_PATH'})
```
RandellP: Is this a standard names used in lots of other places?  Otherwise it doesn't seem to line up with the values, which are actually a list of helpers to load.  Plus it doesn't have "BEAKER" in front like the previous names

rakefile
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end unless ENV['HELPER_PATH'] == ''
+    command.add_option do |option|
+      option.name = '--preserve-hosts'
+      option.message = 'Whether to preserve hosts or not'
+      option.add_argument do |arg|
+        arg.name = 'always'
+        arg.add_env({:name => 'PRESERVE_HOSTS'})
```
RandellP: should this be BEAKER_PRESERVER_HOSTS?

rakefile
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end
+    command.add_option do |option|
+      option.name = '--type'
+      option.message = 'pe or foss'
+      option.add_argument do |arg|
+        arg.name = 'pe'
+        arg.add_env({:name => 'INSTALL_TYPE'})
```
RandellP: BEAKER_INSTALL_TYPE ?

rakefile
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end
+    puts command.to_str
```
RandellP: Maybe some text to example what is being outputted (in case it somehow gets messed up and isn't obvious...)
like "Command built from options: ..."

rakefile
```diff
@@ -0,0 +1,89 @@
+require 'rototiller'
+
+task :default => :acceptance
```
RandellP: I don't see an "acceptance" task defined...

rakefile
samwoods1: The unless applies to the entire block inside do/end.  This makes the parameter optional.

er0ck: it's like do/while in perl.  first iteration in loop is non-optional

RandellP: Sam explained it all to me.  Do/while is evil though. :)  The decision logic ideally should be at the top.  Because people read top down.  But that is just my opinion.  :)

samwoods1: Yeah, it's probably easier to parse in single line statements.  I moved this to the top.

er0ck: agreed.  one can often put the decision logic at the top so it's a bit easier to read and then use/update that variable inside the loop.  the same thing happens when you have a non-deterministic loop count that needs a break in the middle.  especially in a language that doesn't have a for(x;y;z) type syntax that makes the break easy to see up top <cough> ruby.  of course, that's because in ruby you are usually iterating over an array or hash

er0ck:
and i'm cool with the internal IPs stuff.

er0ck:
18 days.  merging...

PR #136, opened 2017-11-09
(maint) Amazon Updates

doug-rosser:
These updates allow simulations in Amazon ec2.


simulation-runner/config/scenarios/OpsWorksLrgPGSync.json
```diff
@@ -0,0 +1,12 @@
+{
+    "run_description": "role::by_size::medium, 240 agents, 480 second rampup, 0-240 second splay, once",
```
haus: I assume splay was manually configured in PE for this test case?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/OpsWorksLarge.scala
```diff
@@ -0,0 +1,76 @@
...
+// import io.gatling.jdbc.Predef._
+
+class OpsWorksLarge extends SimulationWithScenario {
+
+// 	val httpProtocol = http
+// 		.baseURL("https://aws-lrg-mom.us-west-2.compute.internal:8140")
+
+	val reportBody = ElFileBody("OpsWorksLarge_0005_request.txt")
+
+	val headers_0 = Map("X-Puppet-Version" -> "4.10.4")
```
haus: Was this recording using glisan or hoyt?

haus:
@doug-rosser It looks like all of the simulations you've checked in here are running 4.10.x agents. Was that intentional?

haus:
I guess that makes sense since most of the perf was run against glisan.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/OpsWorksLarge.scala
haus: Never mind, I'm remembering that most of the perf testing was against glisan.

samwoods1:
I'm 👍

PR #137, opened 2017-11-21
(PE-3353) Allow pe_dir to be an internal network address for AWS install

samwoods1:


config/options/pez.rb
```diff
@@ -0,0 +1,15 @@
+(eval File.read('config/options/defaults.rb')).merge({
+  :test_tag_and => 'risk:high',
+  :answers      => {
+    "puppet_enterprise::profile::master::r10k_remote" => "https://github.com/puppetlabs/control-repo.git",
+    "puppet_enterprise::profile::master::code_manager_auto_configure" => true
+  },
+  :pre_suite    => [
+    'setup/pez.rb',
```
doug-rosser: What are setup/pez.rb and setup/analytics.rb for? They don't exist in the local repo.

samwoods1: After looking at these files, we probably only need the ec2.yaml, I'll ditch these other 2 files.

doug-rosser:
:+1:

PR #142, opened 2017-12-08
(SERVER-2046) Add FOSS 5.x empty control repo gatling recording

Magisus:


simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
```diff
@@ -0,0 +1,71 @@
...
+import io.gatling.http.Predef._
+// import io.gatling.jdbc.Predef._
+
+class FOSS5xEmptyRepo extends SimulationWithScenario {
+
+// 	val httpProtocol = http
+// 		.baseURL("https://sbbr2c2w66jmcpm.delivery.puppetlabs.net:8140")
+// 		.acceptHeader("application/json, text/pson")
+// 		.acceptEncodingHeader("gzip;q=1.0,deflate;q=0.6,identity;q=0.3")
+// 		.userAgentHeader("Puppet/5.3.3 Ruby/2.4.2-p198 (x86_64-linux)")
```
justinstoller: Can we get rid of this comment?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
```diff
@@ -0,0 +1,71 @@
...
...AAB3NzaC1yc2EAAAADAQABAAABAQDFeLm%2B%2BlQ2XDRHcGuqOf%2B%2FhGGNiZy0MITk6AmqAszur3f344ZmrEH5tViqaNtw8JNe330ocUP2I1SKYnqdHExkHUcrPjfLpNrdForcMnAmc3Q%2Bn3DA98uyOWrnOu9FYegUuv7XaPK8uDokJAK3qiBu59XwAtysldpJ7kdQ2BAQpqIap%2FtgRpuRrNfpyscui5UfaD3dHvWuMHHxSk3a7oWBt7acm0IqKePxXeaSNX5IO2UUokAcm45jHHJzh7TnY9Q5PAe1XtdRESrB8ADGgDSgUhG4cjOnvBwMvfRLR29%2FAML3RDwk7r6z6ZPyaptifwm74gml%2FVXbeCWT0jAigFGn%22%2C%22stage%22%3A%22prod%22%2C%22swapfree%22%3A%22820.00%20MiB%22%2C%22swapfree_mb%22%3A819.99609375%2C%22swapsize%22%3A%22820.00%20MiB%22%2C%22swapsize_mb%22%3A819.99609375%2C%22system_uptime%22%3A%7B%22days%22%3A0%2C%22hours%22%3A5%2C%22seconds%22%3A20007%2C%22uptime%22%3A%225%3A33%20hours%22%7D%2C%22timezone%22%3A%22PST%22%2C%22uptime%22%3A%225%3A33%20hours%22%2C%22uptime_days%22%3A0%2C%22uptime_hours%22%3A5%2C%22uptime_seconds%22%3A20007%2C%22uuid%22%3A%2242311FBE-8C5D-3F1C-57B9-252DA7225CF7%22%2C%22virtual%22%3A%22vmware%22%2C%22whereami%22%3A%22portland%22%2C%22clientcert%22%3A%22${node}%22%2C%22clientversion%22%3A%225.3.3%22%2C%22clientnoop%22%3Afalse%7D%2C%22timestamp%22%3A%222017-12-08T11%3A30%3A42.109494978-08%3A00%22%2C%22expiration%22%3A%222125-12-08T12%3A00%3A42.109796191-08%3A00%22%7D")
```
justinstoller: Can we put the facts in an external file like we're doing with reportBody?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
Magisus: Probably. I think the script doesn't edit destructively. A lot of the other scala files have similar comments.

justinstoller: Yeah, this looks like the input to what creates baseHeaders below. I don't think it should be needed? /cc @haus

justinstoller:
Do we need a scenario json file to be able to run this from Jenkins?

Magisus:
Oh yeah, good point. I still need to add the job itself, yes.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
haus: agree.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
Magisus: I'm not sure... all of the other scala files all have the facts inline. I think the reportBody is in a special format that makes it different (based on what I could find about ELFileBody).

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
Magisus: Think it might be good to delete all of the commented lines in here?

Magisus:
So now I'm worried I messed something up with the recording: all the other jobs list "production" as the environment for code deploy ([example](https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-5.x/Jenkinsfile#L17)), but I ran the simulation with a different environment per the instructions in https://github.com/puppetlabs/puppetlabs-puppetserver_perf_control#how-to-use-this-repo-during-development, which means that the agent requests in the scala file in this PR are requesting the environment `20171208_empty_repo`. Should I change the environment in the Jenkinsfile, or is the role to apply solely controlled by the class in the node configs like [here](https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/simulation-runner/config/nodes/FOSS5xPerfMedium.json#L5)?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/FOSS5xEmptyRepo.scala
Magisus: That being said, I'm sure I could just read it in using whatever the regular file read function is in Scala. I'll take a look.

Magisus: If I can get that to work, I guess it might be worth looking into updating the script to do that as part of post-processing.

Magisus:
I've updated this to deploy the environment associated with the branch I made in the perf control repo, instead of production, based on what I could glean from the readmes about how the various pieces are supposed to work together. How willing would y'all be to merge this and then revert/make updates if it doesn't work right? It seems like setting up an env to run outside of Jenkins would be relatively time consuming (unless someone knows a good way to do it).

justinstoller:
I'd be fine with merging and rolling forward, though we probably want to give the dev workflow a try at some point.

IIRC, it was a matter of running [beaker like this](https://github.com/puppetlabs/gatling-puppet-load-test/tree/master/jenkins-integration/dev#running-beaker) and then logging into that jenkins and running the `refresh-gplt-jobs` job pointed to your fork/branch to create the pipeline, and worked surprisingly well. Not sure how to run the pipeline pointed at a different node than usual though (though if there wasn't another job going on it would be fine to just run it using the "production" hardware). Is that what you were seeing when looking at the READMEs?

Magisus:
Not exactly, I had been looking the doc from Sam Woods' presentation. I can give it a shot; I was mostly on the fence about where to put the various VMs, but if you think it's okay to just use the production hardware, I can do that.

justinstoller:
Yeah, I think a fair bit of the complexity of that setup was Ec2 specific (especially getting the recording working) neither of which you should need. I'd go pooler for the jenkins/driver box and use the hardware if there's not an easy to find toggle somewhere for running against puppet server nodes in the pooler.

Magisus:
Okay, I'll take a look.

Magisus:
The dev workflow appears to have acquired some warts in the last couple years since the tooling has been updated, and fixing it will probably require some separate tickets.

jenkins-integration/jenkins-jobs/common/scripts/job-steps/initialize_ruby_env.sh
```diff
@@ -4,7 +4,7 @@ echo ""
 echo "Setting Up Ruby Environment"
 echo ""
 source /usr/local/rvm/scripts/rvm
-rvm use 2.2.5 || exit 1
```
justinstoller: Is the version of ruby important? Do we just need 2.x? Just wondering if we could less tightly couple this script from the contents of puppetlabs-puppetserver_perf_driver? (or if that's even desirable)

jenkins-integration/jenkins-jobs/job_bootstrap.groovy
```diff
@@ -58,8 +58,8 @@ class DSLHelper {
 // accomplish the same by simply modifying the values in the Jenkins gui
 // on the 'configuration' screen for the seed job.
 
-def git_repo = 'https://github.com/puppetlabs/gatling-puppet-load-test.git'
-def git_branch = 'master'
+def git_repo = 'https://github.com/magisus/gatling-puppet-load-test.git'
+def git_branch = 'empty-repo-5x'
```
justinstoller: Is this still needed? With the latest changes do we expect it to succeed (IIRC, there's maybe still an issue with a hiera file?)

justinstoller:
point of order, should we merge https://github.com/puppetlabs/puppetlabs-puppetserver_perf_driver/pull/19 tag the repo, merge https://github.com/puppetlabs/puppetlabs-puppetserver_perf_driver_dev_control/pull/3  then merge this?

jenkins-integration/jenkins-jobs/job_bootstrap.groovy
Magisus: I didn't mean to commit this. This was only needed for dev testing.

jenkins-integration/jenkins-jobs/common/scripts/job-steps/initialize_ruby_env.sh
Magisus: I don't think the version of Ruby matters particularly. We could update the driver module to set the version it installs as `default` and then switch to that here, to avoid the coupling, but I do know in all of our Jenkins infrastructure in general, we tend to be specific about which version we are using right in the script. Not sure if that's good or bad.

justinstoller: I think it's a good thing in CI because there's (was?) multiple rubies installed to support both teams that needed to run spec tests against multiple rubies as well as projects that needed to run different versions of beaker that used different versions of ruby (eg. puppetserver 1.x and 2.x branches used different major versions of beaker, one requiring 1.9 and the other able to use ruby 2.x).

Unless we're going to be supporting multiple rubies I'd prefer not having to make the change in two places when we want to bump the ruby version.

Magisus: So the RVM default on the real perf driver is 1.9.3. Not sure if we can change that. So we may still have to set the version explicitly here (to something other than 2.4.2, since that version is not installed), unless we change that about the real perf driver.

Magisus:
This one can go in whenever, actually. It shouldn't be affected by the changes to the other repos.

PR #146, opened 2018-01-10
(SERVER-2065) Week-long simulation uses server 5.x

lindboe:
Update the week-long job to use server 5.x as that's what we're
interested in ensuring is stable with JRuby 9k at this point.


jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-1-week/Jenkinsfile
```diff
@@ -5,7 +5,7 @@ node {
 
 pipeline.single_pipeline([
         job_name: 'oss-latest-1-week',
-        gatling_simulation_config: '../simulation-runner/config/scenarios/foss25x-medium-1250-1-week.json',
+        gatling_simulation_config: '../simulation-runner/config/scenarios/foss5x-medium-1250-1-week.json',
```
haus: I think you need to add `jruby_jar: "/opt/puppetlabs/server/apps/puppetserver/jruby-9k.jar",` to get jruby9k in here.  not sure if we want a 1.7 and 9k version for comparison?

lindboe: I looked at the 9k configuration for comparison, saw that file also specifies agent_version: https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-jruby9k/Jenkinsfile#L13. I assume we want that? No idea what the behavior is if it's not specified.

Would 1.7 data be useful?

haus: agent_version sounds good to me. i think having 1.7 data to compare against will be useful yes.

haus: i guess we probably want jit on also then for the 9k work?

haus: we probably eventually want a job with max-requests-per-instance set to 100k like in PE, but until we've picked up the jruby-utils work to offset jruby expiration we'll get a lot of KOs in those runs which would make the data less useful.

lindboe: Cool cool, so i'll update this to have separate jobs for each version of jruby, and we'll see how it does without max-requests-per-instance.

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-1-week-9k/Jenkinsfile
```diff
@@ -0,0 +1,37 @@
...
+        ],
+        archive_sut_files: [
+                "/var/log/puppetlabs/puppetserver/metrics.json"
+        ],
+        jruby_jar: "/opt/puppetlabs/server/apps/puppetserver/jruby-9k.jar",
+        hocon_settings: [
+          [
+            file: "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf",
+            path: "jruby-puppet.compile-mode",
+            value: "jit"
```
lindboe: I'm not totally certain this was the correct thing to do, just copied it from the jit-off-on test

justinstoller: seems legit.

([docs](https://github.com/puppetlabs/puppetserver/blob/5.1.x/documentation/config_file_puppetserver.markdown#settings), [settings init](https://github.com/puppetlabs/jruby-utils/blob/master/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj#L164) and [instance creation](https://github.com/puppetlabs/jruby-utils/blob/master/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj#L81))

jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-latest-1-week/Jenkinsfile
haus: https://www.youtube.com/watch?v=fC_q9KPczAg

PR #148, opened 2018-01-10
(maint) Disable pe-puppetserver-latest cron

haus:
Having pe-puppetserver tests run automatically may someday be
desireable, but right now this just gets in the way by nuking SUTs that
are running tests and by causing conflicts with FOSS puppetserver. Also
this job is installing flanders. This commit disables the cron for now.


lindboe:
Okay normally I don't want to comment things out but that would leave this empty/remove the file which loses the history, so that sucks. Good for now.

PR #155, opened 2018-01-22
(QA-3434) Combine metrics and agent into 1 node

samwoods1:


jenkins-integration/beaker/install/pe/98_sync_codedir.rb
```diff
@@ -1,27 +0,0 @@
-test_name 'Sync code-staging to code via File Sync'
```
haus: This file is still used by PE gatling tests elsewhere in the repo. Please don't remove it.

PR #157, opened 2018-01-23
(SERVER-2078) Add HEAP_OVERRIDES, JAVA_ARGS_ADDITIONS param to jobs

haus:
Previously changing gatling job parameters such as heap settings or
which garbage collector to use was impossible. This commit adds two new
parameters to all of the gatling jenkins jobs that will allow us to
override heap settings and to supply additional java args to jobs such
as garbage collector.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -280,7 +280,7 @@ def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
             server_heap_settings = ""
         }
 
-        if ("${HEAP_OVERRIDES}" != "" && "${HEAP_OVERRIDES}" != null) {
+        if ("${HEAP_OVERRIDES}" != "") {
             server_heap_settings = "${HEAP_OVERRIDES}"
         }
 
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -272,12 +272,22 @@ def step075_customize_puppet_settings(script_dir, puppet_settings) {
     }
 }
 
-def step080_customize_java_args(script_dir, server_java_args, server_era) {
-    if ((server_java_args == null) || (server_java_args == "")) {
+def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
+    if ((server_heap_settings == null || server_heap_settings == "") && "${JAVA_ARGS_ADDITIONS}" == "" && "${HEAP_OVERRIDES}" == "") {
         echo "Skipping java_args configuration because none specified in job"
     } else {
+        if (server_heap_settings == null) {
+            server_heap_settings = ""
+        }
+
+        if ("${HEAP_OVERRIDES}" != "" && "${HEAP_OVERRIDES}" != null) {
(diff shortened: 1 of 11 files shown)


jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -272,12 +272,22 @@ def step075_customize_puppet_settings(script_dir, puppet_settings) {
...
-    if ((server_java_args == null) || (server_java_args == "")) {
+def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
+    if ((server_heap_settings == null || server_heap_settings == "") && "${JAVA_ARGS_ADDITIONS}" == "" && "${HEAP_OVERRIDES}" == "") {
         echo "Skipping java_args configuration because none specified in job"
     } else {
+        if (server_heap_settings == null) {
+            server_heap_settings = ""
+        }
+
+        if ("${HEAP_OVERRIDES}" != "" && "${HEAP_OVERRIDES}" != null) {
```
justinstoller: We check that the `HEAP_OVERRIDES` here are not null but above we only check to see if they are an empty string. Can we always make the assumption that we make in the first condition (params will always be a string) and remove this null check, or should we also add null checks above?

justinstoller: Mmmm... Does the interpolation (`"${HEAP_OVERRIDES}"`) mean that if the variable is null it will be a blank string?

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -272,12 +272,22 @@ def step075_customize_puppet_settings(script_dir, puppet_settings) {
     }
 }
 
-def step080_customize_java_args(script_dir, server_java_args, server_era) {
-    if ((server_java_args == null) || (server_java_args == "")) {
+def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
+    if ((server_heap_settings == null || server_heap_settings == "") && "${JAVA_ARGS_ADDITIONS}" == "" && "${HEAP_OVERRIDES}" == "") {
```
justinstoller: mimicking the params following, is this the same as `"${server_heap_settings}" == ""`

haus: I have no idea. One is a jenkins param and one is a groovy var.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
haus: Oh the null check was me trying to figure out what was wrong. But... null interpolated will not be the empty string. it will be "null".

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
haus: oh wait. yea no, not at all. because "${server_heap_settings}" would be "null" if it is null.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
haus: so i can pull out the heap_overrides null check because it won't do the right thing at all.

PR #159, opened 2018-01-24
(maint) Move max requests per instance to string

haus:
Seems like hocon expects the value to be a string and not an int. I'm
not sure if this is a limitation in hocon or in beaker's hocon helper.


justinstoller:
Looks like this ran at http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/oss-puppetserver-latest-jruby9k-max-request-per-instance/2/

haus:
:ohyea:

PR #161, opened 2018-01-25
(SERVER-2089) Add scenarios and simulation configs for 12/24hr jobs

haus:
Previously we only had 2 hour and 1 week runs, but having something in
the middle like 12 hr and 24 hr will be useful for chasing down some
timeouts and GC tuning. This commit adds 12 and 24 hour scenario config
files and jenkins jobs for both jruby 9k and 1.7.


lindboe:
Do we need both 12 and 24 hr?

haus:
I don't know if we totally need them, seemed like after we got 12hr good 24 might be a good next step before going to a week.

lindboe:
If the issue consistently replicates at 11 hrs, 12 is enough; but also, I guess it depends when you're kicking them off. If you're kicking it off in the morning, may as well run 24 hrs because you won't be here 12 hours later.

haus:
Exactly. I would like to kick off 12 hour runs now for tomorrow, but might want to kick off a 24 hour one tomorrow morning. So I think both are probably handy for now?

lindboe:
Cool, +1

PR #164, opened 2018-01-25
(SERVER-2063) Collect the simulation.log.gz from runs

haus:
This commit begins archiving the simulation.log.gz from each run. In
order to do this effectively we also clean up the simulation logs from
the workspace after collection.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -367,6 +367,12 @@ def step900_collect_driver_artifacts() {
     // NOTE: this DSL step requires the puppet-gatling-jenkins plugin.  It also
     // depends on some data that gets created via 025_collect_facter_data.sh
     puppetGatlingArchive()
+
+    // Always archive the simulation.log.gz from the run.
+    archive "simulation-runner/results/**/*.log.gz"
+    dir('simulation-runner/results') {
+        deleteDir()
+    }
 }
 
 SCRIPT_DIR = "./jenkins-integration/jenkins-jobs/common/scripts/job-steps"
--- a/jenkins-integration/jenkins-jobs/scenarios/latest-test-changes/Jenkinsfile
+++ b/jenkins-integration/jenkins-jobs/scenarios/latest-test-changes/Jenkinsfile
@@ -1,0 +1,26 @@
+node {
+    checkout scm
+    pipeline = load 'jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy'
+}
+
+pipeline.single_pipeline([
+        job_name: 'oss-latest',
+        gatling_simulation_config: '../simulation-runner/config/scenarios/foss5x-medium-1-node-1-iteration-test.json',
+        server_version: [
+                type: "oss",
+                version: "latest"
+        ],
+        code_deploy: [
+                type: "r10k",
+                control_repo: "git@github.com:puppetlabs/puppetlabs-puppetserver_perf_control.git",
+                basedir: "/etc/puppetlabs/code/environments",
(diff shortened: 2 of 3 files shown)


haus:
I have a test run of this going, I'll report back tomorrow.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -367,6 +367,12 @@ def step900_collect_driver_artifacts() {
     // NOTE: this DSL step requires the puppet-gatling-jenkins plugin.  It also
     // depends on some data that gets created via 025_collect_facter_data.sh
     puppetGatlingArchive()
+
+    // Always archive the simulation.log.gz from the run.
+    archive "simulation-runner/results/**/*.log.gz"
+    dir "simulation-runner/results" {
+        deleteDir
```
lindboe: Is this really necessary if the box will just be razored later?

haus: this is on the jenkins machine.

lindboe: oh, then I have no idea what archive does.

Does this still allow the gatling reports to be displayed in jenkins?

haus: i stuck it in `step900_collect_driver_artifacts` as that seemed like the right place for this.

lindboe: also related question: if puppetGatlingArchive() doesn't do this, what does it do

haus: I believe the reports are served from the `jobs/$job/builds/$build_number` directory, and this operates in the `jobs/$job/workspace` area. I'll know for sure after the current run completes.

haus: puppetGatlingArchive seems to be the magic to inline the html report. the important bits seem to be here: https://github.com/puppetlabs/puppet-gatling-jenkins-plugin/blob/be8328bc9d39c638cf0a123beb8a3ec5bfe98c2e/src/main/java/com/puppetlabs/jenkins/plugins/puppetgatling/PuppetGatlingPublisher.java

lindboe: Ah, okay. If you don't mind I'd like to wait for that to be sure, don't want to delete this and have it break for a first time run of anything

haus: Waiting is fine. I'm pretty confident based on poking around in the jenkins FS.

haus:
You can see this in action here: http://sclb4agkxecbfvc.delivery.puppetlabs.net:8080/job/latest-test-changes/

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -367,6 +367,12 @@ def step900_collect_driver_artifacts() {
     // NOTE: this DSL step requires the puppet-gatling-jenkins plugin.  It also
     // depends on some data that gets created via 025_collect_facter_data.sh
     puppetGatlingArchive()
+
+    // Always archive the simulation.log.gz from the run.
+    archive "simulation-runner/results/**/*.log.gz"
+    dir('simulation-runner/results') {
```
lindboe: SINGLE QUOTES :D

haus: NOTHING MAKES ANY SENSE

PR #165, opened 2018-01-25
(maint) Add readme and re-add force file sync

samwoods1:


NEW_README.md
```diff
@@ -0,0 +1,67 @@
...
+    * When asked for Puppet classes, enter the configuration which defaults to: role::by_size::large
+        * You can use a different puppet class by specifying it as the `PUPPET_SCALE_CLASS` environment variable for the performance_gatling rake task.
+### Getting a pre-recorded run started
+For the following steps, GatlingClassName is the value entered into the ClassName field during the recording step.
+From root of the gatling-puppet-load-test source dir on the Gatling driver (metrics):
+* First time only:
+    * edit /usr/share/sbt/conf/sbtopts and change “-mem” to 2048 (and uncomment)
+* For each new simulation:
+    * `cd ~/gatling-puppet-load-test/simulation-runner`
+    * `export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net`
```
haus: rubygems.delivery.puppetlabs.net won't work at some point in the near future (should be artifactory).

samwoods1: True, this sprint we're actually working on automating the gatling record and playback steps so I will make sure that we update it to use artifactory.  Thanks!

PR #166, opened 2018-01-25
(maint) Add rake support for installing different PE builds

samwoods1:


samwoods1:
@RandellP readme is updated

PR #171, opened 2018-02-01
 (PE-23417) Gatling scenarios for the latest Hoyt and Irving builds

doug-rosser:
We may not want to merge this branch/PR as-is, but I wanted everyone to get a look at the raw recordings and results.

HoytLargeRC10 - [tested against 2017.1.0-RC10](http://int-resources.ops.puppetlabs.net/drosser/PE-23417/results/rchoyt-metrics/results/HoytLargeRC10-1517445706208/)

HoytLargeGA - [tested against 2017.1.0](http://int-resources.ops.puppetlabs.net/drosser/PE-23417/results/gahoyt-metrics/results/HoytLargeGA-1517445734009/)

HoytLargeZ3 - [tested against a dev build of 2017.3.3](http://int-resources.ops.puppetlabs.net/drosser/PE-23417/results/lzhoyt-metrics/results/HoytLargeZ3-1517445754391/)

IrvingLargeRC3 - [tested against 2018.1.0-RC3](http://int-resources.ops.puppetlabs.net/drosser/PE-23417/results/rc3irvg-metrics/results/IrvingLargeRC3-1517445779808/)

IrvingLargeRC4 - [tested against 2018.1.0-RC4](http://int-resources.ops.puppetlabs.net/drosser/PE-23417/results/rc4irvg-metrics/results/IrvingLargeRC4-1517445798202/)

NOTE - You will notice that some of the recordings have an extra http call like:

```scala
                .exec(http("")
                        .get("/puppet/v3/file_metadatas/locales?environment=production&links=follow&recurse=true&source_permissions=ignore&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&ignore=%2A.pot&ignore=config.yaml&checksum_type=md5")
                        .headers(headers_0))
```

I don't know if this behavior is constant, intermittent, etc. but it's annoying and needs to be investigated...


haus:
@doug-rosser are these all running with static catalogs enabled and agents already having the catalog applied? I notice there is only one file metadata request for all of these.

doug-rosser:
From a comment by Josh Cooper, the aforementioned "locale" call was deliberately added to newer versions of puppet-agent. I will test with and without that call when I have a prospective fix.

doug-rosser:
@haus Yes, static catalogs and the agent has been run previously to establish a steady-state.

doug-rosser:
Looks like I'm missing node configs. Update incoming...

doug-rosser:
@haus I think these are ready to merge and I have :+1: from my team. Do you agree or is there another approach we should be using?

haus:
@doug-rosser seem good to me. I'm not sure of better approaches to be honest. Might be worth cleaning up old RC builds at some point but I'm not sure when that would be. So i'm good to merge these.

doug-rosser:
Ack! I don't know what the heck happened but rebasing the branch made this PR a nightmare. Closing and will re-open.

PR #172, opened 2018-02-02
(QA-3371) Automate gatling execution in Beaker for ApplesToApples

samwoods1:
Removed classification of the metrics/agent node since we are no longer recording an agent run.
Added static gatling config files for PerfTestLarge (using role::by_size::large) nodes and ApplesToApples scenario (750 agents, 8 iterations)
Created Beaker test to actually kick off the gatling run.


rakefile
```diff
@@ -64,6 +67,16 @@ rototiller_task :performance do |t|
         end
       end
     end
+    unless ENV['BEAKER_TESTS'] == ''
```
er0ck: does this not need a check for nil?

setup/install_gatling/30_classification/50_classify_nodes_via_site_pp.rb
```diff
@@ -2,7 +2,7 @@
 skip_test 'Installing PE, not FOSS' unless ENV['BEAKER_INSTALL_TYPE'] == 'foss'
 
 def generate_sitepp()
-  "node /.*agent.*/ { include #{ENV['PUPPET_SCALE_CLASS']} }\nnode /#{agent.hostname}/ { include #{ENV['PUPPET_SCALE_CLASS']} }\nnode 'default' {}"
+  "node /.*agent.*/ { include #{ENV['PUPPET_SCALE_CLASS']} }\nnode 'default' {}"
```
er0ck: i realize this was already like this.  could this be put in a heredoc or something more readable which would reproduce the newlines more effectively?

simulation-runner/config/scenarios/ApplesToApples.json
```diff
@@ -0,0 +1,12 @@
+{
+    "run_description": "'role::by_size_large' role from perf control repo, 750 agents, 8 iterations",
```
er0ck: should the file be named with 750/8 somehow?  or do we think we might not have other scenarios saved?  YAGNI, etc

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
```diff
@@ -0,0 +1,72 @@
...
+		.exec(http("filemeta pluginfacts")
+			.get("/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&source_permissions=use&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&ignore=%2A.pot&checksum_type=md5")
+			.headers(headers_0))
+		.pause(1)
+		.exec(http("filemeta plugins")
+			.get("/puppet/v3/file_metadatas/plugins?environment=production&links=follow&recurse=true&source_permissions=ignore&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&ignore=%2A.pot&checksum_type=md5")
+			.headers(headers_0))
+		.pause(1)
+		.exec(http("")
+			.get("/puppet/v3/file_metadatas/locales?environment=production&links=follow&recurse=true&source_permissions=ignore&ignore=.svn&ignore=CVS&ignore=.git&ignore=.hg&ignore=%2A.pot&ignore=config.yaml&checksum_type=md5")
```
er0ck: so... this will only work in Irving+, IIRC
i'm probably okay with that.

er0ck: oh.  the `X-Puppet-Version` header will probably limit how far we can take this "agent" as well, yeah?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
```diff
@@ -0,0 +1,72 @@
...
+
+class PerfTestLarge extends SimulationWithScenario {
+
+	val reportBody = ElFileBody("PerfTestLarge_0006_request.txt")
+
+	val baseHeaders = Map("Accept" -> "application/json, text/pson",
+		"Accept-Encoding" -> "gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
+		"User-Agent" -> "Puppet/5.3.3 Ruby/2.4.2-p198 (x86_64-linux)")
+
+	val headers_0 = baseHeaders ++ Map("X-Puppet-Version" -> "5.3.3")
```
er0ck: looks like this could be merged into baseHeaders.  but i realize this was copied in here, as-is.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
```diff
@@ -0,0 +1,72 @@
...
+
+	val headers_6 = baseHeaders ++ Map(
```
er0ck: at somepoint we should probably have better names for these.  looks like this one is only used for the report put

rakefile
samwoods1: This is basically just to make this optional, if it is nil, then it will get the default value.  If you don't want to run any tests, you have to explicitly set the BEAKER_TESTS env var to an empty string.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
samwoods1: Yeah, I mentioned in the spike ticket that there was going to be some tech-debt left and definitely some room to improve, but for the simplest fix I went with leaving everything as is.

simulation-runner/config/scenarios/ApplesToApples.json
samwoods1: My thought was that scenarios should be something generic that can be run over and over again, like whatever configuration we use for apples to apples.  Down the road, we may end up changing what we want to use for apples to apples, and I'd rather not have to deal with also changing the file name?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
samwoods1: Yes and yes. We will probably need to modify this file slightly per version/release. We still need to figure out what the best way to catch changes to the agent runs should be so that we can ensure the requests stay up to date.  Both sound like good future tickets.

rakefile
er0ck: true

RandellP: This is good.  I was sad when I first found out that beaker didn't have a -notests arg...  and that by default even setting it to '' doesn't stop it from running the default tests.

er0ck: you can just leave --tests off.  often i `touch empty.rb` and use that as my "test", ie when a rake task implies --test

RandellP: yeah, I have it a bit confused in my head.  It was that without the if he has here, you can't turnt he tests off in a simple intuitive way, like setting the var to ''.  Seems to me if I pass --tests '' to beaker, it shouldn't run tests.

setup/install_gatling/30_classification/40_classify_nodes_via_nc.rb
```diff
@@ -8,8 +8,7 @@ def classify_nodes(classifier)
   classifier.find_or_create_node_group_model(
       'parent' => '00000000-0000-4000-8000-000000000000',
       'name' => 'perf-agent-group',
-      'rule' => ['or', ['~', ['fact', 'clientcert'], '.*agent.*'],
-                 ['~', ['fact', 'clientcert'], "#{agent.hostname}"]],
+      'rule' => ['~', ['fact', 'clientcert'], '.*agent.*'],
```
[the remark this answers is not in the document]
er0ck: this uses the NC api to classify the nodes via clientcert fact as in:
https://github.com/puppetlabs/scooter/blob/master/lib/scooter/httpdispatchers/consoledispatcher.rb

simulation-runner/config/scenarios/ApplesToApples.json
er0ck: yeah, but i'm also fine if we ticket to fix this up later.  we have other tickets to tackle in this sprint 🤷‍♂️

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
er0ck: i don't think we know the abstraction yet.  i'd hope we'd program it in as parameters to a function in the gatling stuff, or some similar abstraction

samwoods1: No, what it means is that this file will have some variables in it that need to be replaced depending on which version we are running.  This works for current version, but some work will need to be done to make it work for all versions going forward.  We may be able to use some of Gatling's tooling or scala code to include or exclude different requests based on the build we are testing, or we may need to do something dirty like parse and modify this file...

setup/install_gatling/30_classification/40_classify_nodes_via_nc.rb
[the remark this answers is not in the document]
samwoods1: The group we want them classified into is always the same, but the class/modules we want to apply to that class change.  At some point we will certainly want multiple groups of nodes with various 'roles', but that's all future work.

simulation-runner/config/scenarios/ApplesToApples.json
RandellP: that actually makes good sense.  Lets just be sure to get it ticketed.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
RandellP: sounds like a good candidate to ticket for later consideration.

setup/install_gatling/30_classification/40_classify_nodes_via_nc.rb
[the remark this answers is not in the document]
RandellP: right but it class name is taken from the env at the time the rule is created...   actually I guess in this case it works because we are doing the whole run as one entity.  When we split out setup and run we will have to put this part in the run section and be sure to clear old rules.  I was still thinking of how we had two steps before.

simulation-runner/config/scenarios/ApplesToApples.json
samwoods1: @RandellP Good point about the 600/900. I'll change it to 600. 
I still disagree about the filename. The way it is currently, its easy to change what we want an ApplesToApples run to actually do without much trouble, just update the file contents. If we match the filename with the number of instances and repetitions, then any time we want to update the scenario, we have to change the file contents, file name and references to the file, and it also makes it harder to figure out which scenario we use for the apples to apples run without looking into the code to figure out which scenario is called instead of just finding the scenario named ApplesToApples.

setup/install_gatling/30_classification/40_classify_nodes_via_nc.rb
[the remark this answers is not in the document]
samwoods1: @RandellP Once the group is set up with the above rule, then any agent matching that regex will use that role.  We use the '.*agent.*' regex so we get any agent with the word 'agent' in the hostname, so no need to change the rules run over run.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
samwoods1: Created QA-3445

simulation-runner/config/scenarios/ApplesToApples.json
RandellP: We can ticket the idea of changing the name be a discussion for later.  I understand your points, and they are good.  I have some long term thoughts that may end up not mattering, or might be a very good reason... :)  But we can hash that out when we groom the ticket and see where we end up.

setup/install_gatling/30_classification/40_classify_nodes_via_nc.rb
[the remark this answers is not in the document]
RandellP: It was the classname I was thinking about.  when we split up the rake tasks we will want to be sure we put this code under the rake task for running the recording and I think we will also want to ensure it matches the recording.  Or maybe it has to come from the recording.  Either way, we wouldn't want it in the general setup as we would use the same setup for running different recordings with different classes.  But for now, for what we have, this is fine.

er0ck:
did anyone else run this?  we aint got no tests, yet

samwoods1:
I amended the commit to also update the readme.

NEW_README.md
```diff
@@ -16,9 +18,15 @@ This will configure multiple nodes.  In the simplest case you need 2 nodes: a Pu
 * Execute:
     * export BEAKER_HOSTS=\<your hosts file>
```
RandellP: You don't need to do this anymore...  at least it is optional.   It should either be mentioned as optional, or moved up to the section about doing something custom.

PR #173, opened 2018-02-02
(SERVER-2108) Add dynamic gatling job

haus:
Currently in order to test new combinations of node count, duration, or
catalog size a new job and scenario must be constructed. This is slow
and tedious, so this commit adds a new jenkins job that constructs a
custom gatling scenario on the fly and uses it to kick off a job against
a node config for the desired catalog size.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -127,6 +127,28 @@ def get_agent_version(agent_version) {
   }
 }
 
+String generate_gatling_scenario(hours, size, count) {
+    repetitions = hours.toInteger() * 2
+    filename = "foss5x-${size}-${count}-${hours}-hours-dynamic.json"
+    node_configs = [
+            EMPTY: "FOSS5xEmptyRepo.json",
+            MEDIUM: "FOSS5xPerfMedium.json",
+    ]
+    scenario_hash = [
+            run_description: "${size} role from perf control repo, ${count} agents, ${hours} hours",
+            nodes: [[
+                            node_config: node_configs[size],
+                            num_instances: count.toInteger(),
+                            ramp_up_duration_seconds: 1800,
+                            num_repetitions: repetitions,
+                            sleep_duration_seconds: 1800,
+                    ]]
+    ]
+    scenario_json = JsonOutput.toJson(scenario_hash)
+    writeFile(file: filename, text: scenario_json)
+    return filename
+}
+
 def step000_provision_sut(SKIP_PROVISIONING, script_dir) {
     echo "SKIP PROVISIONING?: ${SKIP_PROVISIONING} (${SKIP_PROVISIONING.class})"
     if (!SKIP_PROVISIONING) {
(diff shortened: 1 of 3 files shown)


haus:
A version of this job has been deployed here: http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/puppetserver-infinite/

jenkins-integration/jenkins-jobs/scenarios/puppetserver-infinite/JobDSL.groovy
```diff
@@ -0,0 +1,5 @@
+job.parameters {
+    stringParam('NUMBER_OF_HOURS', '2', 'Length of time to run the gatling simulation.')
+    choiceParam('CATALOG_SIZE', ['MEDIUM', 'EMPTY'], 'Catalog to use in simulation.')
+    choiceParam('NODE_COUNT', ['600', '100', '200', '300', '900', '1200', '1500', '1800'], 'Number of nodes to include in simulation.')
```
lindboe: It's probably useful to include in the description that these numbers are selected such that gatling will distribute them evenly

haus: Yea. That's a good idea. I also need to PR updating the scenarios to even that out.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -127,6 +127,28 @@ def get_agent_version(agent_version) {
   }
 }
 
+String generate_gatling_scenario(hours, size, count) {
```
lindboe: Maybe a comment indicating you're going to need to make sure you're in the correct dir or you won't be able to find the corresponding node_configs file because it's a relative path

jenkins-integration/jenkins-jobs/scenarios/puppetserver-infinite/JobDSL.groovy
Magisus: Does "choiceParam" show up as a drop-down or something, or do you have to know what your options are?

lindboe: It does! If you go to http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/puppetserver-infinite/ and look at build with params you should see it

Magisus: That's cool, never seen that before. Is it possible in JJB, I wonder, and we just don't use it?

lindboe: Not sure. It wasn't listed as an option in the JobDSL docs, but it was in an example and the source code, so it's weird all around.

Magisus:
Found the comment with the deployment, derp. Looks good!

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
haus: Good call.

lindboe:
Note: this got deployed to the real driver because we were having such problems with the script security plugin on the dev jenkins instance that we literally couldn't deploy this. This might be better on later jenkins where we can use the permissive-script-security plugin, so we should prioritize trying to update jenkins.

jenkins-integration/jenkins-jobs/scenarios/puppetserver-infinite/JobDSL.groovy
```diff
@@ -0,0 +1,6 @@
+job.parameters {
+    stringParam('NUMBER_OF_HOURS', '2', 'Length of time to run the gatling simulation.')
+    choiceParam('CATALOG_SIZE', ['MEDIUM', 'EMPTY'], 'Catalog to use in simulation.')
+    // NODE_COUNT is offered as a list because gatling will not handle values that don't divide 1800 evenly (multiples of 300 also seem fine, which is why 1200 and 1500 are also there).
```
lindboe: I was thinking more in the description, as just a comment to the operator about why they can't pick arbitrary numbers. All we have to say is that Gatling won't distribute the load evenly for other options.

haus: ah got it

PR #174, opened 2018-02-05
(SERVER-2108) Fix scenario file path

haus:
Previously the custom scenario file was set to just the filename, and
not the full path to the scenario file, when being generated. This
unfortunately meant that none of the functions could find the file
later. This commit prepends the path to the filename to ensure it can be
found.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -424,7 +424,8 @@ def single_pipeline(job) {
         if (job_name == "puppetserver-infinite") {
 
             dir('simulation-runner/config/scenarios') {
-                job["gatling_simulation_config"] = generate_gatling_scenario(NUMBER_OF_HOURS, CATALOG_SIZE, NODE_COUNT)
+                scenario_name = generate_gatling_scenario(NUMBER_OF_HOURS, CATALOG_SIZE, NODE_COUNT)
+                job["gatling_simulation_config"] = "../simulation-runner/config/scenarios/${scenario_name}"
             }
 
             if (CATALOG_SIZE == "EMPTY") {


jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -424,7 +424,8 @@ def single_pipeline(job) {
         if (job_name == "puppetserver-infinite") {
 
             dir('simulation-runner/config/scenarios') {
-                job["gatling_simulation_config"] = generate_gatling_scenario(NUMBER_OF_HOURS, CATALOG_SIZE, NODE_COUNT)
+                scenario_name = generate_gatling_scenario(NUMBER_OF_HOURS, CATALOG_SIZE, NODE_COUNT)
+                job["gatling_simulation_config"] = "../simulation-runner/config/scenarios/${scenario_name}"
```
Magisus: The relative path is fine here? We won't ever need to reference this from some other location than we are doing currently?

haus: The job hash drives everything else as far as i know. I have a for real job based on this branch running here as proof: http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/puppetserver-infinite/18/console

justinstoller: The double dots are weird. I'm guessing that this assumes you're reading this variable from a specific directory? I think that should be the responsibility of the caller. ie no `"../"` on this variable.

haus: I don't disagree with the sentiment, but it is "../" in *all* of the Jenkinsfiles so I'm going for consistency.

haus: (for example https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-5.x/Jenkinsfile#L8)

Magisus: Eh, seems fine then I guess.

haus: Or I guess I should probably say, if we ever need to reference it from somewhere else we'll be making a bunch of other changes already, as it is hardcoded in all of the Jenkinsfiles and seems to be an assumption of everything.

justinstoller: Hrmm... that sucks. I guess we're at least consistent about always considering this data to be executed within the `jenkins-integration` sub directory of the repo: https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/jenkins-integration/jenkins-jobs/scenarios/oss-puppetserver-5.x/Jenkinsfile#L21

PR #177, opened 2018-02-07
(QA-3352) Use beaker-benchmark to manage atop

samwoods1:
We needed to log atop results in a CSV format to make it so we can do some data analysis, and use beaker-benchmark to get averages and summarize the results.
Also updated the code to store results to include important gatling and atop logs that are stored.


tests/gatling_kickoff.rb
```diff
@@ -1,6 +1,26 @@
...
+          "PUPPET_GATLING_SIMULATION_ID=PerfTestLarge sbt run",
+         {:accept_all_exit_codes => true}
+    end
+
+  ensure
+    perf_result = stop_monitoring(master, 'Apples to Apples gatling run', atop_log, '/opt/puppetlabs')
+    if perf_result
+      perf_result.log_summary
+      # Write summary results to log so it can be archived
+      perf_result.log_csv "PERF_#{today}/atop_summary.csv"
```
RandellP: maybe you have some requires and such elsewhere that make the "today" work, but standalone this won't run.  Also, if doing manual runs, and you do more than one a day, one will overwrite the other.

tests/zzz_copy_sut_archive_files.rb
```diff
@@ -1,16 +1,15 @@
...
-
-  FileUtils.mkdir_p(archive_dir)
-
-  archive_files.each do |s|
-    Beaker::Log.notify("Copying archive file '#{s}'")
-    scp_from(master, s, archive_dir)
-  end
+  today = Date.today().to_s
+  # truncate the job name so it only has the name-y part and no parameters
+  job_name = (ENV['JOB_NAME'] || 'unknown_jenkins_job')
```
RandellP: This seems to assume that you are in jenkins.  What about manual runs like when we are debugging a performance issue by tweaking the setup and rerunning the simulation?

tests/zzz_copy_sut_archive_files.rb
```diff
@@ -1,16 +1,15 @@
...
+                 .sub(/[A-Z0-9_]+=.*$/, '')
+                 .gsub(/[\/,.]/, '_')[0..200]
+  archive_name = "#{job_name}__#{ENV['BUILD_ID']}__#{today}__perf-files.tgz"
```
RandellP: If outside of jenkins this name will be the same all day long.

tests/zzz_copy_sut_archive_files.rb
```diff
@@ -1,16 +1,15 @@
...
-
-  archive_dir = "../puppet-gatling/#{job_name}/sut_archive_files"
-
-  FileUtils.mkdir_p(archive_dir)
-
-  archive_files.each do |s|
-    Beaker::Log.notify("Copying archive file '#{s}'")
-    scp_from(master, s, archive_dir)
-  end
+  today = Date.today().to_s
```
RandellP: I don't see date being required...  (anywhere in gatling repo at least)

tests/zzz_copy_sut_archive_files.rb
samwoods1: It just uses that to name the file. If there is no job name set, it defaults to 'unknown_jenkins_job' which is what you get when you run it outside of jenkins.

tests/zzz_copy_sut_archive_files.rb
samwoods1: Yes.  This is primarily for use in CI, although it could be handy when running locally just because it will copy all of the results you care about into one place.  This is also completely stolen from pe_acceptance_tests :-)

tests/zzz_copy_sut_archive_files.rb
samwoods1: Date is part of core ruby, doesn't need to be required.

tests/zzz_copy_sut_archive_files.rb
samwoods1: Also, you still get per node directories as well, so your results per run will be preserved unless you do it on the same set of nodes.

samwoods1:
@RandellP There was some reason that I am not remembering that we didn't run zzz_copy_sut_archive_files.rb as a post-suite in pe_acceptance_tests...  I can't remember what it was?

tests/zzz_copy_sut_archive_files.rb
samwoods1: Ok, so I was wrong: https://stackoverflow.com/questions/11631975/why-does-date-exist-in-ruby-before-it-is-required/11633934#11633934

samwoods1: But, this does work, it must be required by one of the modules mixed in by Beaker.

RandellP: It is best practice to require it anyway... to show the dependency?  Ruby should be smart enough to do nothing if it is already required.   And that would make the code resilient in the case that for some reason whatever did require it stops doing so... like the beaker modularization effort that trys to load less things unless needed.

tests/zzz_copy_sut_archive_files.rb
RandellP: right but the name implies it was still a jenkins job...   And all the reg exps don't need to be done on that known string either.

tests/zzz_copy_sut_archive_files.rb
RandellP: I was thinking same set of nodes... like multiple dev runs trying out different tweaks, then wanting to compare the results of each to look for performance changes.

tests/gatling_kickoff.rb
```diff
@@ -1,6 +1,26 @@
...
+    step "Execute gatling scenario" do
+      on metric, "cd /root/gatling-puppet-load-test/simulation-runner/ && " +
+          "PUPPET_GATLING_MASTER_BASE_URL=https://#{master.hostname}:8140 " +
+          "PUPPET_GATLING_SIMULATION_CONFIG=config/scenarios/ApplesToApples.json " +
+          "PUPPET_GATLING_SIMULATION_ID=PerfTestLarge sbt run",
+         {:accept_all_exit_codes => true}
+    end
+
+  ensure
+    perf_result = stop_monitoring(master, 'Apples to Apples gatling run', atop_log, '/opt/puppetlabs')
```
RandellP: internally stop monitoring looks odd to me...
I see it do this
scp_from(infrastructure_host, atop_log_name, '/tmp')
followed by this
CSV.parse(File.read("/tmp/#{File.basename(atop_log_name)}"), { :col_sep => ' ' }) do |row|

The first looks like it would create /tmp/tmp/atop.log on the testrunner.
The second looks like it would open /tmp/atop.log to parse.
That seems like it should fail.

tests/gatling_kickoff.rb
```diff
@@ -1,6 +1,26 @@
-step "Execute gatling scenario" do
-  on metric, "cd /root/gatling-puppet-load-test/simulation-runner/ && " +
-      "PUPPET_GATLING_MASTER_BASE_URL=https://#{master.hostname}:8140 " +
-      "PUPPET_GATLING_SIMULATION_CONFIG=config/scenarios/ApplesToApples.json " +
-      "PUPPET_GATLING_SIMULATION_ID=PerfTestLarge sbt run"
+require 'beaker-benchmark'
+
+test_name 'Kickoff Gatling' do
+
+  atop_log = '/tmp/atop.log'
```
RandellP: This name being generic could cause collisions betweens runs launched from teh same testrunner... as beaker-benchmark copies to the test runner.

post/zzz_copy_sut_archive_files.rb
```diff
@@ -0,0 +1,21 @@
...
+    job_name = 'unknown_or_dev_job'
+  end
+
+  archive_name = "#{job_name}__#{ENV['BUILD_ID']}__#{now}__perf-files.tgz"
+  archive_root = "PERF_#{now}"
+
+  # Archive the gatling result htmls from the metrics box and the atop results from the master
+  archive_file_from(metric, '/root/gatling-puppet-load-test/simulation-runner/results', {}, archive_root, archive_name)
+  archive_file_from(master, '/tmp/atop.log', {}, archive_root, archive_name)
+end
```
RandellP: I don't see the atop_cvs_summary getting archived from the test_runner.

tests/gatling_kickoff.rb
```diff
@@ -1,6 +1,26 @@
...
+          "PUPPET_GATLING_SIMULATION_ID=PerfTestLarge sbt run",
+         {:accept_all_exit_codes => true}
+    end
+
+  ensure
+    perf_result = stop_monitoring(master, 'Apples to Apples gatling run', atop_log, '/opt/puppetlabs')
+    if perf_result
+      perf_result.log_summary
+      # Write summary results to log so it can be archived
+      perf_result.log_csv "PERF_#{Date.today}/atop_summary.csv"
```
RandellP: I don't see this file getting archived...

post/zzz_copy_sut_archive_files.rb
```diff
@@ -0,0 +1,22 @@
...
+    job_name = 'unknown_or_dev_job'
+  end
+
+  archive_name = "#{job_name}__#{ENV['BUILD_ID']}__#{now}__perf-files.tgz"
+  archive_root = "PERF_#{now}"
+
+  # Archive the gatling result htmls from the metrics box and the atop results from the master
+  Dir.mkdir_p "#{archive_root}/#{master.hostname}"
+  FileUtils.cp_r "tmp/atop/#{@@session_timestamp}/#{master.hostname}", "#{archive_root}/#{master.hostname}"
+  archive_file_from(metric, '/root/gatling-puppet-load-test/simulation-runner/results', {}, archive_root, archive_name)
```
er0ck: should this be conditional? or will we always have gatling results?

rakefile
```diff
@@ -4,6 +4,9 @@ task :default => :performance
 
 desc 'Run Performance setup'
 rototiller_task :performance do |t|
+  raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
+      ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE' == 'foss']
+
   t.add_env({:name => 'BEAKER_PE_DIR', :message => 'The PE download directory, example: http://enterprise.delivery.puppetlabs.net/archives/releases/2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
```
er0ck: i just noticed we don't add_env BEAKER_INSTALL_TYPE to rototiller's env vars.  is that intended?

setup/install_gatling/00_pre_install/20_rpm_setup.rb
```diff
@@ -33,11 +33,6 @@
     on hosts, 'yum install -y lsof || true'
   end
 
-  step 'enable atop service' do
-    on hosts, 'chkconfig atop on'
-    on hosts, 'service atop start'
-  end
```
er0ck: this isn't required now?

simulation-runner/config/scenarios/ApplesToApples.json
```diff
@@ -1,5 +1,5 @@
 {
-    "run_description": "'role::by_size_large' role from perf control repo, 750 agents, 8 iterations",
+    "run_description": "'role::by_size_large' role from perf control repo, 600 agents, 8 iterations",
```
er0ck: why was this changed in this PR?

rakefile
samwoods1: Its down on line 104, it's a cmd line option.

rakefile
```diff
@@ -4,6 +4,9 @@ task :default => :performance
 
 desc 'Run Performance setup'
 rototiller_task :performance do |t|
+  raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
+      ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE' == 'foss']
```
er0ck: misplaced `]`

er0ck: if this was in puppetlabs fork, i could just fix it O:)

simulation-runner/config/scenarios/ApplesToApples.json
samwoods1: Probably should have been a different PR, just noticed this was wrong and fixed it.

rakefile
er0ck: oops.  👍

post/zzz_copy_sut_archive_files.rb
samwoods1: To support the case where we just want to set things up to record, this should be conditional. It sort of is since you can set BEAKER_POST_SUITE, but probably good to do a check to make sure this file exists before trying to do anything.

setup/install_gatling/00_pre_install/20_rpm_setup.rb
samwoods1: No, it's being managed by beaker-benchmark now.

simulation-runner/config/scenarios/ApplesToApples.json
er0ck: in the future that might be better.  but in the meantime, especially with "magic numbers" this might benefit from a commit statement about why it was wrong and why it was changed to what it was changed... to

samwoods1: 👍

post/zzz_copy_sut_archive_files.rb
```diff
@@ -0,0 +1,25 @@
...
+  # truncate the job name so it only has the name-y part and no parameters
+  if ENV['JOB_NAME']
+    job_name = ENV['JOB_NAME']
+                   .sub(/[A-Z0-9_]+=.*$/, '')
+                   .gsub(/[\/,.]/, '_')[0..200]
+  else
+    job_name = 'unknown_or_dev_job'
+  end
+
+  archive_name = "#{job_name}__#{ENV['BUILD_ID']}__#{now}__perf-files.tgz"
```
RandellP: build id would only be there for a jenkins job right?   And since archive root has #{now} does the file name need it as well?  Does it need build or jobname either?  Maybe a meta data file instead with any useful information about the source of the data should be added to the archive, instead of encoding it in a file name that is already in a unique directory.  (though if the tgz is all we are going to upload then you would need at least the "now" to keep it unique.

post/zzz_copy_sut_archive_files.rb
```diff
@@ -0,0 +1,25 @@
...
+  archive_root = "PERF_#{now}"
+
+  # Archive the gatling result htmls from the metrics box and the atop results from the master (which are already copied locally)
+  if (Dir.exist?("tmp/atop/#{@@session_timestamp}/#{master.hostname}"))
+    Dir.mkdir_p "#{archive_root}/#{master.hostname}"
+    FileUtils.cp_r "tmp/atop/#{@@session_timestamp}/#{master.hostname}", "#{archive_root}/#{master.hostname}"
+    archive_file_from(metric, '/root/gatling-puppet-load-test/simulation-runner/results', {}, archive_root, archive_name)
+  end
```
RandellP: I don't think this will pick up the csv.  It looks like the csv goes to pwd/tmp...  and pwd should be /root/gatling-puppet-load-test/simulation-runner
Also, what happens to the summary.  looks like it just gets printed to the terminal output?

post/zzz_copy_sut_archive_files.rb
samwoods1: Correct, build id will only be there for a jenkins job.  The file name is the name of the archive file that gets created when it zips up the archive directory, it won't be under the folder itself, it will be in the working directory so it does also need to be unique. The reason we want the other info there is because we'll be downloading these from jenkins and they'll show up in our downloads directory, and this is a simple way to keep them organized without having to unzip the archive and open it to see what it was an archive of.

post/zzz_copy_sut_archive_files.rb
samwoods1: pwd is just /root/gatling-puppet-load-test, no simulation-runner.
And yes, the summary is just logged to stdout

post/zzz_copy_sut_archive_files.rb
RandellP: But if not in jenkins buildID will be undefined.

samwoods1: Correct, when it is run locally it will look like this:
unknown_or_dev_job____1518473738__perf-files.tgz
That's fine, since when we run it locally we aren't having to worry about having it organized by build, it will just be sorted by run time, so the most recent run will be at the bottom.

er0ck: i'm personally fine with this for the time being.  it woulda been cool to replace `nil` with `'local'` or something.  but we don't know how long this will be around for.  so i'm okay with it like this and we can fix it later if it's annoying/confusing.

PR #178, opened 2018-02-07
(maint) Minor gatling jenkins job improvements

haus:
This PR aims for better history and defaults on jobs. It updates the jobs to default to re-provisioning hosts, and also begins collecting more artifacts.

### DIFF
--- a/jenkins-integration/jenkins-jobs/scenarios/latest-test-changes/Jenkinsfile
+++ b/jenkins-integration/jenkins-jobs/scenarios/latest-test-changes/Jenkinsfile
@@ -22,6 +22,8 @@ pipeline.single_pipeline([
         ],
         archive_sut_files: [
                 "/var/log/puppetlabs/puppetserver/metrics.json",
-                "/var/log/puppetlabs/puppetserver/gc.log"
+                "/var/log/puppetlabs/puppetserver/gc.log",
+                "/etc/sysconfig/puppetserver",
+                "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf"
         ]
 ])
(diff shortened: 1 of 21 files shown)


haus:
fwiw it looks like razoring the node takes ~15 minutes right now.

jenkins-integration/jenkins-jobs/scenarios/latest-test-changes/Jenkinsfile
```diff
@@ -22,6 +22,8 @@ pipeline.single_pipeline([
         ],
         archive_sut_files: [
                 "/var/log/puppetlabs/puppetserver/metrics.json",
-                "/var/log/puppetlabs/puppetserver/gc.log"
+                "/var/log/puppetlabs/puppetserver/gc.log",
+                "/etc/sysconfig/puppetserver",
+                "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf"
```
lindboe: @Magisus with what you were doing to archive data to s3, would we have to make changes every time we add a new archive file or will new things get picked up automatically?

lindboe:
I don't know why I expected something other than mainly the same three lines repeated over and over again. Woo artifacts!

jenkins-integration/jenkins-jobs/scenarios/latest-test-changes/Jenkinsfile
Magisus: If archiving these from the job puts them all in one directory on the master, then we can just upload that whole directory.

haus: All of the files that are from the SUT end up in a directory called `sut_archive_files` (http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/puppetserver-infinite/26/artifact/puppet-gatling/puppetserver-infinite/sut_archive_files/gc.log).

The simulation logs come from the jenkins server and they end up in a directory called `simulation-runner` (http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/puppetserver-infinite/26/artifact/simulation-runner/results/puppetserver-infinite-1518224079732/simulation.log.gz). They both live under the artifact level of the jenkins job hierarchy, but that doesn't appear to map to an actual directory on disk.

PR #180, opened 2018-02-13
(SERVER-2108) Expose jruby version to puppetserver-infinite job

haus:
When running tests it is desireable to be able to choose between 9k and
1.7, so this commit adds such functionality to the puppetserver-infinite
job.
PR also includes updating the default java args to include full timestamps and tenuring details.


lindboe:
I think I would've expected an enum param that would've passed the correct string to the jruby_jar argument of `step081_customize_jruby_jar`, because now there are two different ways to manipulate the jar path which could cause some confusion

PR #181, opened 2018-02-13
(SERVER-2108) Update jar customization to handle no param case

haus:
Because the JRUBY_VERSION parameter was only added to the
puppetserver-infinite job and because we are on an older version of the
pipelines plugin, jobs that don't define the JRUBY_VERSION parameter
fail with a MissingPropertyException. In our current version of the
pipelines plugin, the recommended approach seems to be a try/catch
block. When we update to a more recent version this hackery can be
unwound.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -322,8 +322,18 @@ def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
 
 def step081_customize_jruby_jar(script_dir, jruby_jar, server_era) {
     def jar_string = ""
+    def JRUBY_VERSION_SET = true
 
-    if ("${JRUBY_VERSION}" == "") {
+    // This garbage can be replaced with either getBinding().hasVariable("JRUBY_VERSION")
+    // or via a params map (params.JRUBY_VERSION == null)
+    // https://wiki.jenkins.io/display/JENKINS/Pipeline+Groovy+Plugin#PipelineGroovyPlugin-2.18%28Sep23%2C2016%29
+    try {
+        println("JRUBY_VERSION is ${JRUBY_VERSION}")
+    } catch (MissingPropertyException e) {
+        JRUBY_VERSION_SET = false
+    }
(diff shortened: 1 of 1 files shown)


jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -322,8 +322,18 @@ def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
 
 def step081_customize_jruby_jar(script_dir, jruby_jar, server_era) {
     def jar_string = ""
+    def JRUBY_VERSION_SET = true
+
+    // This garbage can be replaced with either getBinding().hasVariable("JRUBY_VERSION")
+    // or via a params map (params.JRUBY_VERSION == null)
+    // https://wiki.jenkins.io/display/JENKINS/Pipeline+Groovy+Plugin#PipelineGroovyPlugin-2.18%28Sep23%2C2016%29
```
justinstoller: So this garbage can go away when https://tickets.puppetlabs.com/browse/SERVER-2092 is done (it seems like updating the plugins are implicit in the description of that ticket)?

haus: Yes, exactly. Based on https://wiki.jenkins.io/display/JENKINS/Pipeline+Groovy+Plugin#PipelineGroovyPlugin-2.18%28Sep23%2C2016%29 we just need 2.7 or newer of the plugin.

justinstoller: Would it be valuable to update this comment or comment in SERVER-2092 that we need to update this when we work that ticket?

haus: I'm happy to drop a note in SERVER-2092 about the sadness.

lindboe:
Hmm this is another reason it would be nice to have puppetserver-infinite in its own pipeline type. Is that worth doing to avoid this?

haus:
My inclination is still no, but that might change if/when we look at pulling in a specific puppet branch for accumulated changes.

PR #182, opened 2018-02-13
(SERVER-2017) More server simulations please

lindboe:
Create a new kind of pipeline that can be used to spin up a puppetserver
instance for some hands-on testing. This is basically a gatling run
without gatling or cleanup (so any desired artifacts will have to be
archived manually).

Mainly just want to see if this works since the dev workflow is a big hassle.
Still want to create a tool to simulate agent load without gatling, but
not sure that kicking it off immediately from this job is the right way
to go.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -557,4 +557,69 @@ def multipass_pipeline(jobs) {
     }
 }
 
+def spin_up_puppetserver(job) {
+    // This is just a way of setting up a system that's easy to try stuff out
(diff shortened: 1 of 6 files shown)


haus:
@lindboe did you set the metadata for the small suts manually or via policy files (https://github.com/puppetlabs/puppetlabs-modules/blob/production/site/profile/files/razor/puppetserver-perf/tasks/puppetserver-sut-centos/7.task/kickstart.erb#L26)?

lindboe:
Oh yes I forgot to add that in this, I only did the razor box. Thank you!

haus:
This looks pretty reasonable. The node name is still funky, but :woman_shrugging:, the mapping seems fine.

lindboe:
Updated

PR #185, opened 2018-02-16
(maint) Install correct sbt version to avoid 2 installs

samwoods1:


er0ck:
how does this avoid 2 installs?

what is the mkdir for?

samwoods1:
We install sbt explicitly here, but when we execute 'sbt run' to kick off Gatling, it requires a different version and installs that version as well.
The newer version of sbt included this file and directory already, but the version Gatling requires did not, so I had to create it.

samwoods1:
Included the above comment in the commit description.

doug-rosser:
:+1:

PR #190, opened 2018-02-23
(SERVER-2123) Create job to run gatling against puppet branch

haus:
In order to be able to better test how some changes to puppet affect
performance, this commit adds a job that can run gatling against a
puppet branch or ref. It clones and checks out the ref locally and
rsyncs it on top of the puppet-agent installation, which is gross, but
ultimately far simpler and faster than crafting a vanagon pipeline.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -212,6 +212,12 @@ def step020_install_server(SKIP_SERVER_INSTALL, script_dir, server_era, agent_ve
     }
 }
 
+def step021_clone_and_rsync(script_dir, remote, ref) {
+    withEnv(["PUPPET_REMOTE=${remote}", "PUPPET_REF=${ref}"]) {
+        sh "${script_dir}/021_clone_and_rsync.sh"
+    }
+}
+
 def step025_collect_facter_data(job_name, gatling_simulation_config, script_dir, server_era) {
     withEnv(["PUPPET_GATLING_SIMULATION_CONFIG=${gatling_simulation_config}",
              "PUPPET_GATLING_SIMULATION_ID=${job_name}",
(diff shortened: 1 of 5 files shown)


haus:
Please hold. Something is weird here.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -212,6 +212,14 @@ def step020_install_server(SKIP_SERVER_INSTALL, script_dir, server_era, agent_ve
     }
 }
 
+def step021_clone_and_rsync(script_dir, remote, ref) {
```
Magisus: Are there rules around this numbering scheme (here and the way we tend to do steps in the main CI Jenkins pipelines)? I've always been a little confused by it.

haus: No rules, it just mirrors the order in the pipeline itself. I figured right after the server install would be the best spot, but I don't care too much.

haus:
The issue is with puppet itself, and only puppet post 5.4.0 (https://tickets.puppetlabs.com/browse/PUP-8489)

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
Magisus: I was more curious about the intervals, when to increment 1 vs. 5 vs. 10. Always seems sorta arbitrary, which it very well might be :)

haus: Oh yea. Totally arbitrary.

haus: This *felt* closer to the install 🤷‍♀️

lindboe:
Why did we go with adding a new type of scenario/job instead of making this a parameter on the original puppetserver-infinite job?

haus:
mostly because parameters that might not be set are kind of a pain to deal with, but also last time the complexity of param vs jenkinsfile entry came up. I can certainly wedge this into the existing scenario if that seems better.

lindboe:
I'm not sure I understand -- why wouldn't the parameters be set? We are providing a default right?

haus:
maybe i misunderstood the question. i thought you were asking about the new pipeline definition. i think it would work fine merged into the other infinite job, as long as it starts using the new pipeline. i can squash it together.

lindboe:
Oh yeah I guess I wasn't really distinguishing between pipeline/job. I was just thinking that since puppetserver-infinite was supposed to be able to take all the params, that's where I expected this change to go.

haus:
@lindboe updated. Have a job running and will kick off a fresh one in a few minutes to validate the changes, but it looks good so far.

haus:
Hmm. Looks like known_hosts fun times with github now. Update coming.

haus:
Green run from this branch here: http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/job/puppetserver-infinite/48/console

PR #191, opened 2018-02-23
(SERVER-2080) Add step to base pipeline to publish to s3

Magisus:
This commit adds a step to the end of the common pipeline that uploads
the simulation results and all the files from the SUT to our s3 bucket.
All the files are consolidated in a folder with the job name and time
stamp, to ensure unique, identifiable entries in s3.


haus:
Nice!

PR #196, opened 2018-02-27
(maint) Updated perf_helper and spec steps

bcpuppet:
- Updated `perf_helper.rb` to remove the `atop` and `lsof` steps
- Updated `perf_helper_spec.rb` to remove the corresponding test steps


samwoods1:
image

PR #198, opened 2018-03-05
(PE-23417) Gatling scenarios for the latest Hoyt and Irving builds

doug-rosser:
We may not want to merge this branch/PR as-is, but I wanted everyone to get a look at the raw recordings and results.

HoytLargeRC10 - tested against 2017.1.0-RC10

HoytLargeGA - tested against 2017.1.0

HoytLargeZ3 - tested against a dev build of 2017.3.3

IrvingLargeRC3 - tested against 2018.1.0-RC3

IrvingLargeRC4 - tested against 2018.1.0-RC4


RandellP:
seems like a lot of file with the same content.  Anyway to avoid the copy and paste? links maybe?

PR #199, opened 2018-03-06
(SLV-15) Add new tests to perf_helper_spec

bcpuppet:
PLEASE DON'T MERGE

I'd like to get some feedback and resolve the TODO items before merging.

### DIFF
--- a/spec/setup/helpers/perf_helper_spec.rb
+++ b/spec/setup/helpers/perf_helper_spec.rb
@@ -6,6 +6,7 @@ class PerfHelperClass
   # These are all attributes available in Beaker::TestCase that are accessed by our helper
   attr_accessor :hosts, :options, :agents
 
+
   def logger
     @logger = Beaker::Logger.new({:log_level => 'error'})
   end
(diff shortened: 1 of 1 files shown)


spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -20,17 +21,50 @@ def options
...
+
+      # TODO: test both true and false (loadbalancer.ip / master.ip)?
+      expect(subject).to receive(:any_hosts_as?).with((:loadbalancer)).and_return(false)
+      expect(subject).to receive(:on).with(hosts, "echo \"#{master_ip} puppet\" >> /etc/hosts").once
+      subject.set_etc_hosts
+
+    end
+
+  end
+
```
samwoods1: You're going to hate me. In my check-in I'm about to push up for a PR, I removed both of these functions. I found a better way to handle the perf_init, and the set_etc_hosts is no longer necessary with the new AWS account.
With that said, I think the approach is perfect, If we had similar tests for all of the other functions we'd be in great shape.

spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -41,13 +75,217 @@ def options
...
       allow(subject).to receive(:on)
       expect(subject).to receive(:on).with(hosts[0], 'chkconfig iptables off').once
       subject.install_epel_packages
     end
 
   end
 
+  context '.setup_r10k' do
+
+    # TODO: is this the best way to handle masters?
```
samwoods1: I see no problem with this. I don't think we need to be fancy. 
We could test a list of masters with various lists of machines, some of which have master role and some that have compile_master, but then we're crossing the line into testing beaker methods rather than what we care about in our own methods.

spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -41,13 +75,217 @@ def options
...
+    let!(:masters) {[{'platform' => Beaker::Platform.new('centos-6.5-x86_64')}]}
+
+    it 'adds the private key, configures ssh, and installs git on each master' do
+
+      allow(subject).to receive(:masters).and_return(masters)
+      expect(subject).to receive(:select_hosts).with({:roles => ['master', 'compile_master']}).and_return(masters)
+
+      # TODO: should these use the private_key and ssh_config variables or are these matchers sufficient?
```
samwoods1: Looks good to me 👍

spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -41,13 +75,217 @@ def options
...
+      it 'does nothing' do
+        expect(host).to receive(:file_exist?).with(("/etc/cloud/cloud.cfg")).and_return(false)
+        expect(subject).to_not receive(:on)
+        subject.cloud_config(host)
+      end
+    end
+
+  end
+
+  describe '#ec2_workarounds' do
```
samwoods1: Sadly, I just deleted this method too, since it is no longer necessary with the new AWS account.

bcpuppet:
Closing this PR; I'll create a new one based on Sam's updates.

PR #201, opened 2018-03-08
(QA-3457) Switch to using ABS for AWS provisioning

samwoods1:
Updated hosts files to use ABS hypervisor
Broke out rake tasks into multiple tasks, one to provision, one to execute beaker, one to encapsulate both of those and finally, one to deprovision.
Removed workarounds for AWS that are no longer necessary: 10_etc_hosts.rb, 40_ec2_fixes.rb, 20_use_internal_ips.rb and some of the FOSS install code in perf_helper.rb
Fixed 20_foss_install/10_install_dev_repos.rb - it was missing the call to the helper method

### DIFF
--- a/rakefile
+++ b/rakefile
@@ -12,9 +12,10 @@ desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and th
 rototiller_task :performance do
   Rake::Task["performance_provision_with_abs"].execute unless ENV['ABS_RESOURCE_HOSTS']
   Rake::Task["performance_without_provision"].execute
-  Rake::Task["performance_deprovision_with_abs"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' || ((@beaker_cmd.nil? || @beaker_cmd.result.exit_code != 0) && ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
+  Rake::Task["performance_deprovision_with_abs"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' ||
+      ((@beaker_cmd.nil? || @beaker_cmd.result.exit_code != 0)&&
+          ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
 end
-
 
 def retry_abs_request(uri, body, timeout, time_increment, invalid_response_bodies = [], valid_response_codes = [202, 200])
   req = Net::HTTP::Post.new(uri, 'Content-Type' => 'application/json')
@@ -46,11 +47,15 @@ def retry_abs_request(uri, body, timeout, time_increment, invalid_response_bodie
 end
 
 desc 'Provision systems for the performance task with ABS - included in the performance task'
-rototiller_task :performance_provision_with_abs do
+rototiller_task :performance_provision_with_abs do |t|
+  t.add_env({:name => 'ABS_OS', :default => 'centos-7-x86-64-west', :message => 'The OS to provision with ABS: centos-7-x86-64-west (for AWS), or centos-7-x86_64 (for vmpooler)'})
   # Creating a random job id, since this is required by ABS and needs to be unique. In CI, this task will not be executed.
   ENV['ABS_JOB_ID'] = rand(100000..999999).to_s unless ENV['ABS_JOB_ID']
   uri = URI("#{BASE_ABS_URI}request")
-  req_body = {'resources': {'centos-7-x86-64-west': 2},'job': { 'id': "#{ENV['ABS_JOB_ID']}", 'tags': {'user': "#{ENV['USER']}"}}}.to_json
+
+  # Allows us to switch between AWS and VMPooler by selecting different ABS os's.
+  # centos-7-x86-64-west is an AWS image, centos-7-x86_64 is vmpooler
+  req_body = {'resources': {"#{ENV['ABS_OS']}": 2},'job': { 'id': "#{ENV['ABS_JOB_ID']}", 'tags': {'user': "#{ENV['USER']}"}}}.to_json
 
   # Wait up to 20 minutes for a response from ABS with provisioned hosts
   res = retry_abs_request(uri, req_body, 1200, 10, ['', nil])
@@ -171,7 +176,7 @@ rototiller_task :performance_without_provision do |t|
     end
     command.add_option do |option|
       option.name = '--preserve-hosts'
-      option.message = 'Whether to preserve hosts or not, if using the default ABS workflow, this will not be respected.'
+      option.message = 'Whether to preserve hosts or not.'
       option.add_argument do |arg|
         arg.name = 'always'
         arg.add_env({:name => 'BEAKER_PRESERVE_HOSTS'})
@@ -211,6 +216,9 @@ desc 'Run Performance setup and a very short iteration of gatling tests'
 rototiller_task :acceptance do |t|
   ENV['ENVIRONMENT_TYPE'] = 'gatling'
   ENV['PUPPET_GATLING_SCENARIO'] = 'acceptance.json'
+  ENV['BEAKER_HOSTS'] ||= 'config/beaker_hosts/pe-vmpooler.cfg'
+  # Need to change this from an AWS OS image to a vmpooler OS image
+  ENV['ABS_OS'] ||= 'centos-7-x86_64'
   Rake::Task["performance"].execute
 end
 
--- a/rakefile
+++ b/rakefile
@@ -1,13 +1,80 @@
 require 'rototiller'
 require 'rspec/core/rake_task'
+require 'net/http'
+require 'timeout'
+require 'json'
+
+BASE_ABS_URI = 'https://cinext-abs.delivery.puppetlabs.net/api/v2/'
 
 task :default => :performance
 
-desc 'Run Performance setup'
-rototiller_task :performance do |t|
+desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
+rototiller_task :performance do
+  Rake::Task["performance_provision_with_abs"].execute unless ENV['ABS_RESOURCE_HOSTS']
+  Rake::Task["performance_without_provision"].execute
+  Rake::Task["performance_deprovision_with_abs"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' || ((@beaker_cmd.nil? || @beaker_cmd.result.exit_code != 0) && ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
+end
+
+
+def retry_abs_request(uri, body, timeout, time_increment, invalid_response_bodies = [], valid_response_codes = [202, 200])
+  req = Net::HTTP::Post.new(uri, 'Content-Type' => 'application/json')
+  req.body = body
+  http = Net::HTTP.new(uri.host, uri.port)
+  http.use_ssl = true
+
+  puts "sending request body: #{body}"
+  puts "to uri: #{uri}"
+
+  res = nil
+  Timeout::timeout(timeout) do
+    while res.nil? || invalid_response_bodies.include?(res.body) || !valid_response_codes.include?(res.code)
+      begin
+        res = http.request(req)
+      rescue => ex
+        puts ex
+        puts ex.backtrace
+      end
+      if res.nil? || invalid_response_bodies.include?(res.body)
+        puts "#{Time.now} - Retrying ABS request..."
+        sleep(time_increment)
+      else
+        break
+      end
+    end
+  end
+  res
+end
+
+desc 'Provision systems for the performance task with ABS - included in the performance task'
+rototiller_task :performance_provision_with_abs do
+  # Creating a random job id, since this is required by ABS and needs to be unique. In CI, this task will not be executed.
+  ENV['ABS_JOB_ID'] = rand(100000..999999).to_s unless ENV['ABS_JOB_ID']
+  uri = URI("#{BASE_ABS_URI}request")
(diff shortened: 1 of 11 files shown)


rakefile
```diff
@@ -1,13 +1,85 @@
...
+  req = Net::HTTP::Post.new(uri, 'Content-Type' => 'application/json', )
+  req.body = {'resources': {'centos-7-x86-64-west': 2},'job': { 'id': "#{ENV['ABS_JOB_ID']}", 'tags': {'user': 'jenkins','jenkins_build_url': 'https://jenkins/job/local-perf-test'}}}.to_json
+  http = Net::HTTP.new(uri.host, uri.port)
+  http.use_ssl = true
+
+  puts "sending request body: #{req.body}"
+  puts "to uri: #{uri}"
+  res = nil
+
+  # Wait up to 5 minutes for a response from ABS with provisioned hosts
```
er0ck: this could probably be placed in a method, outside the task.  as-is, it's hard for me to follow what the task is doing exactly.

rakefile
```diff
@@ -1,13 +1,85 @@
...
+        sleep(10)
+      end
+    end
+  end
+  ENV['ABS_RESOURCE_HOSTS'] = res.body
+  puts "ABS_RESOURCE_HOSTS=#{ENV['ABS_RESOURCE_HOSTS']}"
+end
+
+rototiller_task :performance_deprovision_with_abs do
+  raise 'performance_deprovision_with_abs task requires values for env variables: ABS_RESOURCE_HOSTS and ABS_JOB_ID' unless ENV['ABS_RESOURCE_HOSTS'] && ENV['ABS_JOB_ID']
```
er0ck: do we do this currently due to deficiencies in rototiller's required env vars?  there is very likely nothing to handle this exception.

rakefile
```diff
@@ -106,7 +178,7 @@ rototiller_task :performance do |t|
       option.name = '--preserve-hosts'
       option.message = 'Whether to preserve hosts or not'
       option.add_argument do |arg|
-        arg.name = 'always'
+        arg.name = 'never'
```
er0ck: this is probably useful outside of this commit.  should be separated.  e.g. if the other code needs reverting, we'd want this to stay.  correct?

rakefile
```diff
@@ -146,6 +218,13 @@ rototiller_task :acceptance do |t|
   Rake::Task["performance"].execute
 end
 
+desc 'Run Performance setup and a very short iteration of gatling tests locally'
```
er0ck: this looks like another opportunity to split out via separate commit

rakefile
samwoods1: Oh, the check for BEAKER_INSTALL_TYPE makes sense because some of the other env vars are optional depending on what it is set to, but you're right, there's no need for this check.

rakefile
samwoods1: Actually, I need to update my PR, apparently this is not going to be respected either way with the new ABS workflow, so we'll need to do our own cleanup with the new de-provision task.

rakefile
er0ck: yeah, there is a ticket for pattern matching on env var values for tiller.  but i think that thing needs empty env var and other edge cases of the env var interface to be cleared up first anyway

rakefile
samwoods1: Hmm, looks like I copied that task and was going to modify it, but ended up making things work a different way, I'm removing this.

rakefile
```diff
@@ -1,13 +1,85 @@
...
+        sleep(5)
+      end
+    end
+  end
+
+end
+
+desc 'Execute beaker performance setup and tests against existing hosts specified by ABS_RESOURCE_HOSTS env var, intended to be used in CI - use task "performance" for local/dev execution'
+rototiller_task :performance_without_provision do |t|
+  raise "The ABS_RESOURCE_HOSTS env var must be set to a valid value to continue" unless
```
RandellP: seems odd that the error message mentions only one env variable, but the condition checks for either of two env variables.

setup/helpers/perf_helper.rb
```diff
@@ -343,19 +249,33 @@ def perf_install_foss
...
+        master.upgrade_package(nss_package_name)
+      else
+        Beaker::Log.warn("Don't know what nss package to use for #{variant} so not installing one")
+      end
     end
 
-    step "Sign agent certificates" do
-      # Failed agent run will request a certificate
-      on agents, 'puppet agent -t', {:accept_all_exit_codes => true}
-      on master, 'puppet cert sign --all'
```
RandellP: where did the cert signing go?  I don't see anything signing certs to make up for this line being removed.

rakefile
```diff
@@ -1,13 +1,85 @@
...
+        puts ex
+        puts ex.backtrace
+      end
+      if res.nil? || res.body == ''
+        puts 'Awaiting ABS_RESOURCE_HOSTS response from ABS...'
+        sleep(10)
+      end
+    end
+  end
+  ENV['ABS_RESOURCE_HOSTS'] = res.body
```
er0ck: do we really want to be writing to the user's env var space? ever?

samwoods1: ABS requires this env var to be set... What would the alternative be?

setup/helpers/perf_helper.rb
samwoods1: It is part of the normal setup. I had to call it explicitly before because I couldn't use the default setup method, but now that we have access to internal resources it uses the default setup which takes care of it.

rakefile
er0ck: it does?  i thought abs was a remote webservice.

samwoods1: More specifically, beaker-abs requires it

er0ck: ah.  right.  that's annoying, but difficult to provide a better interface i guess.  
i'm glad you are mentioning that you are setting it on the next line, to the user.

carry on

rakefile
```diff
@@ -1,13 +1,79 @@
...
 
 task :default => :performance
 
-desc 'Run Performance setup'
-rototiller_task :performance do |t|
+desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
+rototiller_task :performance do
+  Rake::Task["performance_provision_with_abs"].execute unless ENV['ABS_RESOURCE_HOSTS']
+  Rake::Task["performance_without_provision"].execute
+  Rake::Task["performance_deprovision_with_abs"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' || (@beaker_cmd.result.exit_code != 0 && ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
```
er0ck: what happens when @beaker_cmd has not been set?

setup/helpers/perf_helper.rb
```diff
@@ -343,19 +249,33 @@ def perf_install_foss
       if package_build_version
         Beaker::Log.notify("Installing OSS Puppet Server version '#{package_build_version}'")
         install_puppetlabs_dev_repo master, 'puppetserver', package_build_version,
-                                    repo_config_dir, @install_opts
-      else
+                                    repo_config_dir, install_opts
+        else
```
er0ck: errant indent?

rakefile
samwoods1: Hmm, looks like this would throw an exception and never run, which probably would be the desired behavior. We probably would want to assume a failure if @beaker_cmd has not been set.

er0ck: i guess that's fine.  do you think the user will know why or what task to run first?

samwoods1: I tried to update the readme also. I am also updating the descriptions of the tasks to try to be more clear.

samwoods1:
I added a new commit to this PR to support running against VMPooler for the acceptance tests.

rakefile
```diff
@@ -1,13 +1,83 @@
...
 
 task :default => :performance
 
-desc 'Run Performance setup'
-rototiller_task :performance do |t|
+desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
+rototiller_task :performance do
+  Rake::Task["performance_provision_with_abs"].execute unless ENV['ABS_RESOURCE_HOSTS']
+  Rake::Task["performance_without_provision"].execute
+  Rake::Task["performance_deprovision_with_abs"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' || ((@beaker_cmd.nil? || @beaker_cmd.result.exit_code != 0) && ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
```
RandellP: This looks fine, but is pretty hard to read (mostly due to it's length and where line breaks might end up).  Not worth a new commit to fix, but if you end up making a commit to this PR for another reason, it would be nice to make this more readable in that commit.

rakefile
```diff
@@ -1,13 +1,83 @@
...
+  res
+end
+
+desc 'Provision systems for the performance task with ABS - included in the performance task'
+rototiller_task :performance_provision_with_abs do
+  # Creating a random job id, since this is required by ABS and needs to be unique. In CI, this task will not be executed.
+  ENV['ABS_JOB_ID'] = rand(100000..999999).to_s unless ENV['ABS_JOB_ID']
+  uri = URI("#{BASE_ABS_URI}request")
+
+  # Allows us to switch between AWS and VMPooler
```
RandellP: how?

rakefile
```diff
@@ -104,7 +174,7 @@ rototiller_task :performance do |t|
     end
     command.add_option do |option|
       option.name = '--preserve-hosts'
-      option.message = 'Whether to preserve hosts or not'
+      option.message = 'Whether to preserve hosts or not, if using the default ABS workflow, this will not be respected.'
```
RandellP: Thanks to the great code you added, this is no accepted... isn't it?

rakefile
```diff
@@ -143,6 +214,8 @@ desc 'Run Performance setup and a very short iteration of gatling tests'
 rototiller_task :acceptance do |t|
   ENV['ENVIRONMENT_TYPE'] = 'gatling'
   ENV['PUPPET_GATLING_SCENARIO'] = 'acceptance.json'
+  ENV['BEAKER_HOSTS'] = 'config/beaker_hosts/pe-vmpooler.cfg'
+  ENV['ABS_OS'] = 'centos-7-x86_64'
```
RandellP: on line 55 you set this if not set... why does it need setting here?  And if here, why on 55?  And why different values.  It may all be good, and if so, some more comments would really help.

RandellP:
I thought we were concerned about always running acceptance in vmpooler, since aws has lots of it's own challenges?   I think we could create two tasks acceptance_abs and acceptance_vmpooler.  Then have acceptance run the vmpooler one.  That way we can run vmpooler for fast feedback, and then use abs as a final check before merge.

Also, the readme needs an update mentioning how to make a run happen in vmpooler one way or another.  I think the section "Set up Puppet Enterprise and Gatling but do not execute a gatling scenario" is where it is most useful.  You wouldn't want to do a full run and execute in vmpooler... but a no execute run would be good for dev work.

samwoods1:
@RandellP I think the vast majority of the time we are just going to want to run the acceptance tests in vmpooler. With the new AWS account, and the fact that we're using the base Centos7 image in both AWS and VMPooler now, things should be much more stable.  If we really need to run acceptance tests in AWS instead of vmpooler, we can kick it off by simply changing some env vars before the run.

RandellP:
tests passed.  The readme could use a blurb about how to run acceptance in aws.  Just document the env vars so a person doesn't have to go digging to find out how to do it.

rakefile
```diff
@@ -1,13 +1,85 @@
...
+      else
+        break
+      end
+    end
+  end
+  res
+end
+
+desc 'Provision systems for the performance task with ABS - included in the performance task'
+rototiller_task :performance_provision_with_abs do
```
er0ck: these tasks should probably be organized by namespace later...
something like `performance:provision_abs` etc

but i'm fine deferring UX changes like this until later

rakefile
```diff
@@ -1,13 +1,85 @@
...
+  # Creating a random job id, since this is required by ABS and needs to be unique. In CI, this task will not be executed.
+  ENV['ABS_JOB_ID'] = rand(100000..999999).to_s unless ENV['ABS_JOB_ID']
+  uri = URI("#{BASE_ABS_URI}request")
```
er0ck: does this have a ticket for fixing up later?  we could have easily used a truly unique id 🤷‍♂️ (uuid)

rakefile
```diff
@@ -1,13 +1,85 @@
...
+
+  # Allows us to switch between AWS and VMPooler by selecting different ABS os's.
+  # centos-7-x86-64-west is an AWS image, centos-7-x86_64 is vmpooler
+  abs_os = ENV['ABS_OS'].nil? ? 'centos-7-x86-64-west' : ENV['ABS_OS']
```
er0ck: why isn't this a rototiller env_var?

setup/helpers/perf_helper.rb
```diff
@@ -668,7 +572,9 @@ def setup_gatling_proxy
     end
     step 'install rvm, bundler' do
       on metric, 'gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3'
-      on metric, 'curl -sSL https://get.rvm.io | bash -s stable --ruby=2.2.5'
+      on metric, 'curl -sSL https://get.rvm.io | bash -s stable'
+      # Work around install issue in Centos7: https://github.com/rvm/rvm/issues/4234
+      on metric, 'rvm install 2.2.5 --disable-binary'
```
er0ck: if this is to work around known 2.2.5 rvm issues in centos, this change should be in a separate commit.  oh, i see the comment now 🤣 
but yeah, separate commit for ease of revert later

er0ck: i guess install also implicitly uses that latest version?  or maybe if there is only one version?
the man page is vague on it.

rakefile
samwoods1: It expects an integer... So, no uuid.

er0ck:
:shipit:

PR #202, opened 2018-03-14
(maint) Use `hostname` to designate puppet's server

Magisus:
The value of beaker's `master` variable is not always the same as a perf
node's `hostname`. When invoking puppet on these hosts, we need
to use the hostname as the server to contact, and not the external alias
given to beaker. This commit updates the `puppet agent` invocation to use
`hostname`, which works in both cases.

### DIFF
--- a/jenkins-integration/beaker/install/foss/80_run_agent_on_master.rb
+++ b/jenkins-integration/beaker/install/foss/80_run_agent_on_master.rb
@@ -5,5 +5,9 @@ test_name 'Run puppet agent on the master to prime directories' do
   # ahead of the Gatling run avoids collisions in the priming that can happen
   # when multiple agent runs happen at the same time.  See PUP-6651 for more
   # information.
-  on(master, "puppet agent -t --server #{master}")
+  #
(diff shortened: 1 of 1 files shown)


Magisus:
This works, but feels a little weird, like it doesn't actually solve the underlying problem of the node not knowing its DNS aliases. I'd appreciate advice on a better way to fix this problem.

lindboe:
Why can't we change the Beaker master variable to point to the right thing? AFAIK the "alternate" node name (like node3993) is only necessary to trigger Razor

Magisus:
I can try running with the name puppet puts on the cert: perf-bl3996.delivery.puppetlabs.net. If that is what's needed, should we maybe updated confluence to reflect that? Where did the 11-12-13 numbers come from anyway?

lindboe:
Whoops sorry, missed this comment. 11-12,13 are arbitrary, but so were the node assignments in Razor, and DNS recognizes those nodes as 11-12-13 so it seemed simpler to keep that

Magisus:
This seems like the wrong approach, closing.

PR #205, opened 2018-03-19
(QA-3456) Use Gatling feeders to simulate facts change

doug-rosser:
With this change, facts are no longer a hard-coded string inside the .scala file itself. Most of the hard-coded facts string has been moved to external file (IrvingFactsChange_facts_raw.txt) with a small "header" inline in order to position our custom facts so they are in the correct part of the URL-encoded JSON sent to Puppet Server.

There are two classes of changeable facts. The first set (IrvingFactsChange_custom_facts.csv) is for facts that change with every agent run. There should be enough lines in the file to supply data for every agent run. In other words, for 600 agents running four hours (8 runs per agent) there should be 4800 + 1 lines. If there aren't enough lines for the full Gatling job run, the test will produce an error. This is a feature, not a bug.

The second class of facts are unique to each agent, but do not change during the run (user-files/facts/IrvingFactsChange_custom_hostf.csv). There should be one line for every agent configured for the test.

The facts are all pre-generated before the test in order to remove the possibility of slowing down the test driver side. The facts are all loaded into memory, so facts on the order of gigabytes will probably produce memory errors on the driver node. Facts should be in the following format:

`%22customfact1%22%3A%22KO8mxN5FntvC6NwW%22%2C`

To disambiguate the above:

```
%22 customfact1 %22 %3A %22 KO8mxN5FntvC6NwW %22 %2C
 "  customfact1  "   :   "  KO8mxN5FntvC6NwW  "   ,
```
The key "customfact1" as well as its value "KO8mxN5FntvC6NwW" do not really matter, as long as the values in bulk contain enough data to be effectively random.

A simple script(s) can be utilized to generate facts.

```
echo "Fact" >> tmp.csv
for x in `seq 1 4800` ; do echo "\
%22customfact1%22%3A%22`pwgen -s -1 16`%22%2C\
%22customfact2%22%3A%22`pwgen -s -1 24`%22%2C\
%22customfact3%22%3A%22`pwgen -s -1 32`%22%2C\
" >> tmp.csv ; done
```
```
echo "HostFact" >> tmp.csv
for x in `seq 1 600` ; do echo "\
%22hostfact1%22%3A%22`pwgen -s -1 16`%22%2C\
%22hostfact2%22%3A%22`pwgen -s -1 24`%22%2C\
%22hostfact3%22%3A%22`pwgen -s -1 32`%22%2C\
" >> tmp.csv ; done
```


er0ck:
@doug-rosser this is great.  can you put much of the stuff from above in the commit message as well?

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/IrvingFactsChange.scala
```diff
@@ -0,0 +1,89 @@
+package com.puppetlabs.gatling.node_simulations
```
er0ck: is there a reason this is a new file? is there a reason to not update an existing simulation file?  
also, the other simulation files are titled with the pe version number, not codename.  curious if there is a reason for breaking this pattern.

simulation-runner/facts/IrvingFactsChange_custom_facts.csv
```diff
@@ -0,0 +1,4801 @@
+Fact
+%22customfact1%22%3A%22vNSUclWAYnEJVu18%22%2C%22customfact2%22%3A%22jN5JseeQUdNtsILhepVR63mj%22%2C%22customfact3%22%3A%22iFfvllM0i37wjY1vBNg99FxgOpCNAbNh%22%2C
```
samwoods1: Why is it necessary to encode this in the file?  Seems like it would be more readable and easier to update/maintain if we stored it not encoded and executed a method to encode it inline? Although I suppose that could also take up CPU cycles... Not sure which trade-off would be most beneficial here.

simulation-runner/config/nodes/IrvingFactsChange.json
```diff
@@ -0,0 +1,7 @@
+{
+  "simulation_class": "com.puppetlabs.gatling.node_simulations.IrvingFactsChange",
+  "certname_prefix": "gat-agent-",
```
samwoods1: If we use "perf-agent-" instead of "gat-agent-" here, it would be simpler to use with our existing scripts, we could just override the config/node/*.json path and use as is.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/IrvingFactsChange.scala
samwoods1: If we are not updating an existing simulation file in this check-in, we at least need another ticket to update the ApplesToApples simulation.

samwoods1:
> There are two classes of changeable facts. The first set (IrvingFactsChange_custom_facts.csv) is for facts that change with every agent run. There should be enough lines in the file to supply data for every agent run. In other words, for 600 agents running four hours (8 runs per agent) there should be 4800 + 1 lines. If there aren't enough lines for the full Gatling job run, the test will produce an error. This is a feature, not a bug.
> 
> The second class of facts are unique to each agent, but do not change during the run (user-files/facts/IrvingFactsChange_custom_hostf.csv). There should be one line for every agent configured for the test.

It would be great if this were automated, so that we look at the scenario json file, determine how many entries we need for both, and generate the csv files on the spot before executing the scenario rather than checking those into source. Following this model would require us to check in additional files for every different number of iterations/agents.

> The facts are all loaded into memory, so facts on the order of gigabytes will probably produce memory errors on the driver node.
If this becomes an issue, we can switch to using a file reader that stores one line at a time in memory instead of pulling it all into memory at the same time. I think we may run into this sooner rather than later once we start adding more and more configurability to the tests.

simulation-runner/facts/IrvingFactsChange_custom_facts.csv
doug-rosser: Eventually, we will want to determine how to express "complexity" for these facts inside the json configuration file (summer intern project?) but in any case, the data should be left in the "native" format as modifying/encoding data on the fly really doesn't really buy us anything (IMO) at the risk of affecting the performance of the driver. End result is that users won't have to directly monkey with the encoded JSON, which I think everyone agrees is an eyesore and prone to mistakes.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/IrvingFactsChange.scala
doug-rosser: Eventually, I'd think we would want the "recording" script to make these modifications automagically. Before that happens, I think we should figure out a safe method to turn fact randomization off, so that other teams won't hesitate to use the integrated code.

simulation-runner/facts/IrvingFactsChange_custom_facts.csv
samwoods1: 👍

PR #206, opened 2018-03-19
(SERVER-2135) (SERVER-2103) (SERVER-2128) Various gatling job additions

Magisus:
(SERVER-2135)
This commit exposes a parameter on the puppetserver-infinite job to
configure the max number of active JRuby instances allowed. When set to
'default', it will use puppetserver's default of the number of CPUs - 1.

(SERVER-2103)
This commit adds a param to the puppetserver-infinite job to configure
max-requests-per-instance. It defaults to 0, meaning JRubies are never
flushed.

(SERVER-2128)
This commit adds a groovy function that walks through the params map and
exports its keys and values into a file for upload to s3.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -435,12 +435,24 @@ def step900_collect_driver_artifacts() {
     archive "simulation-runner/results/**/*.log.gz"
 }
 
+def create_params_file(archive_dir) {
+    def keys = params.keySet()
+    def param_string = ""
+    for (i = 0 i < keys.length; i++) {
+        param_string += "${keys[i]}: ${params.get(keys[i])}\n"
+    }
+    sh "echo ${param_string} > ${archive_dir}/job_params.txt"
+}
+
 def step905_publish_artifacts_to_s3(job_name) {
     def archive_dir = job_name + '-' + (new Date().format("yyyy-MM-dd-HH:mm:ss"))
     sh "mkdir -p ${archive_dir}/results"
     // Copy all gatling output info into the dir to be uploaded
     sh "cp -R ./puppet-gatling/${job_name}/* ${archive_dir}"
     sh "cp simulation-runner/results/**/*.log.gz ${archive_dir}/results"
+
+    create_params_file(archive_dir)
+
     step([
         $class: 'S3BucketPublisher',
         entries: [
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -328,27 +328,17 @@ def step080_customize_java_args(script_dir, server_heap_settings, server_era) {
 
 def step081_customize_jruby_jar(script_dir, jruby_jar, server_era) {
     def jar_string = ""
-    def JRUBY_VERSION_SET = true
 
-    // This garbage can be replaced with either getBinding().hasVariable("JRUBY_VERSION")
-    // or via a params map (params.JRUBY_VERSION == null)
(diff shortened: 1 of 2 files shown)


jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -361,6 +361,15 @@ def step081_customize_jruby_jar(script_dir, jruby_jar, server_era) {
...
+    if ("${MAX_INSTANCES}" != "default") {
+        settings = settings ?: []
+        settings << [
+                      file: "/etc/puppetlabs/puppetserver/conf.d/puppetserver.conf",
+                      path: "jruby-puppet.maxt-active-instances",
+                      value: "${MAX_INSTANCES}"
+                    ]
+    }
+
     if (settings == null) {
```
justinstoller: We'll never go down this branch now, right?

Magisus: We will, if `MAX_INSTANCES` is left "default" AND there is no `hocon_settings` block in the job definition. I think we have at least one setting in all existing jobs, but in theory we could have one without it.

Magisus:
I've added a commit to this PR to fix the most recent failure in the s3 upload job, where it was attempting to copy files from a directory that had already been deleted.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -361,6 +361,15 @@ def step081_customize_jruby_jar(script_dir, jruby_jar, server_era) {
 }
 
 def step085_customize_hocon_settings(script_dir, settings, server_era) {
+    if ("${MAX_INSTANCES}" != "default") {
```
lindboe: FWIW this commit is refactored enough in the MAX_REQUESTS commit that I think you could just combine them

Magisus: Sounds good, will do.

Magisus:
It might already be in puppetserver's log, which we are archiving along with the other job artifacts. I'll take a look.

Magisus:
I take that back, we are not archiving puppetserver's log (and honestly I don't think we should), so it might be worth ticketing that.

lindboe:
Cool. Pending that I'm just curious to hear your thoughts if it would be reasonable to do something else to log the "default" value if we need to, or if we'd rather proceed and deal with that later, I'm open to either.

lindboe:
Ah github hadn't updated. Alright if you want to ticket it I'm good with that.

Magisus:
We do, however, log it. From my machine with max-instances=2:
```
2018-03-20 15:19:32,805 DEBUG [clojure-agent-send-pool-0] [p.s.j.i.jruby-agents] Priming JRubyInstance 1 of 2
```

Magisus:
Ticket: https://tickets.puppetlabs.com/browse/SERVER-2158

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -433,16 +433,27 @@ def step900_collect_driver_artifacts() {
...
-    dir('simulation-runner/results') {
-        deleteDir()
+}
+
+def create_params_file(archive_dir) {
+    def keys = params.keySet()
+    def values = params.values()
+    def param_string = ""
+    for (i = 0 i < keys.length; i++) {
+        param_string += "${keys[i]}: ${values[i]}\n"
```
justinstoller: Do keySets and values ordered in the same way as each other?

Magisus: Good question. I should get the value based on the key in question.

Magisus: Updated.

PR #208, opened 2018-03-21
(maint) Use size() instead of length on param map

Magisus:
For the params map, keySet() does not return an array, but rather an
UnmodifiableSet. Therefore, `length` cannot be used to check its size;
the size() method must be used instead.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -438,7 +438,7 @@ def step900_collect_driver_artifacts() {
 def create_params_file(archive_dir) {
     def keys = params.keySet()
     def param_string = ""
-    for (i = 0; i < keys.length; i++) {
+    for (i = 0; i < keys.size(); i++) {
         param_string += "${keys[i]}: ${params.get(keys[i])}\n"
     }
     sh "echo ${param_string} > ${archive_dir}/job_params.txt"


jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -438,7 +438,7 @@ def step900_collect_driver_artifacts() {
 def create_params_file(archive_dir) {
     def keys = params.keySet()
     def param_string = ""
-    for (i = 0; i < keys.length; i++) {
+    for (i = 0; i < keys.size(); i++) {
```
lindboe: For people who don't know these things, (like me until just now!), it might be helpful to note in the commit that length returns the capacity of an array and size returns the number of elements. If I'm reading things right.

Magisus: Oh is that it? The error I got in Jenkins was rather cryptic:
```
org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: unclassified field java.util.Collections$UnmodifiableSet length
```
I just happened to have seen `length` and `size()` described as interchangeable by the Groovy docs on Arrays http://grails.asia/groovy-array-manipulation-examples.

Magisus: Guess that's what I get for using not-real-docs.

Magisus: Oh look, I found a better thing:
```
map.each { entry ->
    println "Name: $entry.key Age: $entry.value"
}
```

Magisus:
TIL the `params` map doesn't implement `each`:
```
java.lang.UnsupportedOperationException: Calling public static java.util.Map org.codehaus.groovy.runtime.DefaultGroovyMethods.each(java.util.Map,groovy.lang.Closure) on a CPS-transformed closure is not yet supported (JENKINS-26481); encapsulate in a @NonCPS method, or use Java-style loops
```

Magisus:
Updated with commit message.

jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
lindboe: Ah yeah that's at least the case for Java so since it was equally confusing I thought maybe it would be the same here (idk how Groovy interop works, especially now in the Jenkins world)

PR #209, opened 2018-03-22
(maint) Convert KeySet to Array before attempting to index

Magisus:
Java's set type does not allow indexing with bracket notation, since it
is optimized for membership rather than for iteration. This commit
converts the params key set to an array before attempting to iterate
through it.

### DIFF
--- a/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
+++ b/jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
@@ -436,7 +436,7 @@ def step900_collect_driver_artifacts() {
 }
 
 def create_params_file(archive_dir) {
-    def keys = params.keySet()
+    def keys = params.keySet().toArray()
     def param_string = ""
     for (i = 0; i < keys.size(); i++) {
         param_string += "${keys[i]}: ${params.get(keys[i])}\n"


jenkins-integration/jenkins-jobs/common/scripts/jenkins/pipeline.groovy
```diff
@@ -436,7 +436,7 @@ def step900_collect_driver_artifacts() {
 }
 
 def create_params_file(archive_dir) {
-    def keys = params.keySet()
+    def keys = params.keySet().toArray()
```
justinstoller: Are `params` a groovy map? Can we call these methods on it? http://groovy-lang.org/groovy-dev-kit.html#_iterating_on_maps

justinstoller: I don't know if `params` is a map, but from the groovy console in Jenkins it looks like I can use those methods on maps.
http://puppetserver-perf-driver55.delivery.puppetlabs.net:8080/script

Magisus: Nope! Jenkins doesn't support it, I found out yesterday >.>
https://github.com/puppetlabs/gatling-puppet-load-test/pull/208#issuecomment-375123105

justinstoller: That's terrifying

justinstoller: More reason to upgrade our Jenkins install, I guess? It looks like it was fixed awhile back but I think requires Jenkins 2.7+

https://wiki.jenkins.io/display/JENKINS/Pipeline+Groovy+Plugin#PipelineGroovyPlugin-2.33(May30,2017)

PR #213, opened 2018-03-28
(SLV-15) Add tests to perf_helper_spec

billclaytor:
The following tests have been added:
set_etc_hosts
setup_r10k
perf_install_pe
has_cent7_repo?
get_cent7_repo
get_latest_server_version
get_latest_agent_version
get_latest_release_agent_version

~~`perf_install_pe` still needs some work. I'm not sure how `@options[:answers]` should be handled and I need to circle back on some of the other TODO items. Any suggestions are appreciated.~~

EDIT: I've updated `perf_install_pe` to remove `use_meep`, updated `get_cent7_repo` to mock `has_cent7_repo`, and moved the `@options[:answers]` check. This should be ready to merge now.

### DIFF
--- a/spec/setup/helpers/perf_helper_spec.rb
+++ b/spec/setup/helpers/perf_helper_spec.rb
@@ -220,6 +220,7 @@ describe PerfHelperClass do
       it 'returns nil' do
         stub_const('Beaker::Log', test_beaker_log)
         allow(test_beaker_log).to receive(:notify)
+        allow(subject).to receive(:has_cent7_repo?).and_return(false)
         expect(subject.get_cent7_repo(test_response_lines, test_package)).to eq(test_expected_result)
       end
 
--- a/spec/setup/helpers/perf_helper_spec.rb
+++ b/spec/setup/helpers/perf_helper_spec.rb
@@ -99,114 +99,24 @@ describe PerfHelperClass do
 
       test_is_pre_aio_version = false
 
-      context 'when we are using meep' do
+      before {
+        test_options = { :answers => {} }
+        subject.instance_variable_set(:@options, test_options)
+      }
 
-        # TODO: how should @options / @options[:answers] be handled?
-        before {
-          test_options = { :answers => {} }
-          #subject.instance_variable_set(:@options, { })
-          subject.instance_variable_set(:@options, test_options)
-        }
+      test_pe_ver = '2017.3'
+      let!(:master) { {'pe_ver' => test_pe_ver} }
 
-        test_use_meep = true
-        test_pe_ver = '2017.3'
-        let!(:master) { {'pe_ver' => test_pe_ver} }
+      test_updated_options = {:answers=>{"puppet_enterprise::profile::master::r10k_remote"=>"/opt/puppetlabs/server/data/puppetserver/r10k/control-repo", "puppet_enterprise::profile::master::r10k_private_key"=>"/root/.ssh/id_rsa"}}
 
-        test_updated_options = {:answers=>{"puppet_enterprise::profile::master::r10k_remote"=>"/opt/puppetlabs/server/data/puppetserver/r10k/control-repo", "puppet_enterprise::profile::master::r10k_private_key"=>"/root/.ssh/id_rsa"}}
+      it 'includes the dashboard and installs pe' do
 
-        it 'includes the dashboard and installs pe' do
+        expect(subject).to receive(:is_pre_aio_version?).and_return(test_is_pre_aio_version)
 
-          expect(subject).to receive(:is_pre_aio_version?).and_return(test_is_pre_aio_version)
-          expect(subject).to receive(:use_meep?).with(master['pe_ver'] || options['pe_ver']).and_return(test_use_meep)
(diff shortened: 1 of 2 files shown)


spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -49,5 +72,429 @@ def options
...
+        test_updated_options = {:answers=>{"puppet_enterprise::profile::master::r10k_remote"=>"/opt/puppetlabs/server/data/puppetserver/r10k/control-repo", "puppet_enterprise::profile::master::r10k_private_key"=>"/root/.ssh/id_rsa"}}
+
+        it 'includes the dashboard and installs pe' do
+
+          expect(subject).to receive(:is_pre_aio_version?).and_return(test_is_pre_aio_version)
+          expect(subject).to receive(:use_meep?).with(master['pe_ver'] || options['pe_ver']).and_return(test_use_meep)
+          expect(subject).to receive(:master).and_return(master)
+
+          # TODO: @options[:answers] ?
+          # expect(subject.instance_variable_get(:@options)). to eq(test_updated_options)
```
samwoods1: Since perf_install_pe is only reading the values of the options and not changing them, I don't think we need to assert on their values.
I wouldn't worry too much about this, since for our scenarios use_meep? should always evaluate to true. In fact, we should probably remove the if/else and always assume we are using meep.

spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -49,5 +72,429 @@ def options
...
+
+          expect(subject).to receive(:install_lei).once
+          subject.perf_install_pe
+
+        end
+
+      end
+
+      context 'when we are not using meep' do
```
samwoods1: Same as above comment, we probably should just remove this block from PerfHelper since we should never hit it.

spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -49,5 +72,429 @@ def options
...
+        subject.perf_install_pe
+      end
+
+    end
+
+  end
+
+  describe '#has_cent7_repo?' do
+
+    # TODO: stub Net::HTTP to provide the expected response or test the actual URLs?
```
samwoods1: I would definitely stub this, but leaving a TODO is fine for now.

spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -49,5 +72,429 @@ def options
...
+
+      test_is_pre_aio_version = false
+
+      context 'when we are using meep' do
+
+        # TODO: how should @options / @options[:answers] be handled?
+        before {
+          test_options = { :answers => {} }
+          #subject.instance_variable_set(:@options, { })
+          subject.instance_variable_set(:@options, test_options)
```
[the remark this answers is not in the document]
er0ck: i'm fine using this in tests

spec/setup/helpers/perf_helper_spec.rb
er0ck: +1 yagni

spec/setup/helpers/perf_helper_spec.rb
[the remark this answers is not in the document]
billclaytor: I expected to see `@options[:answers]` updated with `r10k_remote` and `r10k_private_key` but when I check with instance_variable_get it doesn't seem to have been updated. I added some puts to perf_install_pe after it sets those values and it appears to be updated when output to the console. I'm probably doing it wrong.

spec/setup/helpers/perf_helper_spec.rb
billclaytor: It is actually stubbing it here and elsewhere; I just wanted to make sure that was the right thing to do.

spec/setup/helpers/perf_helper_spec.rb
billclaytor: Sounds good!

spec/setup/helpers/perf_helper_spec.rb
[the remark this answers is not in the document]
samwoods1: Ah, sorry you're right. I think what you did on line 124 should work, but you need to move it to after the call to subject.perf_install_pe.
You can put the other assertions before that because they have to do with what methods will be called and what params, but if you want to check that values of instance variables changed after executing the method, you need to make sure you assert on those after the method is actually called.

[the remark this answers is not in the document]
billclaytor: Ah, that makes sense, thanks!

spec/setup/helpers/perf_helper_spec.rb
samwoods1: Perfect

billclaytor:
Thanks for the feedback! I've made the corresponding updates and this should be ready to merge now. Please let me know if you have additional suggestions.

er0ck:
it works!  it's been long enough.  if other tests are needed, we'll do 'em later!
:shipit:

PR #214, opened 2018-03-30
(SLV-70) First whack at handling agent retries

doug-rosser:
DO NOT MERGE PRETTY PLEASE WITH SUGAR ON TOP!

Two outstanding issues with this code:
1. It only wraps the first http call, when the 503 error can be
   returned by any of them. The handler needs to be wrapped
   around all the calls... Simple in most languages, but this is
   Scala, after all.
2. I'm sure there are still some bugs. The DSL doesn't act like
   most procedural languages, even after you account for it being
   a functional language. Case in point - you only check for errors,
   not success. IOW most loops and checks will be exited automatically
   on success, you only have to check for Bad Conditions.

To test, modify the settings json to create a Thundering Herd.
Increase the number of agents and reduce the ramp_up_duration until
you start seeing errors in the gatling logs. Also, change the number
of iterations to 1. Then turn on agent retry on the server. Details
of how to do this are at:
https://support.puppet.com/hc/en-us/articles/115003769433-KB-0215-Prevent-a-thundering-herd-Use-max-queued-requests-in-Puppet-Enterprise-2017-3-1-and-later-2017-3-x


johnduarte:
This work should be left open as prior art for the work associated with this epic: https://tickets.puppetlabs.com/browse/SLV-290

PR #219, opened 2018-04-11
(QA-3447) copy gatling files to metrics agent instead of cloning repo

samwoods1:


er0ck:
lgtm.  i'll run it when my beaker pin tests finish

PR #220, opened 2018-04-11
(SLV-87) Merge updated facts into a2a simulation

samwoods1:
I had to update scala (SBT) memory to 3gb to avoid stackoverflow errors.
Also, gave the new 'locales' endpoint a friendly name.


setup/helpers/perf_helper.rb
```diff
@@ -602,7 +602,7 @@ def setup_gatling_proxy
     end
     step 'adjust scala build tool mem' do
       metric.mkdir_p '/usr/share/sbt/conf/'
-      on metric, "echo '-mem   2048' >> /usr/share/sbt/conf/sbtopts"
+      on metric, "echo '-mem 3060' >> /usr/share/sbt/conf/sbtopts"
```
er0ck: this isn't 3GB

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
```diff
@@ -20,6 +20,10 @@ class PerfTestLarge extends SimulationWithScenario {
                          Map("reportFile" -> "IntentionalChangeReport_0044_request.txt")
                        ).circular
 
+	val facts_str  = scala.io.Source.fromFile("user-files/facts/IrvingFactsChange_facts_raw.txt").mkString
+	val facts_fdr  = csv("user-files/facts/IrvingFactsChange_custom_facts.csv")
+	val hostf_fdr  = csv("user-files/facts/IrvingFactsChange_custom_hostf.csv").circular
```
er0ck: these variable names are hard for me to read. i actually thought this was a duplicate assignment

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
```diff
@@ -54,8 +60,8 @@ class PerfTestLarge extends SimulationWithScenario {
...
-			.formParam("transaction_uuid", "feaf75a3-c82c-4e06-9d02-5387f14034a0")
+			.formParam("facts", "%7B%22name%22%3A%22${node}%22%2C%22values%22%3A%7B" + "${Fact}" + "${HostFact}" + facts_str)
```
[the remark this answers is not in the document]
samwoods1: No, I mentioned it in Doug's PR, but we decided to just keep it encoded since it was a static, one time thing. If we were constantly updating the values, then having it auto url-encode would be the way to go.

simulation-runner/src/main/scala/com/puppetlabs/gatling/node_simulations/PerfTestLarge.scala
```diff
@@ -54,8 +60,8 @@ class PerfTestLarge extends SimulationWithScenario {
...
+			.formParam("transaction_uuid", "efcde81b-e435-4630-a940-6229f071b3ed")
```
er0ck: why does the uuid have to change?

PR #223, opened 2018-04-17
(slv-89) use beaker options files for various flavors of gplt setup

er0ck:
This gets us closer to an easier to use setup.  It also helps because we can tell exactly which setup file is needed and in what order, for foss vs. pe.

### DIFF
--- a/setup/helpers/perf_helper.rb
+++ b/setup/helpers/perf_helper.rb
@@ -206,6 +206,8 @@ SSH_CONFIG
         Beaker::Log.notify("Installing OSS Puppet AGENT version '#{puppet_agent_version}'")
         install_puppetlabs_dev_repo master, 'puppet-agent', puppet_agent_version,
                                     repo_config_dir, install_opts
+        install_puppetlabs_dev_repo agent, 'puppet-agent', puppet_agent_version,
+                                    repo_config_dir, install_opts
       else
         abort("Environment variable PUPPET_AGENT_VERSION required for package installs!")
       end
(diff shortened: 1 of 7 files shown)


setup/options/options_foss.rb
```diff
@@ -0,0 +1,24 @@
+{
+  :type                        => 'foss',
+  :pre_suite                   => [
+    'setup/install_gatling/00_pre_install/05_initialize_helpers.rb',
```
billclaytor: It looks like the `pre_install` and `post_install` steps are the same. Would it be possible / worthwhile to specify these in `options_common.rb` and just specify the unique steps in the `options_foss.rb` and `options_pe.rb`?

billclaytor: Makes sense, and there isn't really that much duplication since it's just foss and pe.

setup/helpers/perf_helper.rb
```diff
@@ -251,9 +253,16 @@ def perf_install_foss
...
+      on(master, "puppet config set --section master autosign true")
+      on(master, "service puppetserver start")
       on(master, "puppet agent -t --server #{master}")
     end
 
+    step "Install Puppet Agent." do
+      install_package agent, 'puppet-agent'
+      on(agent, "puppet agent -t --server #{master}")
+    end
+
```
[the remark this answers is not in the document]
RandellP: In the spirit of testing how our customers use it...  might turn it off.  And in our setup it is "supposed" to be only internally accessible... :)  But I suppose if they can get at it at all, they can probably do worse things than add an agent.

billclaytor:
I tested this locally with an acceptance run after updating the rakefile with the ABS provisioning fix (https://github.com/puppetlabs/gatling-puppet-load-test/pull/222) and it works as expected, using the correct options file with no issues.

PR #224, opened 2018-04-19
(SLV-100) Implement reports-only run

billclaytor:
If a Gatling scenario is aborted the reports may not be generated. Gatling can be run in reports only mode to create the reports for a previous run by setting the `reportsOnly` property when running via `Gatling.fromMap(props.build)`. The path of the result folder containing a valid `simulation.log` file must be specified.

`gatling_kickoff.rb` has been updated to check the value of the environment variable `PUPPET_GATLING_REPORTS_ONLY`. If it is equal to "true" that value and the value of the `PUPPET_GATLING_REPORTS_TARGET` variable are passed with the `sbt run` command on the metrics node. Otherwise these variables are passed with "false" and "".

`PuppetGatlingRunner.scala` has been updated to check `PUPPET_GATLING_REPORTS_ONLY`; if it is equal to true the `reportsOnly` property is set with the value of `PUPPET_GATLING_REPORTS_TARGET`.

This was tested with short acceptance runs using vmpooler. One issue encountered during this testing is that the `simulation.log` file was not actually populated with data during the run. When beaker was aborted locally Gatling would actually complete the run on the metrics node, then populate the log file and generate the reports. If the run was aborted by killing the Java process on the metrics node the result was an empty log file. Attempting to run reports for an empty log file results in an exception.

It turns out the default buffer for the log file was too large. Setting `gatling.data.file.bufferSize` to 256 in `gatling.conf` allowed the log file to be populated during a short acceptance run. In this case the reports only run is successful and the reports are generated and copied locally as expected.

### DIFF
--- a/NEW_README.md
+++ b/NEW_README.md
@@ -86,7 +86,7 @@ You'll need to specify the name of the folder on the metrics node containing the
 Look in `~/gatling-puppet-load-test/simulation-runner/results` for a folder that starts with 'PerfTestLarge-' followed by the run id and verify that the folder contains a valid (non-empty) simulation.log file.
 
 If you find that the simulation.log file is not being populated during your run you may need to reduce the log buffer size from the 8kb default.
-Set `gatling.data.file.bufferSize` in gatling.conf to a smaller value like 256 (this may impact performance).
+Set `gatling.data.file.bufferSize` in `simulation-runner/gatling.conf` to a smaller value like 256 (this may impact performance).
 
 To run in reports-only mode, run as you normally would against previously provisioned hosts and set the environment variables PUPPET_GATLING_REPORTS_ONLY=true and PUPPET_GATLING_REPORTS_TARGET=<YOUR_RESULT_FOLDER>.
 
--- a/NEW_README.md
+++ b/NEW_README.md
@@ -55,6 +55,7 @@ Assuming that you ran the performance task with no tests, you can follow the dir
     * When asked for the certname prefix enter any value containing the string ‘agent’. For example perf-agent-0
     * When asked for Puppet classes, enter the configuration which defaults to: role::by_size::large
         * You can use a different puppet class by specifying it as the `PUPPET_SCALE_CLASS` environment variable for the performance_gatling rake task.
+
 ### Getting a pre-recorded run started
 For the following steps, GatlingClassName is the value entered into the ClassName field during the recording step.
 From root of the gatling-puppet-load-test source dir on the Gatling driver (metrics):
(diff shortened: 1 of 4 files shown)


NEW_README.md
```diff
@@ -75,3 +76,17 @@ From root of the gatling-puppet-load-test source dir on the Gatling driver (metr
...
 * atop files containing cpu, mem, disk and network usage overall and broken down per process are available to view on the mom: atop -r /var/log/atop/atop_\<date>
     * See the atop man file for interactive commands
+    
+### Aborting a test run
+The easiest way to immediately kill a test run is to kill the corresponding Java process on the metrics node. Find the process id with top and then use `kill <PID>`.
+
+### Generating reports for aborted runs
+If a Gatling scenario is aborted the reports may not be generated. In this case you can run Gatling in reports-only mode to generate the reports for a previous run. 
+You'll need to specify the name of the folder on the metrics node containing the simulation.log file for the run. 
+Look in `~/gatling-puppet-load-test/simulation-runner/results` for a folder that starts with 'PerfTestLarge-' followed by the run id and verify that the folder contains a valid simulation.log file.
```
RandellP: you say "valid" simulation log.  Is there something specific that needs to be in the file?  Or just anything at all makes it good?

NEW_README.md
```diff
@@ -75,3 +76,17 @@ From root of the gatling-puppet-load-test source dir on the Gatling driver (metr
...
+
+If you find that the simulation.log file is not being populated during your run you may need to reduce the log buffer size from the 8kb default.
+Set `gatling.data.file.bufferSize` in gatling.conf to a smaller value like 256 (this may impact performance).
```
RandellP: Can you document where gatling.conf is here.

NEW_README.md
```diff
@@ -75,3 +76,17 @@ From root of the gatling-puppet-load-test source dir on the Gatling driver (metr
...
+
+To run in reports-only mode set PUPPET_GATLING_REPORTS_ONLY=true and PUPPET_GATLING_REPORTS_TARGET=<YOUR_RESULT_FOLDER>. 
+After the run you should see that the report files have been generated within the result folder and copied to your local machine.
```
RandellP: What do you run?

NEW_README.md
billclaytor: Mainly it has to contain data / can't be empty. During my initial testing I had empty log files which caused errors. Once I updated `gatling.data.file.bufferSize` to ensure that data was written during a short test run I had no further issues generating reports from aborted runs. It may be possible that if Gatling is killed while it is writing to the file that it could contain data but still be invalid, but I wasn't able to reproduce this.

NEW_README.md
billclaytor: Will do.

PR #227, opened 2018-05-03
(SLV-112) update rakefile to use abs token

billclaytor:
ABS now requires a token when making a request. The most common ways to store a token locally are in an environment variable or the user's .fog file.

`retry_abs_request` has been updated to call `get_abs_token` which first checks the ABS_TOKEN environment variable, then checks the user's .fog file for the `abs_token` parameter. If not found the test will abort.

### DIFF
--- a/NEW_README.md
+++ b/NEW_README.md
@@ -6,8 +6,15 @@ At the end of the run, the gatling results and atop results will be copied back 
 
 * Clone the gatling-puppet-load-test repo locally: https://github.com/puppetlabs/gatling-puppet-load-test
 * cd into the gatling-puppet-load-test root directory
-* For apples to apples runs, you can use the checked-in hosts files: pe-perf-test.cfg or foss-perf-test.cfg (default for the performance rake task).
-* In order for us to take advantage of the new AWS account with access to internal network resources, you need to use ABS. The 'performance' task will automatically use ABS to provision 2 AWS instances and then execute tests against those instances.
+* For apples to apples runs, you can use the checked-in hosts files: [pe-perf-test.cfg](config/pe-perf-test.cfg) or [foss-perf-test.cfg](/config/foss-perf-test.cfg) (default for the performance rake task).
+* In order for us to take advantage of the new AWS account with access to internal network resources, you need to use [ABS](https://github.com/puppetlabs/always-be-scheduling). The 'performance' task will automatically use ABS to provision 2 AWS instances and then execute tests against those instances.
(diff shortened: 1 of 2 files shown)


rakefile
```diff
@@ -24,8 +25,28 @@ rototiller_task :performance do
           ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
 end
 
+def get_abs_token_from_fog_file
+  home = ENV['HOME']
+  fog_path = "#{home}/.fog"
+  if File.exist?(fog_path)
+    fog = YAML.load(File.read(fog_path))
+  else
+    abort ".fog file not found in home directory: #{fog_path}"
```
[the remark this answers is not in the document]
billclaytor: I updated this to return nil rather than aborting if the file or token aren't found.

rakefile
```diff
@@ -24,8 +25,28 @@ rototiller_task :performance do
...
+
+  token = fog[:default][:abs_token]
+  if !token
+    abort "ABS token not found in .fog file at #{fog_path}"
+  end
+  token
+end
+
+def get_abs_token
+  ENV['ABS_TOKEN'] || get_abs_token_from_fog_file
```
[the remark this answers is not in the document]
billclaytor: I moved the error handling here per your suggestion; thanks.

NEW_README.md
```diff
@@ -6,8 +6,15 @@ At the end of the run, the gatling results and atop results will be copied back
 
 * Clone the gatling-puppet-load-test repo locally: https://github.com/puppetlabs/gatling-puppet-load-test
 * cd into the gatling-puppet-load-test root directory
-* For apples to apples runs, you can use the checked-in hosts files: pe-perf-test.cfg or foss-perf-test.cfg (default for the performance rake task).
-* In order for us to take advantage of the new AWS account with access to internal network resources, you need to use ABS. The 'performance' task will automatically use ABS to provision 2 AWS instances and then execute tests against those instances.
+* For apples to apples runs, you can use the checked-in hosts files: [pe-perf-test.cfg](config/pe-perf-test.cfg) or [foss-perf-test.cfg](/config/foss-perf-test.cfg) (default for the performance rake task).
```
RandellP: The part about (default for the performance rake task) doesn't work out.  There isn't one default (it wasn't clear which it was saying was the default anyway).  The default is picked based on BEAKER_INSTALL_TYPE.  You didn't technically add the default part... :)  But you happened to be editing the same line. :)

billclaytor: All I changed here was to add markdown to make pe-perf-test.cfg or foss-perf-test.cfg links to the files.

RandellP: Even better. :)

PR #229, opened 2018-05-04
(SLV-77) Add report-utils

billclaytor:
This update adds the extract-csv.rb and compare-results.rb utilities used to extract result data from the Gatling HTML report (actually the JSON files used by the HTML report) and compare the data from two runs. This is a work in progress but I wanted to make these available for other team members to use.

Please see the README.md file for more details and instructions for use.

### DIFF
--- a/report-utils/README.md
+++ b/report-utils/README.md
@@ -1,0 +1,87 @@
+## Report Utilities
+
+Gatling produces a detailed HTML report from the simulation.log file. 
+However, our Apples to Apples test reports include a summarized comparison of the data from two test runs.
+Producing this report currently requires manually copying the data from each HTML report into the corresponding table in a Google Sheets doc.
+
+These utility scripts have been created to assist in that process. They are a work-in-progress. Some manual steps are still required, but this gets us closer to a fully automated process.
+
+#### extract-csv.rb
+
+This script parses the JSON results file for a specified test run and creates a CSV file with the data used in our Apples to Apples test report.
+
+How to use it:
+
+In this example the result folder for our performance run will be `gatling-puppet-load-test/PERF_1524862584`. 
+`PERF_1524862584` contains the following folders:
+
+* `ip-10-227-0-132.amz-dev.puppet.net`
+* `ip-10-227-3-213.amz-dev.puppet.net`
+
+If you know which is your metrics node, start there; otherwise, it will be the one containing a `root` folder.
+In our example the metrics node is `ip-10-227-0-132.amz-dev.puppet.net`. You'll need to provide the script with the full path to the result folder, for example:
+
+```
+bill.claytor:~> cd /Users/bill.claytor/RubymineProjects/gatling-puppet-load-test/report-utils 
+bill.claytor:~/RubymineProjects/gatling-puppet-load-test/report-utils> ls
+README.md		compare-results.rb	extract-csv.rb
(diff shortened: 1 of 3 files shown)


report-utils/extract-csv.rb
```diff
@@ -0,0 +1,46 @@
...
+STD = 'standardDeviation'
+TOTAL = 'total'
+
+# totals row is in the 'stats' node
+totals = group_node['stats']
+
+# transaction rows are in the 'contents' node
+contents = group_node['contents']
+
+catalog = contents[contents.keys[4]]['stats']
```
RandellP: why use the number, and not the same select method as above to find the right key?

report-utils/compare-results.rb
```diff
@@ -0,0 +1,32 @@
+require 'csv'
+
+def diff(a, b)
+  result = ( (b.to_f - a.to_f) / a.to_f ) * 100
+  "#{result.round(2).to_s}%"
+end
+
+raise Exception, 'you must provide two csv files to compare' unless ARGV[1]
+
+result_a_name = ARGV[0].split('/')[ARGV[0].split('/').length - 1].gsub('.csv','')
```
RandellP: you can use array[-1] to get the last element of the array.

report-utils/extract-csv.rb
```diff
@@ -0,0 +1,46 @@
...
+filemeta_mco_plugins = contents[contents.keys[5]]['stats']
+filemeta_plugins = contents[contents.keys[2]]['stats']
+filemeta_pluginfacts = contents[contents.keys[1]]['stats']
+locales = contents[contents.keys[3]]['stats']
+node = contents[contents.keys[0]]['stats']
+report = contents[contents.keys[6]]['stats']
+
+CSV.open("./#{result_folder_name}.csv", "wb") do |csv|
+  csv << ["Duration","max ms","mean ms","std dev"]
```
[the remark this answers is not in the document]
billclaytor: The table format and headings were taken from the 2018.1.0 (Irving) Performance and Scalability Test Results. However, when copying the data from the CSV to paste into the report you don't need to copy the column headings, just the data.

report-utils/extract-csv.rb
RandellP: ok, I see that now.  That json is a pain to read.  you would have to dig deeper into the hash to find the difference ,which can probably be done, but isn't worth the complexity.   Maybe toss in a comment about that.

report-utils/README.md
```diff
@@ -0,0 +1,87 @@
...
+bill.claytor:~/RubymineProjects/gatling-puppet-load-test/report-utils> ls
+PerfTestLarge-1524848074511.csv	README.md			compare-results.rb		extract-csv.rb
+bill.claytor:~/RubymineProjects/gatling-puppet-load-test/report-utils> 
+
+```
+
+If the script succeeds, a CSV file with the name of the performance run will be created. 
+In this example the file is 'PerfTestLarge-1524848074511.csv'. 
+
+#### compare-results.rb
```
samwoods1: We should see if we can use the bigquery stuff that @pinkypie added to store results and do comparisons rather than using the csv files. That way we can easily compare to a known baseline build instead of a csv we have locally.

billclaytor: I definitely agree! This was just a quick attempt to provide a semi-automated way to produce the current report format until we get to that point.

billclaytor:
Closing as this is no longer needed.

PR #230, opened 2018-05-11
(SLV-122) Add abs_helper and spec, update rakefile to use awsdirect endpoints

billclaytor:
ABS has been updated with the new endpoints /awsdirect and /awsdirectreturn to provide a direct response for AWS host requests:
https://github.com/puppetlabs/always-be-scheduling#apiv2awsdirect

This update adds an abs_helper module to extract and encapsulate the ABS-related code used by the performance rake tasks so that it can be tested via RSpec. A corresponding abs_helper_spec file has been added with examples for critical functionality.

The performance rake tasks (`performance`, `performance_provision_with_abs`, `performance_deprovision_with_abs`) have been updated to use the new endpoints via the abs_helper module. Obsolete code and dependencies have been removed.

Requests to the prod endpoint (https://cinext-abs.delivery.puppetlabs.net/api/v2/awsdirect) currently return a 504 Gateway Timeout error as the updated ABS code hasn't been deployed there yet. The test endpoint (https://cinext-abs-test.delivery.puppetlabs.net/api/v2/awsdirect) responds as specified in the README linked above without actually provisioning the hosts. 

The test endpoint is currently specified in abs_helper via the ABS_BASE_URL constant to facilitate testing and review. Performance runs will fail as Beaker will not be able to connect to the non-existent hosts. However, the `performance_provision_with_abs` and `performance_deprovision_with_abs` tasks can be run independently to test the abs_helper updates:
* Running the `performance_provision_with_abs` task will request two hosts (mom and metrics) from the /awsdirect endpoint and populate the 'last_abs_resource_hosts.log' file with the returned hosts. 

* Running the `performance_deprovision_with_abs` task (with no ABS_RESOURCE_HOSTS environment variable) will return the hosts captured in the 'last_abs_resource_hosts.log' file.

Once the prod endpoint is ready I'll update abs_helper to use it and perform a final round of testing with a full performance run.

### DIFF
--- a/rakefile
+++ b/rakefile
@@ -47,7 +47,7 @@ end
 # TODO: replace original version and remove '_awsdirect'
 desc 'Task to deprovision systems that were provisioned with ABS - included in the performance task following the rules of BEAKER_PRESERVE_HOSTS env var.'
 rototiller_task :performance_deprovision_with_abs_awsdirect do
-  abs_resource_hosts = return_abs_resource_hosts
+  abs_resource_hosts = abs_return_resource_hosts
   raise 'Failed to deprovision ABS hosts' unless abs_resource_hosts
 end
 
--- a/rakefile
+++ b/rakefile
@@ -1,6 +1,5 @@
 require 'rototiller'
 require 'rspec/core/rake_task'
-require 'net/http'
 
 require './setup/helpers/abs_helper'
 include AbsHelper
--- a/rakefile
+++ b/rakefile
@@ -5,9 +5,53 @@ require 'timeout'
 require 'json'
 require 'yaml'
 
+require './setup/helpers/abs_helper'
+include AbsHelper
+
 BASE_ABS_URI = 'https://cinext-abs.delivery.puppetlabs.net/api/v2/'
 
 task :default => :performance
+
+# temporary alternate versions of the performace tasks that use awsdirect
+
+# TODO: replace original version and remove '_awsdirect'
+desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
+rototiller_task :performance_awsdirect do
+  if ENV['PUPPET_GATLING_REPORTS_ONLY'] == 'true' && (!ENV['PUPPET_GATLING_REPORTS_TARGET'] || ENV['PUPPET_GATLING_REPORTS_TARGET'] == '')
+    abort 'A valid result folder (i.e. PerfTestLarge-1524848074511) must be specified for PUPPET_GATLING_REPORTS_TARGET if PUPPET_GATLING_REPORTS_ONLY=true'
+  end
+
+  Rake::Task["performance_provision_with_abs_awsdirect"].execute unless ENV['ABS_RESOURCE_HOSTS'] ||
+      (ENV['BEAKER_HOSTS'] &&
+          ENV['BEAKER_HOSTS'] != 'config/beaker_hosts/pe-perf-test.cfg' &&
+          ENV['BEAKER_HOSTS'] != 'config/beaker_hosts/foss-perf-test.cfg')
+  Rake::Task["performance_without_provision"].execute
+  Rake::Task["performance_deprovision_with_abs_awsdirect"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' ||
+      ((@beaker_cmd.nil? || @beaker_cmd.result.exit_code != 0)&&
+          ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
+end
+
+# TODO: replace original version and remove '_awsdirect'
+desc 'Provision systems for the performance task with ABS - included in the performance task'
+rototiller_task :performance_provision_with_abs_awsdirect do |t|
+  raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
+      ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE'] == 'foss'
+  t.add_env({:name => 'BEAKER_PE_DIR', :message => 'The PE download directory, example: http://enterprise.delivery.puppetlabs.net/archives/releases/2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
+  t.add_env({:name => 'BEAKER_PE_VER', :message => 'The PE version to install on hosts, example: 2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
+  t.add_env({:name => 'ABS_OS', :default => 'centos-7-x86-64-west', :message => 'The OS to provision with ABS: centos-7-x86-64-west (for AWS), or centos-7-x86_64 (for vmpooler)'})
+
+  abs_resource_hosts = abs_get_resource_hosts
+  raise 'Unable to provision hosts via ABS' unless abs_resource_hosts
+end
+
+# TODO: replace original version and remove '_awsdirect'
+desc 'Task to deprovision systems that were provisioned with ABS - included in the performance task following the rules of BEAKER_PRESERVE_HOSTS env var.'
+rototiller_task :performance_deprovision_with_abs_awsdirect do
+  abs_resource_hosts = return_abs_resource_hosts
+  raise 'Failed to deprovision ABS hosts' unless abs_resource_hosts
+end
+
+#######
 
 desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
 rototiller_task :performance do
@@ -25,60 +69,61 @@ rototiller_task :performance do
           ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
 end
 
-def get_abs_token_from_fog_file
-  home = ENV['HOME']
-  fog_path = "#{home}/.fog"
-  token = nil
-
-  if File.exist?(fog_path)
-    fog = YAML.load(File.read(fog_path))
-    token = fog[:default][:abs_token]
-  else
-    puts ".fog file not found in home directory: #{fog_path}"
-  end
-
-  if !token
-    puts "ABS token not found in .fog file at #{fog_path}"
-  end
-  token
-end
-
-def get_abs_token
-  ENV['ABS_TOKEN'] ? token = ENV['ABS_TOKEN'] : token = get_abs_token_from_fog_file
-  if !token
-    abort "An ABS token must be set in either the ABS_TOKEN environment variable or the abs_token parameter in the .fog file"
-  end
-  token
-end
-
-def retry_abs_request(uri, body, timeout, time_increment, invalid_response_bodies = [], valid_response_codes = [202, 200])
-  req = Net::HTTP::Post.new(uri, {'Content-Type' => 'application/json', 'X-Auth-Token' => get_abs_token})
-  req.body = body
-  http = Net::HTTP.new(uri.host, uri.port)
-  http.use_ssl = true
-
-  puts "sending request body: #{body}"
-  puts "to uri: #{uri}"
-
-  res = nil
-  Timeout::timeout(timeout) do
-    while res.nil? || invalid_response_bodies.include?(res.body) || !valid_response_codes.include?(res.code)
-      begin
-        res = http.request(req)
-      rescue => ex
-        puts ex
-        puts ex.backtrace
-      end
-      if res.nil? || invalid_response_bodies.include?(res.body)
-        puts "#{Time.now} - Retrying ABS request..."
-        sleep(time_increment)
-      else
-        break
-      end
-    end
-  end
-  res
-end
+# TODO: remove; these have been extracted to abs_helper
+# def get_abs_token_from_fog_file
+#   home = ENV['HOME']
+#   fog_path = "#{home}/.fog"
+#   token = nil
+#
+#   if File.exist?(fog_path)
+#     fog = YAML.load(File.read(fog_path))
+#     token = fog[:default][:abs_token]
+#   else
+#     puts ".fog file not found in home directory: #{fog_path}"
+#   end
+#
+#   if !token
+#     puts "ABS token not found in .fog file at #{fog_path}"
+#   end
+#   token
+# end
+#
+# def get_abs_token
+#   ENV['ABS_TOKEN'] ? token = ENV['ABS_TOKEN'] : token = get_abs_token_from_fog_file
+#   if !token
+#     abort "An ABS token must be set in either the ABS_TOKEN environment variable or the abs_token parameter in the .fog file"
+#   end
+#   token
+# end
+#
+# def retry_abs_request(uri, body, timeout, time_increment, invalid_response_bodies = [], valid_response_codes = [202, 200])
+#   req = Net::HTTP::Post.new(uri, {'Content-Type' => 'application/json', 'X-Auth-Token' => get_abs_token})
+#   req.body = body
+#   http = Net::HTTP.new(uri.host, uri.port)
+#   http.use_ssl = true
+#
+#   puts "sending request body: #{body}"
+#   puts "to uri: #{uri}"
+#
+#   res = nil
+#   Timeout::timeout(timeout) do
+#     while res.nil? || invalid_response_bodies.include?(res.body) || !valid_response_codes.include?(res.code)
+#       begin
+#         res = http.request(req)
+#       rescue => ex
+#         puts ex
+#         puts ex.backtrace
+#       end
+#       if res.nil? || invalid_response_bodies.include?(res.body)
+#         puts "#{Time.now} - Retrying ABS request..."
+#         sleep(time_increment)
+#       else
+#         break
+#       end
+#     end
+#   end
+#   res
+# end
 
 desc 'Provision systems for the performance task with ABS - included in the performance task'
 rototiller_task :performance_provision_with_abs do |t|
@@ -119,6 +164,7 @@ rototiller_task :performance_deprovision_with_abs do
   retry_abs_request(uri, req_body, 30, 5)
 end
 
+desc 'Run Performance tests using previously provisioned hosts'
 rototiller_task :performance_against_already_provisioned do
   ENV['BEAKER_HOSTS'] ||= "log/latest/hosts_preserved.yml"
   ENV['ABS_RESOURCE_HOSTS'] = File.open("last_abs_resource_hosts.log").readlines[0] if File.exist? "last_abs_resource_hosts.log"
--- a/rakefile
+++ b/rakefile
@@ -1,9 +1,6 @@
 require 'rototiller'
 require 'rspec/core/rake_task'
 require 'net/http'
-require 'timeout'
-require 'json'
-require 'yaml'
 
 require './setup/helpers/abs_helper'
 include AbsHelper
--- a/rakefile
+++ b/rakefile
@@ -30,16 +30,16 @@ rototiller_task :performance_provision_with_abs do |t|
   t.add_env({:name => 'BEAKER_PE_VER', :message => 'The PE version to install on hosts, example: 2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
   t.add_env({:name => 'ABS_OS', :default => 'centos-7-x86-64-west', :message => 'The OS to provision with ABS: centos-7-x86-64-west (for AWS), or centos-7-x86_64 (for vmpooler)'})
 
-  abs_resource_hosts = abs_get_resource_hosts(abs_get_a2a_hosts)
+  abs_resource_hosts = get_abs_resource_hosts(get_a2a_hosts)
   raise 'Unable to provision hosts via ABS' unless abs_resource_hosts
 end
 
 desc 'Task to de-provision systems that were provisioned with ABS - included in the performance task following the rules of BEAKER_PRESERVE_HOSTS env var.'
 rototiller_task :performance_deprovision_with_abs do
-  abs_resource_hosts = ENV['ABS_RESOURCE_HOSTS'].nil? ? abs_get_last_abs_resource_hosts : ENV['ABS_RESOURCE_HOSTS']
+  abs_resource_hosts = ENV['ABS_RESOURCE_HOSTS'].nil? ? get_last_abs_resource_hosts : ENV['ABS_RESOURCE_HOSTS']
   raise 'This task requires an array of hostnames to be specified via the ABS_RESOURCE_HOSTS environment variable or the last_abs_resource_hosts.log file' unless abs_resource_hosts
 
-  returned_hosts = abs_return_resource_hosts(abs_resource_hosts)
+  returned_hosts = return_abs_resource_hosts(abs_resource_hosts)
   raise 'Failed to de-provision ABS hosts' unless returned_hosts == abs_resource_hosts
 end
 
--- a/rakefile
+++ b/rakefile
@@ -5,8 +5,6 @@ require 'net/http'
 require './setup/helpers/abs_helper'
 include AbsHelper
 
-BASE_ABS_URI = 'https://cinext-abs.delivery.puppetlabs.net/api/v2/'
-
 task :default => :performance
 
 desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
@@ -33,7 +31,7 @@ rototiller_task :performance_provision_with_abs do |t|
   t.add_env({:name => 'BEAKER_PE_VER', :message => 'The PE version to install on hosts, example: 2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
   t.add_env({:name => 'ABS_OS', :default => 'centos-7-x86-64-west', :message => 'The OS to provision with ABS: centos-7-x86-64-west (for AWS), or centos-7-x86_64 (for vmpooler)'})
 
-  abs_resource_hosts = abs_get_resource_hosts
+  abs_resource_hosts = abs_get_resource_hosts(abs_get_a2a_hosts)
   raise 'Unable to provision hosts via ABS' unless abs_resource_hosts
 end
 
@@ -42,8 +40,8 @@ rototiller_task :performance_deprovision_with_abs do
   abs_resource_hosts = ENV['ABS_RESOURCE_HOSTS'].nil? ? abs_get_last_abs_resource_hosts : ENV['ABS_RESOURCE_HOSTS']
   raise 'This task requires an array of hostnames to be specified via the ABS_RESOURCE_HOSTS environment variable or the last_abs_resource_hosts.log file' unless abs_resource_hosts
 
-  abs_returned_hosts = abs_return_resource_hosts(abs_resource_hosts)
-  raise 'Failed to de-provision ABS hosts' unless abs_returned_hosts
+  returned_hosts = abs_return_resource_hosts(abs_resource_hosts)
+  raise 'Failed to de-provision ABS hosts' unless returned_hosts == abs_resource_hosts
 end
 
 desc 'Run Performance tests using previously provisioned hosts'
--- a/rakefile
+++ b/rakefile
@@ -12,47 +12,6 @@ BASE_ABS_URI = 'https://cinext-abs.delivery.puppetlabs.net/api/v2/'
 
 task :default => :performance
 
-# temporary alternate versions of the performace tasks that use awsdirect
-
-# TODO: replace original version and remove '_awsdirect'
-desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
-rototiller_task :performance_awsdirect do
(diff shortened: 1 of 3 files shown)


RandellP:
The diffs feel weird.  I see code that is noted as added that I thought was added in a previous PR.    I also see code that is being removed saying it was moved to the helper, but I don't see any helper changes...

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+
+  ABS_BEAKER_TYPE = 'centos-7-x86-64-west'
+  ABS_BEAKER_ENGINE = 'aws'
+
+  # TODO: replace when ready
+  # ABS_AWS_REAP_TIME = '86400'
+  ABS_AWS_REAP_TIME = '1200'
+
+  def abs_get_base_url
+    ENV['ABS_BASE_URL'] = ABS_BASE_URL unless ENV['ABS_BASE_URL']
```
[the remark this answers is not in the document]
samwoods1: Ok, so after taking a look at the entire file, I think there's a simpler way to do this without needing all of the abs_get_* methods. I'm leaving the additional comments I already made because I think they're useful to learn from, but you don't need to do both things.
Typically what I would do with Ruby here is pretty simple. You can create a constructor that gets all of the final values for you, then just access them as properties. So... like this:
```
def initialize
  @abs_base_url = ENV['ABS_BASE_URL'] ? ENV['ABS_BAS_URL'] : ABS_BASE_URL
  ...
end
```

rakefile
```diff
@@ -47,7 +47,7 @@ end
 # TODO: replace original version and remove '_awsdirect'
```
samwoods1: I think you don't need this todo anymore?

billclaytor: I was thinking that when we're no longer using the old endpoint these new versions would replace the current versions; no reason to keep the '_awsdirect' designation at that point:

performance_awsdirect >> performance
performance_provision_with_abs_awsdirect >> performance_provision_with_abs
performance_deprovision_with_abs_awsdirect >> performance_deprovision_with_abs

If we do want to keep the current versions and have the option of using either endpoint I'll remove the todos.

setup/helpers/abs_helper.rb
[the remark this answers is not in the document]
billclaytor: I've updated abs_helper to use this style, thanks!

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
+require 'net/http'
+require 'timeout'
+require 'json'
+require 'yaml'
+
+module AbsHelper
+
+  # TODO: replace when ready
+  #ABS_BASE_URL = 'https://cinext-abs.delivery.puppetlabs.net/api/v2'
```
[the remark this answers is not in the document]
billclaytor: I get a 504 gateway timeout when making a valid request to the abs endpoint. However, the update seems to be there because I get valid errors if parameters are missing. I have also gotten a 504 from the abs-test endpoint but most of the time it returns the expected response.

samwoods1:
@billclaytor It looks like there are a few unit test failures in the distelli run.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+  end
+
+  describe '#abs_get_resource_hosts' do
+
+    context 'when a valid response is returned' do
+
+      it 'sets the environment variable, logs the hosts, and returns the response(s)' do
+        ENV['ABS_TOKEN'] = TEST_ABS_TOKEN
+        expect(subject).to receive(:abs_request_awsdirect).at_least(:once).and_return(test_http_response)
```
RandellP: Looks like abs_get_awsdirect_request_body should be mocked as well

RandellP: and maybe this should check that the returned body from the mock gets passed to abs_request_awsdirect

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
```
RandellP: abs_is_valid_response should also be mocked to avoid a break in it breaking this test.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+        expect(test_http_response).to receive(:nil?).at_least(:once).and_return(false)
+        expect(test_http_response).to receive(:code).at_least(:once).and_return(TEST_VALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).at_least(:once).and_return(TEST_AWSDIRECT_RESPONSE_BODY)
```
RandellP: All three of these don't actually happen in the method under test.  So if the stuff mentioned above gets mocked, these all go away.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+    context 'when an invalid response code is returned' do
+
+      it 'reports the error and returns nil' do
+        ENV['ABS_TOKEN'] = TEST_ABS_TOKEN
+        test_expected_result = nil
+
+        expect(subject).to receive(:abs_request_awsdirect).at_least(:once).and_return(test_http_response)
+
+        expect(test_http_response).to receive(:nil?).at_least(:once).and_return(false)
```
RandellP: Same comment as previous test.  If abs_is_valid_response is mocked, don't need to mock these three items, and you don't need to test a bad code versus a bad body... as this method actually doesn't care why it is bad.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+        expect(test_http_response).to receive(:code).at_least(:once).and_return(TEST_INVALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).at_least(:once).and_return(TEST_INVALID_RESPONSE_BODY)
+
+        expect(subject.abs_get_resource_hosts).to eq(test_expected_result)
+      end
+
+    end
+
+    context 'when an invalid response body is returned' do
```
RandellP: This test can be dropped if the above mocks abs_is_valid_response.  This method doesn't differentiate between bad code and bad body.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+    end
+
+    context 'when an invalid response code is returned' do
+
+      it 'reports the error and returns nil' do
+        ENV['ABS_TOKEN'] = TEST_ABS_TOKEN
+        test_expected_result = nil
+
+        expect(subject).to receive(:abs_request_awsdirect).at_least(:once).and_return(test_http_response)
```
RandellP: This should be exactly once.  as you want it to stop if the first one fails.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+        expect(test_http_response).to receive(:nil?).at_least(:once).and_return(false)
+        expect(test_http_response).to receive(:code).at_least(:once).and_return(TEST_INVALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).at_least(:once).and_return(TEST_INVALID_RESPONSE_BODY)
+
+        expect(subject.abs_get_resource_hosts).to eq(test_expected_result)
```
RandellP: Should verify output messages (test descriptor has "reports error"), and that it returned the hosts by mocking the abs_return_resource_hosts and ensuring it gets called.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+        expect(test_http_response).to receive(:nil?).at_least(:once).and_return(false)
+        expect(test_http_response).to receive(:code).at_least(:once).and_return(TEST_VALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).at_least(:once).and_return(TEST_INVALID_RESPONSE_BODY)
+
+        expect(subject.abs_get_resource_hosts).to eq(test_expected_result)
+      end
+
+    end
+
+    context 'when the response is nil' do
```
RandellP: This test can be dropped if the above mocks abs_is_valid_response. This method doesn't differentiate between bad code and bad body, and nil.

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+        break
+
+      else
+        # reformat to satisfy beaker
+        responses << JSON.parse(abs_reformat_resource_host(response.body))
+      end
+
+    end
+
+    if responses.empty?
```
RandellP: This feels odd.  The only way to get here I think is if the first request returned nil.  Which would be an invalid response.  Feels like this should be if invalid response then print message about any a provisioning error and that any provisions hosts were terminated.  But That messaging is already there I think.  Unless nil isn't considered an invalid response.  So I think this if part can be tossed, and the elseif converted to an if.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+      it 'submits requests to return the hosts and returns the host array' do
+        ENV['ABS_RESOURCE_HOSTS'] = TEST_ABS_RESOURCE_HOSTS
+
+        expect(subject).to receive(:abs_get_base_url).at_least(:once).and_return(TEST_BASE_URL)
+
+        expect(subject).to receive(:abs_get_awsdirectreturn_request_body).with(TEST_HOSTNAME).at_least(:once).and_return(test_body)
+        expect(subject).to receive(:abs_request_awsdirect).with(test_uri, test_body).at_least(:once).and_return(nil)
+
+        expect(subject.abs_return_resource_hosts).to include(TEST_ABS_RESOURCE_HOSTS)
```
RandellP: not sure what this is doing.  (the include).  It could use 'to eq(...)'  But the return from the method is not really important...  Might not want to check the return at all.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+    test_uri = TEST_AWSDIRECTRETURN_URI
+    test_body = TEST_AWSDIRECTRETURN_REQUEST_BODY
+
+    context 'when an array of hosts is specified via ABS_RESOURCE_HOSTS' do
+
+      it 'submits requests to return the hosts and returns the host array' do
+        ENV['ABS_RESOURCE_HOSTS'] = TEST_ABS_RESOURCE_HOSTS
+
+        expect(subject).to receive(:abs_get_base_url).at_least(:once).and_return(TEST_BASE_URL)
```
RandellP: might want to spec that this should be called twice, to make sure it does both hosts

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+        expect(subject).to receive(:abs_get_awsdirectreturn_request_body).with(TEST_HOSTNAME).at_least(:once).and_return(test_body)
+        expect(subject).to receive(:abs_request_awsdirect).with(test_uri, test_body).at_least(:once).and_return(nil)
+
+        expect(subject.abs_return_resource_hosts).to include(TEST_ABS_RESOURCE_HOSTS)
+      end
+
+    end
+
+    context 'when an array of hosts is specified via last_abs_resource_hosts.log' do
```
RandellP: same comments as test above.

rakefile
```diff
@@ -5,81 +5,126 @@ require 'timeout'
...
 
+require './setup/helpers/abs_helper'
+include AbsHelper
+
 BASE_ABS_URI = 'https://cinext-abs.delivery.puppetlabs.net/api/v2/'
 
 task :default => :performance
 
+# temporary alternate versions of the performace tasks that use awsdirect
+
```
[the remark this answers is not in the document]
billclaytor: This was just to facilitate testing and comparison between the current and updated tasks. When we're no longer using the old endpoint these new versions will replace the current versions:
  performance_awsdirect >> performance performance_provision_with_abs_awsdirect >> performance_provision_with_abs performance_deprovision_with_abs_awsdirect >> performance_deprovision_with_abs

rakefile
```diff
@@ -5,81 +5,126 @@ require 'timeout'
 require 'json'
 require 'yaml'
```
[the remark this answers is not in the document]
billclaytor: No, these would be removed:
' json' is still being used by code that would be replaced; I'll add a todo.
 code that requires 'yaml' has already been extracted; I'll remove this.

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+    if File.exist?(fog_path)
+      fog = YAML.load(File.read(fog_path))
+      token = fog[:default][:abs_token]
+    else
+      puts ".fog file not found in home directory: #{fog_path}"
+    end
+
+    if !token
+      puts "ABS token not found in .fog file at #{fog_path}"
+    end
```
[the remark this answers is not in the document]
billclaytor: Will fix

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+    {'mom': abs_get_aws_mom_size, 'metrics': abs_get_aws_metrics_size}
+  end
+
+  def abs_reformat_resource_host(response_body)
+    reformatted_json = nil
+
+    begin
+      json = JSON.parse(response_body)
+      hostname = json['hostname']
+      type = json['type']
```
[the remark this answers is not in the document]
billclaytor: Will fix

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+
+    if responses.empty?
+      puts 'No ABS hosts were provisioned'
+      puts ''
+    elsif !invalid_response
+      abs_resource_hosts = responses.to_json
+      ENV['ABS_RESOURCE_HOSTS'] = abs_resource_hosts
+      puts "ABS_RESOURCE_HOSTS=#{ENV['ABS_RESOURCE_HOSTS']}"
+
+      # TODO: extract
```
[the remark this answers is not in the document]
billclaytor: I meant extract the code that updates the file into a separate method; the method below reads the file. I've added abs_update_last_abs_resource_hosts.

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+  def abs_get_last_abs_resource_hosts
+    last_abs_resource_hosts = File.open("last_abs_resource_hosts.log").readlines[0]
+    last_abs_resource_hosts
+  end
+
+  def abs_return_resource_hosts
+    uri = URI("#{abs_get_base_url}/awsdirectreturn")
+    abs_resource_hosts = ENV['ABS_RESOURCE_HOSTS'].nil? ? abs_get_last_abs_resource_hosts : ENV['ABS_RESOURCE_HOSTS']
+
+    puts "ABS hosts: #{abs_resource_hosts}"
```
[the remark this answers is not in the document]
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+
+        it 'returns nil' do
+          fog_file =
+              ':default:
+              :not_abs_token: xyz123'
+
+          expect(File).to receive(:exist?).and_return(true)
+          expect(File).to receive(:read).and_return(fog_file)
+
+          expect(subject.abs_get_token_from_fog_file).to eq(nil)
```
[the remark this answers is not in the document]
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+      end
+
+    end
+
+    context 'when the .fog file does not exist' do
+
+      it 'reports the error and returns nil' do
+        expect(File).to receive(:exist?).and_return(false)
+        expect(subject).to receive(:puts).at_least(:once)
+
```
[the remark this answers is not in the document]
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+      end
+
+      context 'when a token is not set in the .fog file' do
+
+        it 'reports the error and returns nil' do
+
+          ENV['ABS_TOKEN'] = nil
+          expect(subject).to receive(:abs_get_token_from_fog_file).and_return(nil)
+          expect(subject).to receive(:puts)
```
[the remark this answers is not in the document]
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+    end
+
+    context 'when the response code is not valid' do
+
+      it 'returns false' do
+        expect(test_http_response).to receive(:nil?).at_least(:once).and_return(false)
+        expect(test_http_response).to receive(:code).at_least(:once).and_return(TEST_INVALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).at_least(:once).and_return(TEST_AWSDIRECT_RESPONSE_BODY)
+
+        expect(subject.abs_is_valid_response?(test_http_response)).to eq(false)
```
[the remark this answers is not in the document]
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+    end
+
+    context 'when the response body is empty' do
+
+      it 'returns false' do
+        expect(test_http_response).to receive(:nil?).at_least(:once).and_return(false)
+        expect(test_http_response).to receive(:code).at_least(:once).and_return(TEST_VALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).at_least(:once).and_return(TEST_INVALID_RESPONSE_BODY)
+
+        expect(subject.abs_is_valid_response?(test_http_response)).to eq(false)
```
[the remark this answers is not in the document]
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Will fix

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Will update to use to eq; not sure why I was using include here but perhaps for something that was subsequently updated.

spec/setup/helpers/abs_helper_spec.rb
billclaytor: This should work for any number of hosts.

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,284 @@
...
+  end
+
+  def abs_get_last_abs_resource_hosts
+    last_abs_resource_hosts = File.open("last_abs_resource_hosts.log").readlines[0]
+    last_abs_resource_hosts
+  end
+
+  def abs_return_resource_hosts
+    uri = URI("#{abs_get_base_url}/awsdirectreturn")
+    abs_resource_hosts = ENV['ABS_RESOURCE_HOSTS'].nil? ? abs_get_last_abs_resource_hosts : ENV['ABS_RESOURCE_HOSTS']
```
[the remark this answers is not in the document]
billclaytor: Fixed

rakefile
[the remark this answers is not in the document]
billclaytor: Fixed

setup/helpers/abs_helper.rb
billclaytor: Actually, not sure my comment addressed your point. Looking into it...

spec/setup/helpers/abs_helper_spec.rb
billclaytor: This should be resolved. The method now reports the code and body when the response is not nill, and the spec tests each scenario.

setup/helpers/abs_helper.rb
billclaytor: Looking at this again, the responses array could be empty if either no requests were made (no hosts specified or no token) or if the first response was invalid. If the responses array is empty we don't need to proceed further since no hosts were provisioned. And we don't want to check whether invalid_response is true independently; it may actually be false if no request was made. 

Should this be refactored?

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,779 @@
...
+  describe '#abs_request_awsdirect' do
+
+    context 'when a request is successfully prepared' do
+
+      it 'performs the request, reports and returns the response' do
+        expect(subject).to receive(:abs_get_request_post).and_return(test_http_request)
+        allow(test_http_request).to receive(:body=)
+
+        stub_const('Net::HTTP', test_net_http)
+        allow(test_net_http).to receive(:new).and_return(test_net_http_instance)
```
RandellP: on line 356 you did basically the same thing, but used expect, whereas here you used allow. Was that intentional?

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,779 @@
...
+
+        allow(test_net_http_instance).to receive(:use_ssl=)
+        allow(test_net_http_instance).to receive(:read_timeout=)
+        expect(test_net_http_instance).to receive(:request).with(test_http_request).and_return(test_http_response)
+
+        expect(test_http_response).to receive(:code).and_return(TEST_VALID_RESPONSE_CODE)
+        expect(test_http_response).to receive(:body).and_return(TEST_VALID_RESPONSE_BODY)
+
+        allow(subject).to receive(:puts)
```
RandellP: What is this for?  Would it fail if you didn't have this?  Seems like if the puts got removed it wouldn't fail since it is only and allow.

spec/setup/helpers/abs_helper_spec.rb
RandellP: I think this got fixed, and somehow the comment remained, as it doesn't seem to apply to the code that it is attached to anymore... (mocking abs_is_valid_response when testing it  wouldn't make any sense... )

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -0,0 +1,779 @@
...
+
+    context 'when a valid response_body is provided' do
+
+      it 'returns the expected reformatted response' do
+        expect(subject.abs_reformat_resource_host(TEST_AWSDIRECT_RESPONSE_BODY)).to eq(TEST_REFORMATTED_RESPONSE_BODY)
+      end
+
+    end
+
+    context 'when an invalid response_body is provided' do
```
RandellP: Looking at this test.  It doesn't feel like enough information is getting to the user to tell them what went wrong.  all they know is there was a json parse error somewhere...
And the method goes on to return nil.  The caller doesn't check it, just adds it to the list of responses, and appears to consider the nil the same way it does a valid conversion.

spec/setup/helpers/abs_helper_spec.rb
billclaytor: There are cases where abs_helper adds an extra puts after messages to make the output easier to read. I wasn't sure if it made sense to explicitly test for that so I'm just allowing it. If we should test all output I can update the tests for that.

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Good catch; I've updated this to use expect.

setup/helpers/abs_helper.rb
billclaytor: I've updated this to check for a token via abs_initialize before proceeding and moved the empty responses message to where it seems to make more sense.

spec/setup/helpers/abs_helper_spec.rb
billclaytor: I've updated this to provide more information.

spec/setup/helpers/abs_helper_spec.rb
samwoods1: This is probably enough of a test, I don't see a reason we would need to go down that rabbit hole.

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+    invalid_response = false
+
+    puts
+    puts "Attempting to provision ABS hosts: #{hosts_to_request}"
+    puts
+
+    # ensure the user has a token before proceeding
+    if abs_initialize
+      uri = URI("#{@abs_base_url}/awsdirect")
+      hosts_to_request.each do |host_to_request|
```
samwoods1: We should do this in parallel. Since in-parallel is a dependency for Beaker, you should just be able to:
require in_parallel

include InParallel

#options - 300 second timeout, and true to kill all threads on error in any of the threads
hosts_to_request.each.in_parallel(nil, 300, true) do |host_to_request|

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+        puts "host_to_request: #{host_to_request}"
+
+        role = host_to_request[:role]
```
samwoods1: Why take a hash, convert it to specific vars and pass those to the get_awsdirect_request_body? Can't we simplify by just passing the hash to the method?

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+    success = verify_abs_hosts(hosts) unless abs_resource_hosts.nil?
+    verified_hosts = success ? abs_resource_hosts : nil
+
+    if verified_hosts
+      puts
+      puts "ABS hosts have been successfully provisioned"
+      puts "Pausing to ensure successful configuration"
+
+      # TODO: remove?
+      sleep 30
```
samwoods1: Is this needed? Is there a way to test that configuration is successful instead of just arbitrarily sleeping?

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+      hosts.each do |host|
+        hostname = host["hostname"]
+
+        puts "Returning host: #{hostname}"
+        body = get_awsdirectreturn_request_body(hostname)
+        res = perform_awsdirect_request(uri, body)
+
+        if valid_abs_response?(res)
+          puts "Successfully returned host: #{hostname}"
+          returned_hosts ? returned_hosts << host : returned_hosts = [host]
```
samwoods1: If you initialize 'returned_hosts = []' instead of nil you can avoid this logic and just use 'returned_hosts << host'

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+    has_token = abs_initialize
+    unless has_token
+      puts "Unable to proceed without a valid ABS token"
+      puts
+    end
+
+    if is_valid && has_token
+      uri = URI("#{@abs_base_url}/awsdirectreturn")
+
+      # returned_hosts = []
```
samwoods1: leftover?

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+          puts "Successfully returned host: #{hostname}"
+          returned_hosts ? returned_hosts << host : returned_hosts = [host]
+        else
+          puts "Failed to return host: #{hostname}"
+        end
+      end
+
+    end
+
+    returned_hosts ? returned_hosts.to_json : returned_hosts
```
samwoods1: is this necessary because to_json fails if returned_hosts is nil? We could get rid of this too if we initialized returned_hosts = []

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+
+      new_response_body = {
+        'hostname': hostname,
+        'type':     ABS_BEAKER_TYPE,
+        'engine':   ABS_BEAKER_ENGINE
+      }
+
+      reformatted_json = new_response_body.to_json
+    rescue
+      # TODO: raise?
```
samwoods1: We should rescue JSON::ParseError only if that's what we're looking for here, and I think we should re-raise it. If we can't parse the response body then something else downstream will blow up and we won't have this context.

setup/helpers/abs_helper.rb
```diff
@@ -0,0 +1,657 @@
...
+        hosts = JSON.parse(abs_resource_hosts)
+        host = hosts[0]
+        hostname = host["hostname"]
+        if !hostname.nil? && !hostname.empty?
+          is_valid = true
+        else
+          puts "The specified resource host array is not valid: #{abs_resource_hosts}"
+          puts
+        end
+      rescue
```
samwoods1: should be: 
rescue JSON::ParseError => ex
and we should re-raise so we have this context if it ever fails.

setup/helpers/abs_helper.rb
billclaytor: Removed and updated get_abs_resource_hosts to rescue and re-raise after returning any provisioned hosts.

setup/helpers/abs_helper.rb
billclaytor: Removed

PR #236, opened 2018-08-28
(SLV-203) Update perf_helper to remove private key

billclaytor:
This update removes the keys from perf_helper.rb and consolidates the ssh and git setup into setup_r10k. Rather than creating the id_rsa and config files using hard-coded strings, we now retrieve the files from the following base URL:

http://int-resources.ops.puppetlabs.net/QE%20Shared%20Resources/gatling_test_keys

r10k_deploy currently has code that checks to see if id_rsa exists, and if not attempts to curl id_rsa and id_rsa.pub from the above URL. However, since setup_r10k creates the id_rsa file this code never runs. AWS hosts are not able to access this URL so an attempt would fail. This code has been removed since this is handled by setup_r10k.

r10k_deploy also currently contains a step to create a known_hosts file with the GitHub key to prevent errors when cloning the control repo. This step has been moved to setup_r10k and the known_hosts file is now downloaded with the other files. The only step remaining is "Install and configure r10k"; everything else is handled by setup_r10k.

The files (id_rsa, id_rsa.pub, config, known_hosts) are now downloaded into the local tmp/ssh folder and then transferred to the hosts via scp.

I've also replaced the code (including the keys) in setup/install/00_pre_install/30_r10k_git_setup.rb with a call to setup_r10k. This is what is currently done in setup/install_gatling/00_pre_install/30_r10k_git_setup.rb.

### DIFF
--- a/setup/helpers/perf_helper.rb
+++ b/setup/helpers/perf_helper.rb
@@ -8,6 +8,10 @@ module PerfHelper
   BASE_URL = 'http://builds.puppetlabs.lan'
   R10K_DIR = '/opt/puppetlabs/scratch/perf_testing/r10k'
   R10K_CONFIG_PATH = "#{R10K_DIR}/r10k.yaml"
+  TEST_FILES_URL = "http://int-resources.ops.puppetlabs.net/QE%20Shared%20Resources/gatling_test_keys".freeze
+  TEST_SSH_FILES = ["id_rsa", "id_rsa.pub", "config", "known_hosts"].freeze
+  TEMP_SSH_DIR = "tmp/ssh".freeze
+  ROOT_SSH_DIR = "/root/.ssh".freeze
 
   def set_etc_hosts
     puppet_ip = any_hosts_as?(:loadbalancer) ? loadbalancer.ip : master.ip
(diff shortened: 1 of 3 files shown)


setup/helpers/perf_helper.rb
```diff
@@ -8,6 +8,10 @@ module PerfHelper
   BASE_URL = 'http://builds.puppetlabs.lan'
   R10K_DIR = '/opt/puppetlabs/scratch/perf_testing/r10k'
   R10K_CONFIG_PATH = "#{R10K_DIR}/r10k.yaml"
+  TEST_FILES_URL = "http://int-resources.ops.puppetlabs.net/QE%20Shared%20Resources/gatling_test_keys".freeze
+  TEST_SSH_FILES = ["id_rsa", "id_rsa.pub", "config", "known_hosts"].freeze
```
samwoods1: I don't think we need the public key for anything... we should never be using or storing both the private and public key in the same place.  Having the public key at all here doesn't make sense to me.

samwoods1: The change to the known_hosts file and the config file both seem to be trying to solve the same problem. Can we just use the config file and get rid of the known_hosts? Or is the config for SSH directly and known_hosts only for git clone?

billclaytor: I wasn't sure if the public key was actually needed; it was included in r10_deploy in the "Set up SSH key for github access" step but it looked like that code may not have ever actually run. I'll remove it.

billclaytor: I was thinking the config file was for SSH and known_hosts for git clone. I wanted to try and maintain the existing setup, just with files retrieved from outside the gplt codebase.

MikaelSmith: The private key was the one we want to remove. Is that work still planned?

billclaytor: Yes, there was an issue merging that branch so I've created a new branch and PR: https://github.com/puppetlabs/gatling-puppet-load-test/pull/238

PR #240, opened 2018-09-19
(SLV-224) Enable perf testing against opsworks instance sizes

samwoods1:


rakefile
```diff
@@ -15,13 +15,50 @@ rototiller_task :performance do
...
   Rake::Task["performance_deprovision_with_abs"].execute unless ENV['BEAKER_PRESERVE_HOSTS'] == 'always' ||
       ((@beaker_cmd.nil? || @beaker_cmd.result.exit_code != 0)&&
           ENV['BEAKER_PRESERVE_HOSTS'] == 'onfail')
 end
 
+desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup ' +
+    'and tests for opsworks'
+rototiller_task :opsworks_performance do
+  raise "The test must be Opsworks.rb, or '' if you plan to execute tests later" unless
+    ["", "tests/OpsWorks.rb"].include? ENV["BEAKER_TESTS"]
```
RandellP: Why not do an if not set, set to OpsWorks.  From the name of the task, it's pretty obvious what they want to do...

samwoods1:
@RandellP Yeah, I struggled trying to figure out what to name it, because it actually doesn't mean either of those things. OpsworksLargeCatalogSmall actually means a large opsworks instance, and small catalog. The number of agents also happens to follow the instance size because well, larger instances support more agents. Because we're using small, medium and large to describe instance size and catalog size I tried to be explicit about which thing I was referring to.

config/beaker_hosts/opsworks-perf-test.cfg
```diff
@@ -0,0 +1,55 @@
...
+
+:answers:
+  :console_admin_password: puppetlabs
+:department: eso-dept
+:disable_iptables: false
+:project: scale-testing
+
+:host_tags:
+  :run_type: persistent
+  :sub-proj: hoyt
```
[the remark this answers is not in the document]
samwoods1: Yeah, the problem is that for some reason that was required to be set to something with the old way we used AWS. Not sure if that is still the case now that we're going through ABS. I'll double-check and remove it if not needed.

rakefile
samwoods1: I was thinking that they may not want to run any tests and just set stuff up, but you're right, for that they would set it to an empty string, if it's not set at all, we should set it.

PR #244, opened 2018-09-28
(SLV-249) Update to support puppet 6

RandellP:
rvm interfers with puppet agent on metrics node, so stopped setting it up as we don't need it.
Updated gemfile for dependencies that support the lost cli commands in puppet 6.
like puppet cert.  had to require beaker-pe because beaker-pe-large-environments doesn't currently handle it's dependencies correctly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,12 @@
 source ENV['GEM_SOURCE'] || 'https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/'
 
-gem 'beaker', :git => 'https://github.com/puppetlabs/beaker.git', :branch => 'fix/master/BKR-487_preserved_hosts_yml_fix'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,10 +1,12 @@
 source ENV['GEM_SOURCE'] || 'https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/'
 
-gem 'beaker', :git => 'https://github.com/puppetlabs/beaker.git', :branch => 'fix/master/BKR-487_preserved_hosts_yml_fix'
```
samwoods1: Did the preserved hosts issue for ec2 instances get fixed in core beaker, or will we see this issue again?

RandellP: I thought I remembered this got fix somehow a while back.  I believe it was when we moved to abs.  I should probably test it though.

PR #246, opened 2018-10-05
(SLV-253) Fix CI by fixing avg_dsk -> avg_disk

RandellP:
avg_dsk_write appeared in the assertion code periodically.  But it looks like it should be avg_disk_write in all cases.
So fixed that.


RandellP:
@samwoods1 @billclaytor  I put you both on, but only need one review.  This is pretty straight forward.  I don't think it will make CI green though.  The baseline in big query for avg_disk_write is null in a lot of cases.  So probably have to update the baseline after this PR...

samwoods1:
@RandellP Yeah sorry, I made that change to the schema as part of my stuff I'm working on, but it has taken me way longer to get a PR up than I was hoping.

PR #248, opened 2018-10-10
(SLV-190) Add baseline results and comparison per process

samwoods1:
Removed Baseline_Result class in favor of a simple hash
Generalized baseline comparison logic and moved it into baseline_assert method
Updated avg_dsk_write to avg_disk_write to match property name from beaker-benchmark
Added info on querying bigquery data to readme


samwoods1:
@RandellP Yep, it is already rebased, should be good.

PR #250, opened 2018-11-01
(SLV-266) turn of per process cpu check

RandellP:
currently not working, disabling until fixed.


samwoods1:
Looks like this needs a rebase.

PR #256, opened 2019-01-08
(SLV-246) GPLT - Run 'puppet infrastructure tune' for autoscale setup, capture puppet-metrics-collector output

billclaytor:
This update adds `setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb` to the pre-suite. This test performs the steps to run and apply `puppet infrastructure tune` on the master if the `SCALE_TUNE` environment variable is set to 'true', so it is only run during the autoscale setup or in the standalone 'autoscale_tune' rake task.

Since the autoscale runs produce a separate 'PERF' folder for each iteration the results have been moved to a `results/perf` for individual runs / iterations and `results/scale` for the collected scale run results. The scale results also provide a `results/scale/latest` link which points to the latest 'PERF_SCALE' result folder. This is used in the autoscale rake tasks to copy the latest log output to the latest scale result folder after the beaker run is complete.

The results gathering has also been updated to include the output from puppet-metrics-collector on the master. The output folders are purged before each iteration of the autoscale run so that only the applicable output is captured. 

The autoscale rake tasks currently include 'tiny', 'small', and 'medium' variants which run alternate versions of the scale test to provide a quick way to test the autoscale and results gathering functionality. These can be removed before merging if they are no longer required.

--

### DIFF
--- a/rakefile
+++ b/rakefile
@@ -270,7 +270,7 @@ desc 'Run Scale setup and test for gatling'
 rototiller_task :autoscale do
   ENV['ENVIRONMENT_TYPE'] = 'gatling'
   ENV['BEAKER_TESTS'] = ENV['BEAKER_TESTS'] || 'tests/Scale.rb'
-  ENV['BEAKER_PRESERVE_HOSTS'] = 'always'
+  ENV['BEAKER_PRESERVE_HOSTS'] = ENV['BEAKER_PRESERVE_HOSTS'] || 'always'
   ENV['PUPPET_SCALE_CLASS'] = 'role::by_size::small'
   ENV['SCALE_TUNE'] = ENV['SCALE_TUNE'] || 'true'
   Rake::Task["performance"].execute
@@ -283,7 +283,7 @@ rototiller_task :autoscale_setup do
   Rake::Task["autoscale"].execute
 end
 
-desc 'Run puppet infrastructure tune setup for gatling'
+desc 'Run puppet infrastructure tune setup for gatling on a previously provisioned master'
 rototiller_task :autoscale_tune do
   ENV['ENVIRONMENT_TYPE'] = 'gatling'
   ENV['BEAKER_TESTS'] = 'setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb'
--- a/rakefile
+++ b/rakefile
@@ -272,6 +272,7 @@ rototiller_task :autoscale do
   ENV['BEAKER_TESTS'] = ENV['BEAKER_TESTS'] || 'tests/Scale.rb'
   ENV['BEAKER_PRESERVE_HOSTS'] = 'always'
   ENV['PUPPET_SCALE_CLASS'] = 'role::by_size::small'
+  ENV['SCALE_TUNE'] = ENV['SCALE_TUNE'] || 'true'
   Rake::Task["performance"].execute
   Rake::Task["autoscale_copy_log"].execute unless ENV['BEAKER_TESTS'] == ''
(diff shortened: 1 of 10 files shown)


rakefile
```diff
@@ -267,30 +267,50 @@ rototiller_task :acceptance do |t|
...
+rototiller_task :autoscale_setup do
   ENV['BEAKER_TESTS'] = ''
   Rake::Task["autoscale"].execute
 end
 
+desc 'Run puppet infrastructure tune setup for gatling'
+rototiller_task :autoscale_tune do
+  ENV['ENVIRONMENT_TYPE'] = 'gatling'
+  ENV['BEAKER_TESTS'] = 'setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb'
+  ENV['BEAKER_PRESERVE_HOSTS'] = 'always'
```
RandellP: feels like this shouldn't be so absolute... Like if I ran a sim with it set to on_fail... or something you would have just overwrote it.  Maybe some kind of conditional here...  But it depends on the entry points that can get here...

rakefile
```diff
@@ -267,30 +267,50 @@ rototiller_task :acceptance do |t|
...
 end
 
 desc 'Run Scale setup for gatling'
-rototiller_task :autoscale_setup do |t|
+rototiller_task :autoscale_setup do
   ENV['BEAKER_TESTS'] = ''
   Rake::Task["autoscale"].execute
 end
 
+desc 'Run puppet infrastructure tune setup for gatling'
```
RandellP: looks like it also runs performance at the end.  Should maybe whatever calls this be the thing that calls performance if it wants to continue on?  I dunno.  The overall flow of this rake file is very obfuscated.

results/RESULTS.md
```diff
@@ -0,0 +1,10 @@
+# Performance and Scale Results
+gatling-puppet-load-test results are now copied to this directory rather than the parent directory.
```
RandellP: should "this" be "these"?  referring to the below sections...

rakefile
billclaytor: I was trying to prevent the user from accidentally de-provisioning the hosts after a long scale run by forgetting to set this correctly. I've updated it to the following:
```
ENV['BEAKER_PRESERVE_HOSTS'] = ENV['BEAKER_PRESERVE_HOSTS'] || 'always'
```

rakefile
billclaytor: I've updated the description to: 
```
desc 'Run puppet infrastructure tune setup for gatling on a previously provisioned master
```
This task is only for running standalone to tune (or re-tune) an already provisioned master.

I've updated the NEW_README with more info about the tasks. In a nutshell:
* `autoscale` is the primary task. When run by itself it runs the pre-suite and test.
* `autoscale_setup` provisions and runs the pre-suite by setting `ENV['BEAKER_TESTS'] = ''` and calling the `autoscale` task.
* `autoscale_tune` runs the tune test from the pre-suite against previously provisioned nodes.
* `autoscale_provisioned` runs the standard scale test against previously provisioned nodes.
* `autoscale_provisioned_x` runs variations on the scale test against previously provisioned nodes.
* `autoscale_copy_log` copies the latest log files to the latest perf scale results folder after the beaker run completes.

So you can either run `autoscale` to provision and run the test or `autoscale_setup` to provision and `autoscale_provisioned` or a variation to run the test.

results/RESULTS.md
billclaytor: Updated

PR #258, opened 2019-01-24
(SLV-317) remove private key for control repo

RandellP:
control repo is a public github repo.  But we were accessing it using ssh, so we needed a key.  change to http and no longer need the key.

This was driven by the untimely death of the int-resources server that were using to store the key for the repo.


spec/setup/helpers/perf_helper_spec.rb
```diff
@@ -78,28 +78,9 @@ def options
 
     it 'adds the private key, configures ssh, and installs git on each master' do
```
billclaytor: This should probably be updated to: `it ' installs git on each master'`

PR #259, opened 2019-02-12
(SLV-206) Johnson scale test enhancements

billclaytor:
This update provides improvements to tuning and results output for Johnson scale testing:

* add AWS environment variables to beaker env output
* update scale results setup to output current PE tune
* allow scale base instances to be set by environment variable
* update tuning step:
  * optionally use '--force' option to allow use on smaller instances
  * use --hiera option to isolate yaml output
  * output current tune after application
* update master log copy step to avoid file duplication
* added a proof-of-concept to convert CSV files to HTML tables during post-test results step


setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb
```diff
@@ -1,25 +1,32 @@
...
-    puts "Extracted the following data:"
-    puts data
+    puts "Extracted the following output:"
+    puts tune_output
+    puts
 
     puts "Appending to common.yml"
     puts
-    on master, "echo \"#{data}\" >> #{common_yaml}"
+    on master, "echo \"#{tune_output}\" >> #{common_yaml_path}"
```
RandellP: If I read this right you are putting the tune results in both the common and the node specific yaml file?  If so why both?  Could make for confusion is someone tried to edit command and can't figure out why the edit doesn't work...

setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb
```diff
@@ -1,25 +1,32 @@
...
-    output = on(master, "puppet infrastructure tune").output
-
-    # get data between '---' and the first empty line
-    data = output.match(/## Specify(.*)## CPU/m)[1].gsub("---", "").strip
-
-    # remove control codes
-    data = data.match(/0;32m(.*)\n.*\[/m)[1].strip + "\n"
+    # tune with hiera output and (and --force if specified)
+    hiera_option = " --hiera #{tune_output_dir}"
+    force_option = ENV["PUPPET_GATLING_SCALE_TUNE_FORCE"].eql?("true") ? " --force" : ""
```
RandellP: Looks like PUPPET_GATLING_SCALE_TUNE_FORCE has no default... should it?  And I assume force will cause it to override an existing file.  Any reason we wouldn't want to do that by default?

tests/helpers/perf_run_helper.rb
```diff
@@ -344,10 +411,12 @@ def handle_scale_results(scenario_hash)
...
 
     # check results for KOs
-    check_scale_results(scenario_hash)
+    success = check_scale_results(scenario_hash)
 
     # perform assertions
-    scale_assertions(atop_result, gatling_result)
+    scale_assertions(atop_result, gatling_result) unless !success
+
+    return success
```
RandellP: This appears to return the result of check_scale_results.  But when I read a method like scale_assertions I would expect it might find things that are relevant to the return value...

setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb
billclaytor: It is using the '--hiera' option to generate the output file but it is just copying the contents of that file to the 'common.yaml' file. Getting the output from the file eliminates the extra output from the command. The 'tune_output' dir is just an arbitrary location to output the file.

setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb
billclaytor: This ensures that the command will only be run with the '--force' option if the PUPPET_GATLING_SCALE_TUNE_FORCE has specifically been set to 'true'. The '--force' option removes the enforcement of minimum system requirements. It has been added to allow testing of tuning smaller instances (m4.large, etc...). Nothing currently prevents overwriting the existing file.

tests/helpers/perf_run_helper.rb
billclaytor: I think the idea here was to not bother performing the assertions if check_scale_results has already determined that the iteration was not successful. I was having issues getting the results cleanly at the end if the Beaker run failed, but perhaps that should be handled using the teardown.

setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb
RandellP: right, but isn't the file it is creating (the node file) enough to create the desired change?  So why also copy it to common?  Seems like the changes in common will have no effect.

setup/install_gatling/50_tune/10_puppet_infrastructure_tune.rb
RandellP: oh yeah.  I was thinking force was to overwrite the file...

tests/helpers/perf_run_helper.rb
RandellP: ok

PR #268, opened 2019-03-22
(SLV-111) Add document for ABS

johnduarte:
This commit adds a document describing ABS as well as how to obtain an
ABS token and add it to `$HOME/.fog`.

### DIFF
--- a/docs/abs.md
+++ b/docs/abs.md
@@ -1,0 +1,81 @@
+# Always be Scheduling (ABS)
+
+**NOTE: This facility is not publically available and can only be used by
+personnel employed by Puppet the company.**
+
+[Always be Scheduling](https://github.com/puppetlabs/always-be-scheduling)
+is a host scheduler to ensure that hosts are returned to the requester as
+quickly as possible.  Requests are queued in order to fill them on in the
+order in which they were received.
+The
+[`awsdirect`](https://github.com/puppetlabs/always-be-scheduling#apiv2awsdirect)
+API can also be used to bypass the queue processor.
+
+
+## Finding ABS
+
+ABS is hosted within puppetlabs and managed by mesosphere.  The
+[mesosphere marathon](https://github.com/mesosphere/marathon)
+orchestrator can be queried to discover what FQDN names are being used for the
+ABS application.  This information can be found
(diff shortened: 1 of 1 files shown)


docs/abs.md
```diff
@@ -0,0 +1,77 @@
+# Always be Scheduling (ABS)
+
+**NOTE: This facility is not publically available and can only be used by
+personnel employed by Puppet the company.**
+
+[Always be Scheduling](https://github.com/puppetlabs/always-be-scheduling)
+is a host scheduler to ensure that hosts are returned to the requester as
+quickly as possible.  Requests are queued in order to fill them on in the
```
billclaytor: We may want to note that we use the `awsdirect` endpoint which "allows you to make a direct request for an aws resource instead of the request going through the queue-processor".

billclaytor: Just thinking about the new docs directory here: If we are going to create a new docs directory should any of the other docs (CD4PE.md, etc...) be moved there for consistency? Perhaps even move the NEW_README.md there and rename it to something more descriptive? 

Also, the names of the .md files usually seem to be capitalized. Is this actually a standard? It makes sense for the README.md file to make it stand out but not sure if it is necessary if we start a docs directory. However, we should probably adhere to a standard within a directory for consistency.

docs/abs.md
```diff
@@ -0,0 +1,77 @@
...
+that value into the standard API call for
+[generating a token](https://github.com/puppetlabs/always-be-scheduling#post-token).
+
+
+```
+$ curl -X POST -d '' -u user.name --url https://cinext-abs.delivery.puppetlabs.net/api/v2/token
+Enter host password for user 'testuser':
+```
+
+Once the password has been entered, ABS should return a token value
```
RandellP: which password... :)   I think it was the puppet pass password...

docs/abs.md
johnduarte: @billclaytor I have opened [SLV-356](https://tickets.puppetlabs.com/browse/SLV-356) to address the organization issue.

johnduarte:
I have amended the commit to address your feedback.

docs/abs.md
billclaytor: @johnduarte Thanks!

PR #271, opened 2019-04-01
(SLV-319) Set warmup 300 agents - 10 min iteration

johnduarte:
This commit updates the `WarmUpJit.json` data to run 300 agents in a
single 10 minute iteration.

An inaccurate comment related to this data was removed from the
`ApplesToApples.rb` test.


johnduarte:
This change has not yet been tested to address warm up issues and should not yet be merged.

johnduarte:
This change appears to prevent the initial spike load when kicking off a2a task.

PR #279, opened 2019-04-23
(SLV-418) Enable CLAMPS from rake task

johnduarte:
This PR enables the installation and configuration of CLAMPS by setting the `ENVIRONMENT_TYPE` environment variable to `clamps` and running any of the various performance rake tasks (such as `performance`).


config/beaker_hosts/perf-test.cfg
```diff
@@ -0,0 +1,55 @@
+default_vmname: &default_vmname centos-7-x86_64
+default_platform: &default_platform el-7-x86_64
+default_snapshot: &default_snapshot pe
+
+:root_keys: true
+
+:answers:
+  :console_admin_password: puppetlabs
+:department: eso-dept
```
RandellP: We could just wait for the refactor... but since you were messing with the file I figured I would at least point out that most of this file does nothing.  All the abs stuff is meaningless.  The abs_helper handles all those settings.  So like department, lifetime, volume size...  I am not even sure this file does anything for an abs based run.  I believe it does have meaning for vmpooler acceptance testing though.

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+# users on a system to simulate agents. It will then randomize their puppet
+# agent -t times to simulate a large number of agents.
+def add_clamps_group_ca(classifier, pe_infra)
+  # The Clamps::master class manages auth.conf for the agents - by setting it
+  # to allow all.  This is insecure and worth looking at overriding at some
+  # point, but this should only be used for metrics / load testing on a private
+  # network, so shouldn't be an issue short term.
+  clamps_ca_group = {
+    'name'    => 'Clamps CA',
+    'rule'    => ['or', ['=', 'name', master.node_name]], # pinned node
```
RandellP: why the "or".  There is only one rule

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+    'name'    => 'Clamps - Agent Nodes',
+    'rule'    => ['and', ['=', %w[fact id], 'root'], ['=', 'name', metric.node_name]],
+    'parent'  => pe_infra['id'],
+    'classes' => {
+      'clamps::agent' => {
+        'amqpass'               => '',
+        'master'                => any_hosts_as?(:loadbalancer) ? 'puppet' : master.node_name,
+        'nonroot_users'         => options[:scale][:num_nonroot_users]       || 2,
+        'daemonize'             => options[:scale][:daemonize]               || false,
+        'mco_daemon'            => options[:scale][:mco_daemon],
```
RandellP: where does the value come from?  I don't see it set anywhere.

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+    'parent'  => pe_infra['id'],
+    'classes' => {
+      'clamps::master' => {}
+    }
+  }
+  classifier.find_or_create_node_group_model(clamps_ca_group)
+end
+
+# Creates a node group for creating the fake agents. This node group should
+# match all users that are running as root.
```
RandellP: should that be "only users that are running as root".  saying all implies there are more than one.  And the rule limits it to the metrics node, so there should only be one root user.

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+def add_clamps_group_agent(classifier, pe_infra)
+  clamps_agents_group = {
+    'name'    => 'Clamps - Agent Nodes',
+    'rule'    => ['and', ['=', %w[fact id], 'root'], ['=', 'name', metric.node_name]],
+    'parent'  => pe_infra['id'],
+    'classes' => {
+      'clamps::agent' => {
+        'amqpass'               => '',
+        'master'                => any_hosts_as?(:loadbalancer) ? 'puppet' : master.node_name,
+        'nonroot_users'         => options[:scale][:num_nonroot_users]       || 2,
```
RandellP: do you know if it is getting the options value or the hardcode.  I am not sure the config file where you define the scale section even gets used in abs runs.

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+        'percent_changed_facts' => options[:scale][:percent_facts_to_change] || 15,
+        'splay'                 => options[:scale][:splay]                   || false,
+        'splaylimit'            => options[:scale][:splaylimit]              || false
+      }.reject { |_k, v| v.nil? }
+    }
+  }
+  classifier.find_or_create_node_group_model(clamps_agents_group)
+end
+
+def add_clamps_group_users(classifier, pe_infra)
```
RandellP: comment fo the method?

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+  clamps_users_group = {
+    'name'    => 'Clamps - Agent Users (non root)',
+    'rule'    => ['and', ['not', ['=', %w[fact id], 'root']]],
```
RandellP: If I read this right... it would apply to any non-root user on any node.  Where as the 'Clamps - Agent Nodes' group applies to the root user on only the metrics node.  Seems like they should both limit to the metrics node as far as I can tell.

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+      'clamps' => {
+        'num_static_files'  => options[:scale][:static_files]  || 20,
+        'num_dynamic_files' => options[:scale][:dynamic_files] || 5
+      }
+    }
+  }
+  classifier.find_or_create_node_group_model(clamps_users_group)
+end
+
+def update_pe_agent_rule_clamps(classifier, pe_agent)
```
RandellP: method comment

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,94 @@
...
+  rules = pe_agent['rule']
+  rules << ['not', ['~', %w[fact clientcert], ".*#{metric.node_name}"]]
+  rules << ['=', %w[fact id], 'root']
+  rules[0] = 'or'
+  classifier.update_node_group(pe_agent['id'], 'rule' => rules)
+end
+
+def clamps_enabled?
```
RandellP: method comment

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: This would be set in the beaker options. I have included it for backward compatibility. If it is unset, the value is removed from the parameter list by the `reject` call at the end of the hash.  For current and future PE versions this value would not be set since we are no longer using MCO.

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: This is the way the rule was defined in the original clamps classification file. I don't think it matters which logical operator is used.

config/beaker_hosts/perf-test.cfg
johnduarte: I would suggest to wait for the refactor. This file 'works out of the box', so cleaning it up here is not in alignment with the objectives of the PR.

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: This value is set via the beaker options and it is being picked up by my ABS runs.
https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/config/beaker_hosts/pe-perf-test.cfg#L18-L25

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: No, the users group gets applied to all of the generated agents (e.g `user-1-ip-10-227.0.222-amz-dev.puppet.net`). So, this rule classifies all of the non-root agents running on generated agents.

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: Additional commit made to add method comments.

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: Additional commit made to add method comments.

setup/install_gatling/30_classification/90_clamps.rb
johnduarte: Additional commit made to add method comments.

johnduarte:
@RandellP I have made an additional commit made to add method comments.  I have attempted to answer your other questions to the best of my ability. Let me know if think further changes need to be made.

setup/install_gatling/30_classification/90_clamps.rb
```diff
@@ -0,0 +1,109 @@
...
+  classifier = get_classifier
+
+  step 'add clamps classification' do
+    pe_infra = classifier.get_node_group_by_name('PE Infrastructure')
+    add_clamps_group_ca(classifier, pe_infra)
+    add_clamps_group_agent(classifier, pe_infra)
+    add_clamps_group_users(classifier, pe_infra)
+  end
+
+  step 'remove non-root agents from classification' do
```
RandellP: Need some context as to why they are being removed.

johnduarte: @RandellP I have added a comment below this line to clarify why they must be removed from the PE Agent node group. Thanks.

PR #281, opened 2019-04-26
(SLV-361) Create ctrl repo from prod env via bolt

johnduarte:
This commit adds a bolt task for creating a control repo on a
puppet master host using the current contents of the production
environment code directory as its initial commit.

The `Boltdir` directory tree scaffolding is added to allow bolt
to automatically discover the task.

Show the task:
```
$ bolt task show
...
gplt::create_control_repo_from_production_env
...
```

Run the task:
```
bolt task run \
  --no-host-key-check \
  --private-key ~/.ssh/id_rsa-acceptance \
  --nodes ip-10-xx-xx-xx.amz-dev.puppet.net \
  gplt::create_control_repo_from_production_env
```

### DIFF
--- a/Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
+++ b/Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -e
+
+CTRL_REPO_PATH="${PT_ctrl_repo_path:-/opt/puppet/control-repo.git}"
+[ -z "$CTRL_REPO_PATH" ] && echo "Must specify a control repo path" && exit 1
+
+CTRL_REPO_BASEDIR=$(dirname $CTRL_REPO_PATH)
+
+ENVIRONMENT=production
+PUPPET_PROD_ENV_DIR=/etc/puppetlabs/code/environments/$ENVIRONMENT
+CHECKOUT_DIR=/tmp/control-repo
+
+
+mkdir -p "$CTRL_REPO_BASEDIR"
+
+if [ -d "$CTRL_REPO_PATH" ]; then
+    echo "ERROR: The control repo destination already exists" >&2
+    exit 1
+fi
+
+git init --bare $CTRL_REPO_PATH
+git clone $CTRL_REPO_PATH $CHECKOUT_DIR
+cp -r $PUPPET_PROD_ENV_DIR/* $CHECKOUT_DIR
+rm -fr $CHECKOUT_DIR/modules
+cd $CHECKOUT_DIR \
+    || { echo "ERROR: Failed to cd into $CHECKOUT_DIR" >&2 ; exit 1 ; }
+git add .
+git commit -m "Initial commit"
+git branch -m $ENVIRONMENT
+git push -u origin $ENVIRONMENT
+cd $CTRL_REPO_PATH
+git symbolic-ref HEAD refs/heads/$ENVIRONMENT
+
+rm -rf $CHECKOUT_DIR
+
+# send structured data to bolt
+printf '{"control_repo_path":"%s"}\n' "$CTRL_REPO_PATH"
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -1,0 +1,43 @@
+# Bolt Tasks
+
+This repo includes bolt tasks.  The available tasks are enumerated here.
+
+The following `bolt` commands must be executed from the root directory of a
+checkout of this repository.
+
+## gplt::create_control_repo_from_production_env
+
+This task creates a git repo using the contents from the given node's
+puppet production environment.  Therefore, the specified node should be
+the primary puppet master.
+
+The task takes one optional argument `ctrl_repo_path=<value>` which is used
+as the fully qualified path for creating the git repository.
+
+Note: The bolt task needs to be run as the `root` user (which is the default)
+in order for the default `ctrl_repo_path` to be successfully created.
+
+The built in documentation for the task can be seen by running the following:
+```
+bolt task show gplt::create_control_repo_from_production_env
+
+gplt::create_control_repo_from_production_env - Creates a git repo from puppet production environment directory
+
+USAGE:
+bolt task run --nodes <node-name> gplt::create_control_repo_from_production_env ctrl_repo_path=<value>
+
+PARAMETERS:
+- ctrl_repo_path: String[1]
+    The fully qualified path to be used for creating the control repo
+```
+
+Typical usage on an ABS instance would use the following bolt pattern:
+```
+bolt task run \
+  --user root \
+  --no-host-key-check \
+  --private-key ~/.ssh/id_rsa-acceptance \
+  --nodes ip-10-xx-xx-xx.amz-dev.puppet.net \
+  gplt::create_control_repo_from_production_env \
+  ctrl_repo_path="/opt/puppet/control-repo.git"
+```
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ jenkins-integration/dev/target_machine_local.yml
 # Artifacts created by run
 tmp
 results
+
+# Bolt
+.rerun.json
--- a/Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.json
+++ b/Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.json
@@ -1,0 +1,10 @@
+{
+  "description": "Creates a git repo from puppet production environment directory",
+  "input_method": "environment",
+  "parameters": {
+    "ctrl_repo_path": {
+      "description": "The fully qualified path where the control repo will be created",
+      "type": "Optional[String[1]]"
+    }
+  }
+}


Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
```diff
@@ -0,0 +1,25 @@
...
+git init --bare $CONTROL_REPO_URL
+git clone $CONTROL_REPO_URL $CHECKOUT_DIR
+cp -r $PUPPET_PROD_ENV_DIR/* $CHECKOUT_DIR
+rm -fr $CHECKOUT_DIR/modules
+cd $CHECKOUT_DIR \
+    || { echo "ERROR: Failed to cd into $CHECKOUT_DIR" >&2 ; exit 1 ; }
+git add .
+git commit -m "Initial commit"
+git branch -m $ENVIRONMENT
+git push -u origin $ENVIRONMENT
```
RandellP: When I tried this general flow, then tried to clone the repo to a new place I got this 
warning: remote HEAD refers to nonexistent ref, unable to checkout.
I think we need to change the default branch to be production.

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
```diff
@@ -0,0 +1,25 @@
...
+CHECKOUT_DIR=/root/control-repo
+
+mkdir -p $BASEDIR
+
+if [ -d $CONTROL_REPO_URL ]; then
+    echo "ERROR: The control repo destination already exists" >&2
+    exit 1
+fi
+
+git init --bare $CONTROL_REPO_URL
```
RandellP: Do we need to do some kind of check to ensure that git is installed?  I know it is on our gatling setups, but not sure if it is an a normal install.

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
RandellP: Do we need to do any checking about these git commands exist status?  Like if somehow it failed at the end, would bolt give us a useful error message, or would it expect us to produce something?

johnduarte: Line 24 moves the default `master` branch to `production`. This line then pushes the production branch to the remote.  Since the remote was bare to begin with, the only branch that exists after this line is the `production` branch.

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
johnduarte: I will double check, but since code manager uses git I don't believe a modern PE master can function without git installed already.

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
johnduarte: No, if any of the above fail, the script will exit because we are not ignoring failures. The contents of STDERR are then echoed via bolt as part of the failure json response.

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
johnduarte: Even if git is not installed, bolt will return the STDERR result to the user.
```
$ bolt task run   --no-host-key-check   --private-key ~/.ssh/id_rsa-acceptance   --nodes ip-10-227-2-25.amz-dev.puppet.net gplt::create_control_repo_from_production_env
Started on ip-10-227-2-25.amz-dev.puppet.net...
Failed on ip-10-227-2-25.amz-dev.puppet.net:
  The task failed with exit code 1:
  /tmp/6d70bdb9-09ad-4eda-a8c6-376ffa3fc95b/create_control_repo_from_production_env.sh: line 16: git: command not found
  /tmp/6d70bdb9-09ad-4eda-a8c6-376ffa3fc95b/create_control_repo_from_production_env.sh: line 17: git: command not found
  cp: cannot stat ‘/etc/puppetlabs/code/environments/production/*’: No such file or directory
  /tmp/6d70bdb9-09ad-4eda-a8c6-376ffa3fc95b/create_control_repo_from_production_env.sh: line 20: cd: /root/control-repo: No such file or directory
  ERROR: Failed to cd into /root/control-repo

  {
  }
Failed on 1 node: ip-10-227-2-25.amz-dev.puppet.net
Ran on 1 node in 1.80 seconds
```

johnduarte: Oops, looks like I forgot to include `set -e`. :man_facepalming: 
Here, this is better now:
```
$ bolt task run   --no-host-key-check   --private-key ~/.ssh/id_rsa-acceptance   --nodes ip-10-227-2-25.amz-dev.puppet.net gplt::create_control_repo_from_production_env
Started on ip-10-227-2-25.amz-dev.puppet.net...
Failed on ip-10-227-2-25.amz-dev.puppet.net:
  The task failed with exit code 127:
  /tmp/1f7dc3b5-c5bd-4fbf-af73-61b3f3882f99/create_control_repo_from_production_env.sh: line 18: git: command not found

  {
  }
Failed on 1 node: ip-10-227-2-25.amz-dev.puppet.net
Ran on 1 node in 1.56 seconds
```

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
RandellP: yeah, but git still throws an error if you clone it.  And the branch is set to master sort of...

➜  blue git clone /tmp/foo.git
Cloning into 'foo'...
done.
warning: remote HEAD refers to nonexistent ref, unable to checkout.

➜  blue cd foo
➜  foo git:(master)

Though git branch returns nothing.  And if you checkout production can you can't checkout master to go back.  But you end up in a really odd state...  And in fact you don't have any files present after the initial clone.

This is what determines the default branch
➜  foo git:(master) cat .git/HEAD
ref: refs/heads/master

running
git symbolic-ref HEAD refs/heads/production
after cd'ing into $CONTROL_REPO_URL
even before you created the production branch will fix it.

Boltdir/.rerun.json
```diff
@@ -0,0 +1 @@
+[{"target":"ip-10-227-2-25.amz-dev.puppet.net","status":"failure"}]
```
RandellP: I don't think you meant to commit this file

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
johnduarte: Thanks for finding this @RandellP . I have amended the script to update the bare repo as you have indicated.

Boltdir/.rerun.json
johnduarte: :man_facepalming: I have removed it from the PR.

johnduarte:
Ok, @RandellP I think I have addressed your concerns. Please let me know if there are further updates required.

docs/tasks.md
```diff
@@ -0,0 +1,23 @@
...
+The built in documentation for the task can be seen by running the following:
+```
+bolt task show gplt::create_control_repo_from_production_env
+```
+
+Typical usage on an ABS instance would use the following bolt pattern:
+```
+bolt task run \
+  --no-host-key-check \
+  --private-key ~/.ssh/id_rsa-acceptance \
```
RandellP: need --user root

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
```diff
@@ -0,0 +1,29 @@
...
+cp -r $PUPPET_PROD_ENV_DIR/* $CHECKOUT_DIR
+rm -fr $CHECKOUT_DIR/modules
+cd $CHECKOUT_DIR \
+    || { echo "ERROR: Failed to cd into $CHECKOUT_DIR" >&2 ; exit 1 ; }
+git add .
+git commit -m "Initial commit"
+git branch -m $ENVIRONMENT
+git push -u origin $ENVIRONMENT
+cd $CONTROL_REPO_URL
+git symbolic-ref HEAD refs/heads/$ENVIRONMENT
```
RandellP: Should add output on where the repo was created, and where the checked out copy can be found.  It's at the top of the output, but the git output gets long, so having it again at the bottom will make it be easier to find and be seen.

docs/tasks.md
```diff
@@ -0,0 +1,23 @@
+# Bolt Tasks
+
+This repo includes bolt tasks.  The available tasks are enumerated here.
+
```
RandellP: Add generic comment about the repo needing to be the CWD for the below example commands to work

docs/tasks.md
johnduarte: I believe that bolt defaulted to `--user root` if it was not specified.

Ref: https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#ssh-transport-configuration-options

Boltdir/site-modules/gplt/tasks/create_control_repo_from_production_env.sh
johnduarte: I have pushed an update incorporating a required parameter for the control repo path.  This value is returned as structured data to bolt at the end of the task.

docs/tasks.md
johnduarte: I have added the comment in the latest update.

johnduarte:
@RandellP I have updated the PR to incorporate your most recent round of feedback. Let me know if think other changes should be made.

docs/tasks.md
RandellP: When I ran it without --user root it tried to ssh into the box as randell...  which failed.  And frankly that doc doesn't sound right. Why would you want it to ssh as root instead of as the current user by default.   Either way they either have a bug in their code or documentation or the --private-key changes the default user.  Or could be a new feature to a very recent release... I am probably not on the latest.  I would add it in to the documentation to be safe, and to guard from future bolt changes.

PR #282, opened 2019-04-29
(SLV-390) Fail if local execs fail 

johnduarte:
Add `fail_test` guard calls for local execs.

Update perf_helper and its spec tests to conform to rubocop linting rules.

### DIFF
--- a/setup/helpers/perf_helper.rb
+++ b/setup/helpers/perf_helper.rb
@@ -413,6 +413,15 @@ node 'default' {}
     on(host, "cat #{manifestfile}")
   end
 
+  # Get the `puppet config print` value
+  #
+  # @param [Beaker::Host] host
+  # @param [String] config
+  # @returns [String]
+  def get_puppet_config(host, config)
+    on(host, puppet("config print #{config}")).stdout.chomp
+  end
+
   def configure_gatling_auth
     # Make room for local copies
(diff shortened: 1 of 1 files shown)


setup/helpers/perf_helper.rb
```diff
@@ -21,20 +25,26 @@ def install_epel_packages
...
     step 'disable selinux on metrics server' do
       # disable selinux immediately, but not persistent after a reboot
       on agents, 'setenforce 0 || true'
       # required to disable selinux between reboots
-      on agents, "sed -i 's/^SELINUX=.*/SELINUX=disabled/g' /etc/sysconfig/selinux && cat /etc/sysconfig/selinux" unless
+      sed_cmd = "sed -i 's/^SELINUX=.*/SELINUX=disabled/g' "\
+                '/etc/sysconfig/selinux '\
+                '&& cat /etc/sysconfig/selinux'
+      on agents, sed_cmd unless
+          # FIXME: What is `agents[0]` intended to be?
```
RandellP: I think this was just a catch for in case hosts were centos6...  It assume all hosts are the same os though... probably should loop through all agents and run the sed command on anything that isn't centos6 instead.

RandellP: of course the comment says only on the metrics box... so I guess it is assuming there is only one agent and it is the metrics box.  :(

PR #288, opened 2019-05-13
(SLV363)Task take public github repo as source

RandellP:
updated the gplt::create_control_repo_from_production_env task to be able to accept a public github repo to be the source of the local repo it creates.  Also accepts a branch, which it will convert to the production branch in the new repo.
This allows for easy creation of a repo to use with the simulation from any branch of any public control repo.
The intent is to then point code manager at this repo as part of setup.

updated docs on the task as well

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -7,12 +7,12 @@ checkout of this repository.
 
 ## gplt::create_control_repo_from_production_env
 
-This task creates a git repo using the contents from the given node's
-puppet production environment.  Therefore, the specified node should be
+This task creates a git repo using either a path (commonly the contents from the given node's
+puppet production environment) or a public GitHub URL.  Normally the specified node should be
 the primary puppet master.
 
-The task takes one optional argument `ctrl_repo_path=<value>` which is used
-as the fully qualified path for creating the git repository.
+The task takes optional arguments which can be used to identify the source for the new git repo or
+to specify the fully qualified path for where the new repo should be created.
 
 Note: The bolt task needs to be run as the `root` user (which is the default)
 in order for the default `ctrl_repo_path` to be successfully created.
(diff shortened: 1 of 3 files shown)


docs/tasks.md
```diff
@@ -7,12 +7,12 @@ checkout of this repository.
 
 ## gplt::create_control_repo_from_production_env
 
-This task creates a git repo using the contents from the given node's
-puppet production environment.  Therefore, the specified node should be
+This task creates a git repo using from either a path, commonly the contents from the given node's
```
billclaytor: Conflict with 'using' vs 'from'... suggested fix:
"This task creates a git repo using either a path (commonly the contents of the given node's Puppet production environment) or a public GitHub URL."

RandellP:
@billclaytor I force pushed an amended update for the doc file to fix it based on your feedback.

docs/tasks.md
```diff
@@ -21,14 +21,20 @@ The built in documentation for the task can be seen by running the following:
 ```
 bolt task show gplt::create_control_repo_from_production_env
 
-gplt::create_control_repo_from_production_env - Creates a git repo from puppet production environment directory
+gplt::create_control_repo_from_production_env - Creates a local git repo for puppet code
 
 USAGE:
-bolt task run --nodes <node-name> gplt::create_control_repo_from_production_env ctrl_repo_path=<value>
+bolt task run --nodes <node-name> gplt::create_control_repo_from_production_env ctrl_repo_path=<value> code_source_path=<value> code_source_url=<value> code_source_branch=<value>
```
billclaytor: Should this have separate usage examples for `code_source_path` and `code_source_url`?

RandellP: That is bolt generated.  We have no control, and can't add anything to it.

PR #298, opened 2019-06-25
(SLV-483) Automate provisioning of pe_xl nodes, creation of Bolt inventory and parameter files

billclaytor:
This update adds the following:
* new provisioning methods to abs_helper using the [parallel](https://github.com/grosser/parallel) gem
* a `util/abs` directory with the `provision_pe_xl_nodes.rb` script and a `templates` directory with parameterized HA / no-HA versions of the Bolt inventory (nodes_ha.yaml / nodes_no_ha.yaml) and parameter (params_ha.json / params_no_ha.json) files

Note: run the `provision_pe_xl_nodes.rb` script from the `gatling-puppet-load-test` directory:
```
ruby ./util/abs/provision_pe_xl_nodes.rb

or

ruby ./util/abs/provision_pe_xl_nodes.rb my_id ~/tmp/my/output/dir
```

TODO: add a doc to the docs folder


setup/helpers/abs_helper.rb
```diff
@@ -654,24 +660,143 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+      #   else
+      #     puts "The specified resource host array is not valid: #{abs_resource_hosts}"
+      #     puts
+      #   end
+      # rescue StandardError
+      #   # TODO: raise?
+      #   puts "JSON::ParserError encountered parsing the hosts array: #{abs_resource_hosts}"
+      # end
+
+      hosts = parse_abs_resource_hosts(abs_resource_hosts)
```
johnduarte: Why is this taken out of the rescue block?  There seem like a number of things that should be caught here (e.g. `parse_abs_resource_hosts` returns nil).

setup/helpers/abs_helper.rb
```diff
@@ -654,24 +660,143 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+  #   hostname = provision_host_for_role(role, "c5.xlarge", "40")
+  #
+  # TODO: spec test(s)
+  def provision_host_for_role(role, size = AWS_DEFAULT_SIZE, volume_size = AWS_VOLUME_SIZE)
+    host_to_provision = get_host_to_provision(role, size, volume_size)
+    abs_resource_hosts = get_abs_resource_hosts(host_to_provision)
+    raise "Unable to provision host via ABS" unless abs_resource_hosts
+
+    hosts = parse_abs_resource_hosts(abs_resource_hosts)
+    hostname = hosts[0]["hostname"]
```
johnduarte: What if hosts is nil?

setup/helpers/abs_helper.rb
```diff
@@ -654,24 +660,143 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+  #
+  def parse_abs_resource_hosts(abs_resource_hosts)
+    begin
+      hosts = JSON.parse(abs_resource_hosts)
+
+    # TODO: JSON::ParserError?
+    rescue StandardError => e
+      # TODO: raise? (current expected behavior is puts)
+      puts "#{e.inspect} encountered parsing the hosts array: #{abs_resource_hosts}"
+      hosts = nil
```
johnduarte: Should this return an empty array since the return type is defined as an array?

setup/helpers/abs_helper.rb
```diff
@@ -654,24 +660,143 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+    puts roles
+    puts
+
+    # this will be evaluated in abs_initialize
+    ENV["ABS_ID"] = abs_id unless abs_id.nil?
+    num_threads = roles.length > MAX_THREADS ? roles.length : MAX_THREADS
+
+    # uses in_threads to allow variable modification
+    Parallel.map(roles.each, in_threads: num_threads) do |role|
+      # role = "#{abs_id}_#{role}" unless abs_id.nil?
```
johnduarte: Should this be deleted?

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -1099,5 +1101,71 @@ class AbsHelperClass
       end
     end
   end
+
+  # TODO: review test cases
```
johnduarte: These should be implemented.

setup/helpers/abs_helper.rb
billclaytor: I've brought it back.

setup/helpers/abs_helper.rb
billclaytor: I've updated `parse_abs_resource_hosts` to raise the error, and `valid_abs_resource_hosts?` rescues.

setup/helpers/abs_helper.rb
billclaytor: See above; updated to raise the error.

setup/helpers/abs_helper.rb
billclaytor: Yes; that was left over after an update to using a separate tag for the id.

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Spec tests have been implemented

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -822,20 +832,25 @@ class AbsHelperClass
...
-  # TODO: implement
   describe "#backoff_sleep" do
     context "when called with a positive number of tries" do
-      it "sleeps for a multiple of the number of tries" do
-        skip
+      it "sleeps for a multiple of the number of tries and returns nil" do
+        tries = 2
+        sleep_time = 2**tries
+        message = "Sleeping for #{sleep_time} seconds after attempt #{tries}..."
+        expect(subject).to receive(:sleep).with(2**tries)
```
johnduarte: s/2**tries/sleep_time/

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -822,20 +832,25 @@ class AbsHelperClass
...
+        expect(subject).to receive(:puts).with(message)
+
+        expect(subject.backoff_sleep(tries)).to eq(nil)
```
johnduarte: The default return value of any ruby method is nil.  This test seems unnecessary.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -822,20 +832,25 @@ class AbsHelperClass
...
   describe "#backoff_sleep" do
     context "when called with a positive number of tries" do
-      it "sleeps for a multiple of the number of tries" do
-        skip
+      it "sleeps for a multiple of the number of tries and returns nil" do
+        tries = 2
+        sleep_time = 2**tries
+        message = "Sleeping for #{sleep_time} seconds after attempt #{tries}..."
+        expect(subject).to receive(:sleep).with(2**tries)
+        expect(subject).to receive(:puts).with(message)
```
johnduarte: Is the STDOUT message a crucial component to this method that needs to be tested?

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -1099,5 +1114,105 @@ class AbsHelperClass
       end
     end
   end
+
+  describe "#parse_abs_resource_hosts" do
+    context "when a valid hosts JSON string is specified" do
+      it "returns the hosts array" do
```
johnduarte: Parsing a valid JSON string does not necessarily return an array. The method states that is takes a JSON string, but doesn't define what a "valid hosts JSON string" is.
```
irb(main):002:0> require 'json'
=> true
irb(main):003:0> hosts = JSON.parse("{}")
=> {}
```

setup/helpers/abs_helper.rb
```diff
@@ -655,23 +663,123 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+  #
+  # @return [String] The hostname of the provisioned host
+  #
+  # @example
+  #   hostname = provision_host_for_role(role)
+  #     or
+  #   hostname = provision_host_for_role(role, "c5.xlarge", "40")
+  #
+  # TODO: spec test(s)
+  def provision_host_for_role(role, size = AWS_SIZE, volume_size = AWS_VOLUME_SIZE)
```
johnduarte: This method should be removed.  The functionality is covered by `provision_hosts_for_roles` below.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require "./setup/helpers/abs_helper.rb"
+include AbsHelper # rubocop:disable Style/MixinUsage
+
```
RandellP: Assuming this is the user interface.... needs some documentation about what it is supposed to do and such.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -0,0 +1,85 @@
...
+  # HA?
+  master_replica = hosts.detect { |m| m[:role] == "master_replica" }
+  suffix = if master_replica
+             "ha"
+           else
+             "no_ha"
+           end
+
+  # TODO: better alternative to having separate files for ha / no ha?
```
RandellP: Template files for the nodes yaml doesn't feel right.  They are the same file with just a list of hostname at the end.  Use a HERE doc for the top part and then just take in an array of hostnames to append to the bottom.  You can even parameterize the group name in the here doc.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -0,0 +1,85 @@
...
+  nodes_yaml = File.read("#{template_dir}/nodes_#{suffix}.yaml")
+  params_json = File.read("#{template_dir}/params_#{suffix}.json")
+
```
RandellP: And just because I am curious... why do we need two files with pretty much the same info in them?

billclaytor: These two files are specified when running the `pe_xl` plan:
https://github.com/reidmv/reidmv-pe_xl/blob/master/documentation/basic_usage.md#basic-usage-instructions

```
 bolt plan run pe_xl \
   --inventory nodes.yaml \
   --modulepath ~/modules \
   --params @params.json
```

setup/helpers/abs_helper.rb
```diff
@@ -655,23 +663,123 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
     else
 
       begin
-        hosts = JSON.parse(abs_resource_hosts)
-        host = hosts[0]
-        hostname = host["hostname"]
+        hosts = parse_abs_resource_hosts(abs_resource_hosts)
+        hostname = hosts[0]["hostname"]
```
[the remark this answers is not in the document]
billclaytor: The validity is confirmed by ensuring that the first entry has a hostname. This check could be moved into `parse_abs_resource_hosts` and cause a raise if not found which would simplify `valid_abs_resource_hosts?`. I added `parse_abs_resource_hosts` to avoid parsing the JSON in multiple places but wanted to maintain the current behavior to avoid further refactoring at this point. 

`valid_abs_resource_hosts?` was intended to verify the contents of the `last_abs_resource_hosts.log` file while in the case of `provision_host_for_role` it is parsing the output of `get_abs_resource_hosts` which is now being called with a single host per role rather than the initial 'Apples2Apples' design which passed the `master` and `metrics` hosts in the same array.

The whole thing needs a redesign at this point :)

setup/helpers/abs_helper.rb
```diff
@@ -655,23 +663,123 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+  #
+  # @return [Array<Hash>] The provisioned hosts
+  #
+  # @example
+  #   roles = ["master","puppet_db"]
+  #   hosts = provision_hosts_for_roles(roles)
+  #     or
+  #   hosts = provision_host_for_role(roles, "slv-test", "c5.xlarge", "40")
+  #
+  # TODO: spec test(s)
```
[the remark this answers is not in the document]
billclaytor: Removed

setup/helpers/abs_helper.rb
```diff
@@ -655,23 +663,123 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
+  def provision_hosts_for_roles(roles, abs_id = nil, size = AWS_SIZE, volume_size = AWS_VOLUME_SIZE)
+    hosts = []
+
+    puts "Provisioning hosts for the following pe_xl roles:"
```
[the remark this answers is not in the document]
billclaytor: Good point!

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Updated

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Updated

spec/setup/helpers/abs_helper_spec.rb
billclaytor: It is the feedback that lets the user know what is happening during the host verification process so I wanted to ensure that the values are correct. The exact wording isn't crucial but also probably won't change so it seemed simplest to verify the entire message.

setup/helpers/abs_helper.rb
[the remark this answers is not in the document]
johnduarte: Sorry, I did not express myself very well.  I understand the validation that is particular line provides an the work done in the rescue block.  With those validations in place, the previously existing validations are no longer necessary.  The method can be refactored because either `hostname = hosts[0]["hostname"]` throws an error or is good:

```
def valid_abs_resource_hosts?(abs_resource_hosts)
  is_valid = true
  begin
    hosts = parse_abs_resource_hosts(abs_resource_hosts)
    hostname = hosts[0]["hostname"]
  rescue StandardError => e
    puts "#{e.inspect}  encountered parsing the hosts array: #{abs_resource_hosts}"
    is_valid = false
  end
  return is_valid
end
```

spec/setup/helpers/abs_helper_spec.rb
billclaytor: Updated to move the actual validation into this method (raises an error if invalid) and included an example of a valid hosts JSON string in the description.

setup/helpers/abs_helper.rb
billclaytor: `provision_host_for_role` is called for each iteration of `provision_hosts_for_roles` and provides the hostname.

setup/helpers/abs_helper.rb
```diff
@@ -58,11 +62,14 @@ def abs_initialize
         @aws_reap_time = ENV["ABS_AWS_REAP_TIME"] || AWS_REAP_TIME
       end
 
+      @default_volume_size = ENV["AWS_DEFAULT_SIZE"] || AWS_SIZE
```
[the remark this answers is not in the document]
billclaytor: I initially added these as I intended to use them as the default values for `provision_host_for_role` and `provision_hosts_for_roles` but then realized that `abs_initialize` is actually run by `get_abs_resource_hosts`. I would like to try and eliminate the use of `abs_initialize` when refactoring for 'SLV-497: GPLT - Gemify abs_helper'.

setup/helpers/abs_helper.rb
```diff
@@ -655,23 +663,123 @@ def valid_abs_resource_hosts?(abs_resource_hosts)
...
           is_valid = true
         else
           puts "The specified resource host array is not valid: #{abs_resource_hosts}"
           puts
         end
-      rescue StandardError
+      rescue StandardError => e
         # TODO: raise?
-        puts "JSON::ParserError encountered parsing the hosts array: #{abs_resource_hosts}"
+        puts "#{e.inspect}  encountered parsing the hosts array: #{abs_resource_hosts}"
```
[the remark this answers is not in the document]
billclaytor: Nice! Updated to use `#{e.message}`.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require "./setup/helpers/abs_helper.rb"
+include AbsHelper # rubocop:disable Style/MixinUsage
+
+# TODO: specify as env variable or arg?
+ABS_ID = "slv-473"
```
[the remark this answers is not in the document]
billclaytor: Updated to use ARGV[0] if specified, otherwise "slv" as the default value.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -0,0 +1,85 @@
...
+
+ROLES_CORE = %w[master
+                puppet_db
+                compiler_a
+                compiler_b].freeze
+
+ROLES_HA = %w[master_replica
+              puppet_db_replica].freeze
+
+# TODO: specify as env variable or arg?
```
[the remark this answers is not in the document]
billclaytor: Currently just using ARGV[0] since this is the only input; created SLV-499 based on our discussion of the best way to provide host info to the script and integrate this with Beaker to run GPLT with the provisioned hosts.

util/abs/provision_pe_xl_nodes.rb
billclaytor: I have a 'TODO: add a doc to the docs folder' in the description to write up the doc once the code updates are complete.

util/abs/provision_pe_xl_nodes.rb
billclaytor: Based on discussion we decided to use these templates for now and determine a better approach in the future (SLV-499).

setup/helpers/abs_helper.rb
johnduarte: Ok.

spec/setup/helpers/abs_helper_spec.rb
johnduarte: I don't see this update applied yet.

setup/helpers/abs_helper.rb
```diff
@@ -635,43 +641,156 @@ def verify_abs_hosts(hosts)
...
+  #
+  def parse_abs_resource_hosts(abs_resource_hosts)
+    puts "Parsing the specified abs_resource_hosts JSON:"
+    puts abs_resource_hosts
+    puts
+
+    begin
+      hosts = JSON.parse(abs_resource_hosts)
+      hostname = hosts[0]["hostname"]
+      if hostname.nil? && hostname.empty?
```
johnduarte: :bug: this should be an OR not an AND condition.  I discovered this with the following modified spec test.
```
    context "when an invalid hosts JSON string is specified" do
      invalid_hosts = [
        nil,                      # nil case
        'Not valid JSON',         # not JSON
        '{}',                     # valid JSON, not hosts array
        '[{"hostname": ""}]',     # hostname.is_empty?
        '[{ "hostname": "legit" }, {} ]'  # invalid host[1]
      ]
      invalid_hosts.each do |i|
        it "raises an error" do
          expect { subject.parse_abs_resource_hosts(i) }
            .to raise_error(RuntimeError, /encountered parsing the hosts array/)
        end
      end
    end
```

setup/helpers/abs_helper.rb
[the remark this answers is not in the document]
billclaytor: I actually moved the hostname check to `parse_abs_resource_hosts` since parsing should fail if it doesn't exist. I wanted to differentiate between the 'invalid JSON' vs 'invalid abs_resource_hosts' cases and provide a corresponding error message, however I can consolidate if that seems unnecessary.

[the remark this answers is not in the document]
johnduarte: It does seem unnecessary to me, but if that is the desire, that is not what is happening now.

If I call `valid_abs_resource_hosts?('Foobar')`. I have passed an _invalid_ JSON string, but I get back the `encountered parsing the hosts array` error.
The same error is returned if I pass a _valid_JSON string ('{}')
However, if I pass `nil`, then I get the error `A valid JSON string is required...`.

So, I am confused about the intent of the error cases these two methods are trying to capture.

setup/helpers/abs_helper.rb
billclaytor: Updated to add tests for invalid key and empty value.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -1081,23 +1109,176 @@ class AbsHelperClass
...
       end
     end
   end
+
+  describe "#parse_abs_resource_hosts" do
+    before do
+      allow(subject).to receive(:puts)
+    end
+
+    context "when a valid hosts JSON string is specified" do
```
johnduarte: FYI, the simplest valid json which is also an invalid hosts string (`'{}'`) is not covered and will not throw the expected error.

billclaytor: Thanks; updated to raise an error and added an example to the spec.

util/abs/provision_pe_xl_nodes.rb
RandellP: sounds good.  I added note to one of the tickets that possibly more info could be in the inventory file, to reduce the need for hostnames to be in the params.  The inventory file got a lot of love after this module was written...

PR #301, opened 2019-07-03
(SLV-365) Update provision_pe_xl_nodes script and abs_utils doc

billclaytor:
This update includes the following:
- create `nodes.yaml` using heredoc
- move `nodes.yaml` and `params.json` generation into separate methods
- remove `nodes.yaml` templates
- add `ha` parameter to `params.json` templates
- update `params_no_ha.json` template to remove replica roles
- update `abs_utils.md` to reflect script update and fix typos


util/abs/provision_pe_xl_nodes.rb
```diff
@@ -50,6 +70,45 @@
...
+#
+# @example
+#   hosts = provision_hosts_for_roles(roles)
+#   create_nodes_yaml(hosts, output_dir)
+def create_nodes_yaml(hosts, output_dir)
+  File.open("#{output_dir}/nodes.yaml", "w+") do |f|
+    f.puts NODES_YAML_START
+    hosts.each do |host|
+      hostname = host[:hostname]
+      f.puts "      - #{hostname}"
```
johnduarte: Trying to manage the string list of nodes seems a bit brittle.  I am wondering if this could be handled by using YAML's [flow collection sequence](https://yaml.org/spec/1.2/spec.html#id2790088) and [anchor/alias](https://yaml.org/spec/1.2/spec.html#id2786196) facilities to make this more robust.

Rather than appending each hostname, include an anchor/alias in the template and use this method to populate the anchor.
Template:
```
NODES_YAML_START = <<~NODES
  ---
  nodes_anchor: &nodes_array []
  groups:
    - name: pe_xl_nodes
      config:
        transport: ssh
        ssh:
          host-key-check: false
          user: centos
          run-as: root
          tty: true
      nodes: *nodes_array
NODES
```

This method
```
def create_nodes_yaml(hosts, output_dir)
  File.open("#{output_dir}/nodes.yaml", "w+") do |f|
    nodes_yaml = NODES_YAML_START.gsub(/^nodes_anchor:.*$/, "nodes_anchor: #{hosts.to_s}")
    f.puts nodes_yaml
  end
end
```

I haven't tried this. Just wondering if it would prevent chasing down white-space issues in the future.

Thoughts?

billclaytor: I think it would be simpler to just add the nodes array:

```
NODES_YAML = <<~NODES_YAML
  ---
  groups:
    - name: pe_xl_nodes
      config:
        transport: ssh
        ssh:
          host-key-check: false
          user: centos
          run-as: root
          tty: true
NODES_YAML
```

```
def create_nodes_yaml(hosts, output_dir)
  yaml = YAML.safe_load NODES_YAML
  nodes = []

  hosts.each do |host|
    nodes << host[:hostname]
  end

  yaml["groups"][0]["nodes"] = nodes
  File.write("#{output_dir}/nodes.yaml", YAML.dump(yaml))
end
```

I added this for debugging:
```
def check_nodes_yaml(output_dir)
  file = "#{output_dir}/nodes.yaml"
  puts "Checking #{file}"

  yaml = YAML.load_file file
  nodes = yaml["groups"][0]["nodes"]

  puts
  puts "nodes:"
  puts nodes
  puts
end
```

johnduarte: +1

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -87,13 +87,36 @@ def create_pe_xl_bolt_files(hosts, output_dir)
...
+
+  yaml["groups"][0]["nodes"] = nodes
+  File.write("#{output_dir}/nodes.yaml", YAML.dump(yaml))
+end
+
+# Checks the nodes.yaml file to ensure it has been written correctly
+#
+# @author Bill Claytor
+#
+# @param [String] output_dir The directory where the file should be written
```
RandellP: Since you aren't outputting in this method, the name doesn't make sense.  Should be like yaml_dir or something.  And any reason to lock the name of the file to be nodes.yaml?  Like why not take a file path as an input and just check it.  Then this method isn't arbitrarily hardcoding the file name.

billclaytor: Updated

PR #303, opened 2019-07-12
(SLV-365) Update provision_pe_xl_nodes to use HEREDOC for params.json

billclaytor:
The previous proposed update to pe_xl used separate 'without_ha' versions of the plans with the replica host parameters removed. The updated implementation conditionally performs the HA steps within the existing plans which expect these parameters. This removes the need to use templates for `params.json` and makes it easier to include the 'template' as a `HEREDOC`. The templates directory has been removed.

Additionally, the option to specify a non-HA environment is now controlled with a constant rather than requiring the roles array constant to be updated.

~~TODO: update the abs_utils.md doc the with the above change~~


billclaytor:
@RandellP When making this update I did consider replacing the parameters with variables and added this todo before the heredoc:

`# TODO: update to use variables / symbols for all parameter values?`

I did use variables for `ha` and `version` since these are not based on the hosts array. However, because the other parameter names are currently derived from the roles for each host (which may or may not be present) I didn't immediately see a simpler approach without declaring variables for each parameter which would be more steps than the current approach.

PR #304, opened 2019-07-17
(SLV-504) Update the provision_pe_xl_nodes script to use OptionParser

billclaytor:
This update to the provision_pe_xl_nodes script adds the ability to specify options via OptionParser:
```
    -h, --help                       Display the help text
        --noop                       Run in no-op mode
        --test                       Use test data rather than provisioning hosts
        --ha                         Specifies that the environment should be set up for HA
    -i, --id ID                      The value for the AWS 'id' tag
    -o, --output_dir DIR             The directory where the Bolt files should be written
    -v, --pe_version VERSION         The PE version to install
    -t, --type TYPE                  The AWS EC2 instance type to provision
    -s, --size SIZE                  The AWS EC2 volume size to specify
```

Additional updates: 
- Updated `check_nodes_yaml` with more output
- Created `check_params_json`
- Updated `create_nodes_yaml` and `create_params_json` to call the corresponding 'check' method when running in test mode
 - Refactoring

Note:  Resolved the issue with OptionParser on Ruby 2.3.0 by not using the 'store options to a hash' feature:
https://ruby-doc.org/stdlib-2.6.3/libdoc/optparse/rdoc/OptionParser.html#class-OptionParser-label-Store+options+to+a+Hash
`end.parse!(into: options)`

~~TODO: Update abs_utils.md~~

### DIFF
--- a/docs/abs_utils.md
+++ b/docs/abs_utils.md
@@ -34,7 +34,7 @@ The script accepts the following options which override the default values:
     -h, --help                       Display the help text
         --noop                       Run in no-op mode
         --test                       Use test data rather than provisioning hosts
-        --ha                         Specifies that the environment should be set up for HA
+        --ha                         Deploy HA environment
     -i, --id ID                      The value for the AWS 'id' tag
     -o, --output_dir DIR             The directory where the Bolt files should be written
     -v, --pe_version VERSION         The PE version to install
@@ -45,6 +45,7 @@ The script accepts the following options which override the default values:
 #### Default values
 When run without specifying any options the script uses the following default values:
 
+* HA (--ha): false
 * AWS_TAG_ID (-i, --id): slv
 * OUTPUT_DIR (-o, --output_dir): ./
 * PE_VERSION (-v, --pe_version): 2019.1.0
@@ -60,6 +61,24 @@ Run the script from the `gatling-puppet-load-test` directory:
 bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb
 ```
 
+##### id
+Specify the value for the AWS 'id' tag (you will almost always want to specify a value to identify a set of hosts):
+```
+bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb -i non_ha_example
+```
+
+##### ha (and id)
+Deploy HA environment (false unless specified):
+```
+bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb --ha -i ha_example
+```
+
+##### pe_version
+Specify the PE version to install:
+```
+bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb -v 2018.1.3
+```
+
 ##### noop
 This option enables no-op mode which provides output but does not provision hosts or create files.
 ```
@@ -70,18 +89,6 @@ bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb --noop
 This option enables test mode which uses the included test host arrays to generate the Bolt files and verifies the output.
 ```
 bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb --test
-```
-
-##### id
-Specify a unique ID for each set of hosts:
-```
-bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb -i example
-```
-
-##### pe_version
-Specify the PE version to install:
-```
-bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb -v 2018.1.3
 ```
 
 ##### all options
--- a/docs/abs_utils.md
+++ b/docs/abs_utils.md
@@ -16,7 +16,7 @@ To avoid the proliferation of rake tasks the following Ruby scripts have been ad
 This script was created to assist in working with the [`pe_xl`](https://github.com/reidmv/reidmv-pe_xl) module.
 It provisions the nodes used by the module and generates the [Bolt](https://github.com/puppetlabs/bolt) [inventory](https://puppet.com/docs/bolt/latest/inventory_file.html) and [parameter](https://puppet.com/docs/bolt/latest/writing_tasks.html#concept-21) files populated with the provisioned hosts.
 
-EC2 hosts using the GPLT defaults (c5.2xlarge / 80GB ) are provisioned for the following roles:
+EC2 hosts are provisioned for the following roles:
 
 #### Core roles
 * master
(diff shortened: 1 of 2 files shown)


util/abs/provision_pe_xl_nodes.rb
```diff
@@ -1,9 +1,84 @@
...
+
+  # TODO: noop and test options are mutually exclusive;
+  # error when both are specified?
+  #
+  # omitted short versions to avoid collisions
+  opts.on("--noop", "Run in no-op mode") { options[:noop] = true }
+  opts.on("--test", "Use test data rather than provisioning hosts") { options[:test] = true }
+
+  # TODO: this description seems awkward; suggestions?
+  opts.on("--ha", "Specifies that the environment should be set up for HA") { options[:ha] = true }
```
johnduarte: Suggestion:
"Deploy HA environment"

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -61,6 +135,7 @@
 NODES_YAML
 
 # TODO: update to use variables / symbols for all parameter values?
+# TODO: should this use `<<-` vs `<<~`?
```
johnduarte: ABS: Always Be Squiggling!

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -164,30 +319,58 @@ def check_nodes_yaml(file)
...
+
+  puts "Parsing JSON..."
+  puts
+
+  json = JSON.parse contents
+  puts json
+  puts
+
+  puts "Verifying JSON..."
+  install = json["install"]
```
johnduarte: How is this a validation? If the JSON contents are an empty hash '{}', the results is:
```
Verified parameter 'install':
```
I would expect this function to ensure that the `install` key exists, that it's value is not empty, and possibly return a boolean.  Otherwise I have to know to really keep an eye on the output below to to the validation by eye manually.

billclaytor: Great point, thanks! I updated `check_params_json` and `check_nodes_yaml` to raise an error if the expected parameter does not match the expected value.

util/abs/provision_pe_xl_nodes.rb
billclaytor: Updated; removed comment :)

util/abs/provision_pe_xl_nodes.rb
billclaytor: Great suggestion; updated!

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -1,9 +1,83 @@
...
+    puts DESCRIPTION
+    puts opts
+    puts
+    exit
+  end
+
+  opts.on("--ha", "Specifies that the environment should be set up for HA")
+
+  # TODO: noop and test options are mutually exclusive;
+  # error when both are specified?
```
RandellP: are they?  can't you do a noop using the test data?

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -12,40 +86,39 @@
...
+else
+  HA = false
+  ROLES = ROLES_CORE
+end
 
-# TODO: update to use OptionParser
-# currently uses ARGV[0], ARGV[1] if specified, otherwise these defaults
-ABS_ID = "slv"
-OUTPUT_DIR = "./"
+NOOP = options[:noop] || false
```
RandellP: When I write defaults, I put them in constants, so that I can refer to them in the help text as well as use them in this section... then have one place to change them if I need to.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -1,9 +1,83 @@
...
+  * Core roles:
+   - master
+   - puppet_db
+   - compiler_a
+   - compiler_b
+
+  * HA roles:
+   - master_replica
+   - puppet_db_replica
+
```
RandellP: Need to describe the defaults here.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -85,6 +158,67 @@
...
+    Output directory for Bolt inventory and parameter files: #{OUTPUT_DIR}
+    PE version: #{PE_VERSION}
+    AWS EC2 id tag: #{AWS_TAG_ID}
+    AWS EC2 instance type: #{AWS_INSTANCE_TYPE}
+    AWS EC2 volume size: #{AWS_VOLUME_SIZE}
+
+PROVISION_MESSAGE
+
+NOOP_MESSAGE = "*** Running in no-op mode ***"
+NOOP_EXEC = <<~NOOP_EXEC
```
RandellP: Just a note.  The use of the noop mode here seems to be mainly to validate that the options got through.  Cause the output isn't substantially different than the input, and just tells the user it would call some internal methods with some options.  If that is the case... it may have been better as just spec tests...  It's fine, but next time that might be a better way to go if in this kind of case.  If the script does substantial processing of the options, that is when a noop mode makes sense, as a user might want to check that it did what they expected before running.

RandellP: BTW, the most common use of noop modes I have seen is when it will output actual commands it would run on the shell.

docs/abs_utils.md
```diff
@@ -28,54 +28,88 @@ EC2 hosts using the GPLT defaults (c5.2xlarge / 80GB ) are provisioned for the f
...
+When run without specifying any options the script uses the following default values:
 
-# NOTE: set HA to true for a HA environment
-HA = false
-ROLES = if HA
-          ROLES_CORE + ROLES_HA
-        else
-          ROLES_CORE
-        end
+* AWS_TAG_ID (-i, --id): slv
```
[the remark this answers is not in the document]
billclaytor: Updated

docs/abs_utils.md
```diff
@@ -28,54 +28,88 @@ EC2 hosts using the GPLT defaults (c5.2xlarge / 80GB ) are provisioned for the f
...
+bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb --test
 ```
 
-#### Optional arguments
-The script accepts the following optional arguments:
-* id - This value will be specified for the `id` tag when provisioning the EC2 instance.
-This can be helpful when filtering instances in the EC2 console.
-The default value is `slv`.
+##### id
+Specify a unique ID for each set of hosts:
```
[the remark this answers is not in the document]
billclaytor: Updated

docs/abs_utils.md
```diff
@@ -16,7 +16,7 @@ To avoid the proliferation of rake tasks the following Ruby scripts have been ad
 This script was created to assist in working with the [`pe_xl`](https://github.com/reidmv/reidmv-pe_xl) module.
 It provisions the nodes used by the module and generates the [Bolt](https://github.com/puppetlabs/bolt) [inventory](https://puppet.com/docs/bolt/latest/inventory_file.html) and [parameter](https://puppet.com/docs/bolt/latest/writing_tasks.html#concept-21) files populated with the provisioned hosts.
 
-EC2 hosts using the GPLT defaults (c5.2xlarge / 80GB ) are provisioned for the following roles:
+EC2 hosts are provisioned for the following roles:
```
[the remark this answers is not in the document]
billclaytor: Updated to list default, non-HA with id, HA with id.

util/abs/provision_pe_xl_nodes.rb
billclaytor: Well, you can specify them together:
```
bill.claytor:~/RubymineProjects/_forks/gatling-puppet-load-test> ruby util/abs/provision_pe_xl_nodes.rb --noop --test
*** Running in no-op mode ***
*** Running in test mode ***

Would have provisioned pe_xl nodes with the following options:
  HA: false
  Output directory for Bolt inventory and parameter files: ./
  PE version: 2019.1.0
  AWS EC2 id tag: slv
  AWS EC2 instance type: c5.2xlarge
  AWS EC2 volume size: 80

Would have called:

  hosts = provision_hosts_for_roles(["master", "puppet_db", "compiler_a", "compiler_b"],
                                    slv,
                                    c5.2xlarge,
                                    80)

to provision the hosts, then:

  create_pe_xl_bolt_files(hosts, ./)

to create the Bolt inventory and parameter files.
```

But the output is not correct because in test mode it would not have called `provision_hosts_for_roles`.

util/abs/provision_pe_xl_nodes.rb
billclaytor: Updated to output the defaults after the options since it seems awkward to list default values for the options without first describing them.

util/abs/provision_pe_xl_nodes.rb
billclaytor: Yes, these constants were originally the defaults. I tried to add the options with as little impact to the code as possible; since the constants were already being used in the code it seemed to make sense to leverage them rather than updating to use new variables. I have added the defaults as constants before the description (since they are now used in it).

util/abs/provision_pe_xl_nodes.rb
billclaytor: If it was implemented as spec tests the user wouldn't be able to see the output when trying out the script which was the intent. If this usage does not seem appropriate for `no-op` mode I can remove it.

docs/abs_utils.md
```diff
@@ -60,22 +61,16 @@ Run the script from the `gatling-puppet-load-test` directory:
 bundle exec ruby ./util/abs/provision_pe_xl_nodes.rb
```
RandellP: I would change the header from "default options" to "Most Common Execution".  And I think you commonly pass in the ticket number as the id don't you?  if so add that to the execution.  The idea is to get across to the reader the most common way the script it run... like giving them an opinionated way to run it.  The value here is if I read this in 6 months will I know how we usually run it?  From the current reading I might think that "default options" was just showing me the simplest execution, not the opinionated way to run it.  Does that make sense?

billclaytor: I think it seems odd to show an example of running it with options before showing an example of running it without. This is why I listed the example with no options specified (i.e. default options) first. 

I've updated the section to add a 'Common usage scenarios' heading and more descriptive text.

PR #305, opened 2019-07-17
(maint) Ignore orch mem usage for lovejoy

RandellP:
The mem usage more than doubled but is execpted.

### DIFF
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -816,6 +816,12 @@ module PerfRunHelper
     # global atop results and per process atop results
     baseline = get_baseline_result
     baseline&.each do |key, value|
+      # orchestration has a jump in memory usage in 2019.2 due to plan executor
+      # https://tickets.puppetlabs.com/browse/BOLT-986
+      # Currently we don't have a way to whitelist that... so just turning off the check
+      # Ticket to make a way https://tickets.puppetlabs.com/browse/SLV-515
+      next if key.to_s == "process_orchestration_services_release_avg_mem" && BEAKER_PE_VER =~ /^2019.2*/
+
       assert_value = if key.to_s == "avg_response_time"
                        gatling_result.avg_response_time.to_f
                      elsif key.to_s.start_with? "process"


tests/helpers/perf_run_helper.rb
```diff
@@ -816,6 +816,10 @@ def baseline_assert(atop_result, gatling_result)
     # global atop results and per process atop results
     baseline = get_baseline_result
     baseline&.each do |key, value|
+      # orchestration has a jump in memory usage in 2019.2 due to plan executor
+      # Currently we don't have a way to whitelist that... so just turning off the check
+      next if key.to_s == "process_orchestration_services_release_avg_mem" && BEAKER_PE_VER =~ /^2019.2*/
```
johnduarte: I presume this would be true for future versions as well.  Should the be generalized to accommodate any version greater than 2019.2.0?
Example using Gem module  which can compare semantic versions.
```
next if key.to_s == "process_orchestration_services_release_avg_mem" && (Gem::Version.new(BEAKER_PE_VER) >= Gem::Version.new('2019.2'))
```

tests/helpers/perf_run_helper.rb
```diff
@@ -816,6 +816,10 @@ def baseline_assert(atop_result, gatling_result)
     # global atop results and per process atop results
     baseline = get_baseline_result
     baseline&.each do |key, value|
+      # orchestration has a jump in memory usage in 2019.2 due to plan executor
```
johnduarte: Do we have a reference to this in docs, code, or jira tickets that can be included here?

tests/helpers/perf_run_helper.rb
RandellP: I don't want to blanket turn it off for future versions, as that would mean missing future increased drift.  One solution could be to update the % allowed after lovejoy is released to allow for this drift.  Though that would let smaller drifts through.  I filed https://tickets.puppetlabs.com/browse/SLV-515 (which I just added to the comment here) to hav a way to approve specific drifts, which would allow us to ignore this data drift on any 2019.2+ when compared to pre 2019.2 releases.  So basically with this line I am kicking the can down the road, but making sure it pops up and reminds us to do something. :)

tests/helpers/perf_run_helper.rb
RandellP: done

RandellP:
@billclaytor alright, John approved so it is on you...

PR #306, opened 2019-07-22
(SLV-526) Update spec tests to use simplecov for code coverage

billclaytor:
This update adds simplecov for spec test code coverage. A `spec_helper.rb` file has been added which configures simplecov and requires the files that should be covered by the spec tests. This currently includes `setup/helpers` and `tests/helpers`. The spec test files have been updated to require `spec_helper` instead of the files under test. 

After running the spec tests via `bundle exec rake spec` the coverage report will be displayed in the results:
```
Coverage report generated for Unit Tests to /Users/test.user/gatling-puppet-load-test/coverage. 515 / 1185 LOC (43.46%) covered.
```
An HTML report is generated at `./coverage/index.html`. Based on the current code coverage the threshold has been set to 40%.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -15,17 +15,6 @@ Dir["./setup/helpers/*.rb"].each { |file| require file }
 # test helpers
 Dir["./tests/helpers/*.rb"].each { |file| require file }
 
-RSpec.configure do |c|
-end
-
-RSpec.shared_context "case_info_lets" do
-  let(:something) { "some value" }
-end
-
-def do_something_helpful(value)
-  puts "Do something with #{value}"
-end
-
 SimpleCov.at_exit do
   SimpleCov.result.format!
   SimpleCov.minimum_coverage 40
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+# setup helpers
+Dir["./setup/helpers/*.rb"].each { |file| require file }
+
+# test helpers
+Dir["./tests/helpers/*.rb"].each { |file| require file }
+
+RSpec.configure do |c|
+end
+
+RSpec.shared_context "case_info_lets" do
```
johnduarte: Is this example code necessary?  This feels like a block of copy/pasta.

spec/spec_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:something) { "some value" }
+end
+
+def do_something_helpful(value)
```
johnduarte: Is this example code necessary?  This method feels like a copy/pasta example.

spec/spec_helper.rb
billclaytor: Updated to remove examples

spec/spec_helper.rb
billclaytor: Updated to remove examples

PR #309, opened 2019-08-02
(SLV-430) Added script for creating tar file

RandellP:
Find all the files in the metrics dir that are relevant to a time window and tars them up into a single tar file.
This is part 1.  Part 2 will be to boltify the running of this, and downloading of the resulting file.

Also, told rubocop to stop complaining about explicit return statements.


util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
+# frozen_string_literal: true
```
billclaytor: This script should probably be in a new sub-directory (i.e. 'metrics').

RandellP: The location is temporary.  It will either become some kind of task or part of a plan, and moved to whatever is the right place for it in a future PR.

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+    * tar_file_name (-t, --tar_file_name): Built from start and end time. Ex: #{TAR_FILE_NAME_DEFAULT}
+    * poll_interval (-p, --poll_interval): #{POLL_INTERVAL_DEFAULT}
+    * archive_interval (-a, --archive_interval): #{ARCHIVE_INTERVAL_DEFAULT}
+    * verbose (-v, --verbose): False
+DEFAULTS
+
+options = {}
+
+OptionParser.new do |opts|
+  opts.banner = "Usage: provision_pe_xl_nodes.rb [options]"
```
billclaytor: This should be "collect_metrics_files.rb"

RandellP: update pushed.

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+    options[:end_time] = end_time
+  end
+  opts.on("-m", "--metrics_dir dir_path", String, "Top level metrics directory") do |metrics_dir|
+    options[:metrics_dir] = metrics_dir
+  end
+  opts.on("-o", "--output_dir dir_path", String,
+          "Directory for the collected files and resulting tarfile") do |output_dir|
+    options[:output_dir] = output_dir
+  end
+  opts.on("-t", "--tar_file_name filename", String,
```
billclaytor: Should this automatically add `.tar.gz` if it isn't specified, or assume the user knows what they're doing?

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+def tar_metrics_files
+  v = (VERBOSE ? "v" : "")
+  comm = "tar cz#{v}f #{TAR_FILE_NAME} -C #{OUTPUT_DIRECTORY} ."
+  puts "Tarring metrics with: #{comm}" if VERBOSE
+  system(comm)
+end
+
+process_metrics_dir(METRICS_DIRECTORY)
+tar_metrics_files
+puts "Collected the following tar file with the metrics files: #{TAR_FILE_NAME}"
```
billclaytor: This message is displayed even if the operation was not successful (i.e. if the specified tar file is actually a directory):
```
[root@ip-10-227-2-145 ~]# mkdir metrics
[root@ip-10-227-2-145 ~]# ruby collect_metrics_files.rb -t ~/metrics
tar (child): /root/metrics: Cannot open: Is a directory
tar (child): Error is not recoverable: exiting now
Collected the following tar file with the metrics files: /root/metrics
```

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+#
+# @author Randell Pelak
+#
+# @return [void]
+#
+# @example
+#   tar_metrics_files
+def tar_metrics_files
+  v = (VERBOSE ? "v" : "")
+  comm = "tar cz#{v}f #{TAR_FILE_NAME} -C #{OUTPUT_DIRECTORY} ."
```
billclaytor: Even when an output directory is specified the tar file is written to the current directory:
```
[root@ip-10-227-2-145 ~]# ls
anaconda-ks.cfg  atop_log_soak_json.log  atop_log_warmupjit_json.log  collect_metrics_files.rb  original-ks.cfg  tune_output
[root@ip-10-227-2-145 ~]# mkdir metrics
[root@ip-10-227-2-145 ~]# ruby collect_metrics_files.rb -o ~/metrics
Collected the following tar file with the metrics files: 20190805T085524Z-20190805T165524Z.tar.gz
[root@ip-10-227-2-145 ~]# ls
20190805T085524Z-20190805T165524Z.tar.gz  atop_log_soak_json.log       collect_metrics_files.rb  original-ks.cfg
anaconda-ks.cfg                           atop_log_warmupjit_json.log  metrics                   tune_output
[root@ip-10-227-2-145 ~]# ls metrics
orchestrator  puppetdb  puppetserver
```

It seems like it would be useful if it created a sub-directory within the output directory named with either the default or specified target file name (e.g. '20190805T085524Z-20190805T165524Z' and copied the service directories there. Then that directory would be tarred with the tar file being created in the output directory. For example:
```
[root@ip-10-227-2-145 ~]# ls
anaconda-ks.cfg  atop_log_soak_json.log  atop_log_warmupjit_json.log  collect_metrics_files.rb  metrics  original-ks.cfg  tune_output
[root@ip-10-227-2-145 ~]# ls metrics
20190805T090011Z-20190805T170011Z  20190805T090011Z-20190805T170011Z.tar.gz
[root@ip-10-227-2-145 ~]# ls metrics/20190805T090011Z-20190805T170011Z
orchestrator  puppetdb  puppetserver
```

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+
+DESCRIPTION = <<~DESCRIPTION
+  This script is designed to gather up the tar and json files from the metrics_collector module
+  within a specified time frame.  Going as far as opening up a tar file and extracting
+  only the matching jsons.
+DESCRIPTION
+
+DEFAULTS = <<~DEFAULTS
+  The following defaults values are used if the options are not specified:
+    * start (-s, --start): Now minus 8 hours. Ex: #{START_TIME_DEFAULT}
```
billclaytor: The start and end options don't have short forms:
```
  opts.on("--start seconds", Integer,
          "Epoch time (seconds) that marks the start of the collection window") do |start_time|
    options[:start_time] = start_time
  end
  opts.on("--end seconds", Integer,
          "Epoch time (seconds) that marks the end of the collection window") do |end_time|
    options[:end_time] = end_time
  end
```

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+    * end (-e, --end): Now. Ex: #{END_TIME_DEFAULT}
+    * metrics_dir (-t, --metrics_dir): #{METRICS_DIRECTORY_DEFAULT}
```
billclaytor: The short form for 'metrics_dir' is '-m'.

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+    options[:start_time] = start_time
+  end
+  opts.on("--end seconds", Integer,
+          "Epoch time (seconds) that marks the end of the collection window") do |end_time|
+    options[:end_time] = end_time
+  end
+  opts.on("-m", "--metrics_dir dir_path", String, "Top level metrics directory") do |metrics_dir|
+    options[:metrics_dir] = metrics_dir
+  end
+  opts.on("-o", "--output_dir dir_path", String,
```
billclaytor: It would be helpful if the output directory would be created if it doesn't exist, if not by default then perhaps a -f / --force option.

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+#
+# @return [void]
+#
+# @example
+#   process_service_dir(dir_path, service)
+# TODO: Would be nice to only make the dir if relevant files were found.  And only do it once.
+def process_service_dir(dir_path, service)
+  service_staging_dir = "#{OUTPUT_DIRECTORY}/#{service}"
+  FileUtils.mkdir(service_staging_dir) unless File.directory?(service_staging_dir)
+  service_dir = Dir.open dir_path
```
johnduarte: Sorry, I could not help myself from exploring an alternative implementation using minitar to search for and extract only the json files each tarball contains.

Here is an outline that gathers the json files from the found tarballs, extracts them into a temp directory and calls `process_json_file` on them.  This is not meant to advocate a re-implementation.  I am simply adding it here to consider a couple of things to possibly refactor.

1.  Let `Dir.glob` do the work of walking the file tree.
1.  Use minitar to only get what you need out of the tarball.
 
```
require 'zlib'
require 'minitar'

json_files = Dir.glob("#{service_dir}/**/*.json")
tar_files = Dir.glob("#{service_dir}/**/*.tar.gz")

json_files.each do |jsonfile|
  process_json_file(jsonfile, service)  
end

tar_files.each do |tarfile|
  # use minitar to get only the json files out of the archive
  zr = Zlib::GzipReader.new(File.open(tarfile, 'rb'))
  mtr = Archive::Tar::Minitar::Reader.new(zr)
  arch_jsons = mtr.collect { |x| x.full_name if x.file? && x.full_name.match(/\.json$/) }
  tmp_dir = Dir.mktmpdir
  Archive::Tar::Minitar.unpack(zr, tmp_dir, arch_jsons)
  arch_jsons.each do |a_json|
    process_json_file(File.join(tmp_dir, jsonfile), service)
  end
end
```

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
...
+OptionParser.new do |opts|
+  opts.banner = "Usage: collect_metrics_files.rb [options]"
+
+  opts.on("-h", "--help", "Display the help text") do
+    puts DESCRIPTION
+    puts opts
+    puts DEFAULTS
+    exit
+  end
+
```
billclaytor: It would be helpful to have the option to collect all of the available data. Using `--start 0` seems to work but results in a name like '19700101T000000Z-20190805T192726Z' rather than the actual start of the data.

It would also be helpful to specify --start and --end with the format used in the default name so that a user could specify a date / time range without converting to unix format.

util/collect_metrics_files.rb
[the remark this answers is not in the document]
johnduarte: Script is meant to be run on a gplt master.  The primary intended use pattern currently is running it with bolt via:
```
bolt script run util/collect_metrics_files.rb -n mymaster -u root --no-host-key-check
```

@RandellP can you add a short docs/collect_metrics_files.md file to at least capture this intended execution pattern?  Maybe you can include a couple of sentences about its future as a bolt task/plan component.

util/collect_metrics_files.rb
```diff
@@ -0,0 +1,253 @@
+# frozen_string_literal: true
+
+require "optparse"
+require "time"
+require "tmpdir"
+
+END_TIME_DEFAULT = Time.now.to_i
+START_TIME_DEFAULT = END_TIME_DEFAULT - (8 * 60 * 60) # minus 8 hours
+# TODO: can we get this from the module, it is configurable
```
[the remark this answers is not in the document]
johnduarte: The intent is to include this script in a stand-alone module that gets added to gplt.
@RandellP as short blurb about your roadmap vision in the docs would be great.

util/collect_metrics_files.rb
RandellP: Filed SLV-539 on this.

RandellP:
@billclaytor @johnduarte I totally refactored the script to be a class.  And made a lot of the changes requested.  What I haven't done is create any documentation, or the spec tests.  
I also moved the script into a metrics module which should be it's final home for use in a plan.
We can review and merge when ready this PR while I work on the spec tests next and have them as a separate PR.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,377 @@
...
+    #
+    # @return [boolean] true if yes, false if no
+    #
+    # @example
+    #   json_file_from_target_window(filename)
+    def json_file_from_target_window?(filename)
+      time_str = filename.gsub(/.json$/, "")
+      time_obj = DateTime.parse(time_str)
+      epoch = time_obj.to_time.to_i
+      return true if epoch > @start_epoch && (epoch - @poll_interval) < @end_epoch
```
johnduarte: The comparison operation will return the needed boolean.  There is no need to have two return statements, just...
```
return epoch > @start_epoch && (epoch - @poll_interval) < @end_epoch
```

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,377 @@
...
+    # @author Randell Pelak
+    #
+    # @param [string] file_path absolute path to json file
+    #
+    # @return [void]
+    #
+    # @example
+    #   consider_json_for_inclusion(file_path)
+    def consider_json_for_inclusion(file_path)
+      filename = File.basename(file_path)
```
johnduarte: You have split the filename manipulation between two methods (this one and `json_file_from_target_window`) to get what you want.  I would delete this line and incorporate it into the other method.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,377 @@
...
+    # @author Randell Pelak
+    #
+    # @param [string] filename json filename
+    #
+    # @return [boolean] true if yes, false if no
+    #
+    # @example
+    #   json_file_from_target_window(filename)
+    def json_file_from_target_window?(filename)
+      time_str = filename.gsub(/.json$/, "")
```
johnduarte: Rather than expecting just a filename and gsubbing it, you can isolate the filename without the suffix using `File.basename`.  The input should also be validated to be a json file.

```
raise ArgumentError, "Argument is not a json filename" unless filename.match(/\.json$/i)
time_str = File.basename(filename, ".json")
```

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,377 @@
...
+    #
+    # @param [string] file_path absolute path to json file
+    #
+    # @return [void]
+    #
+    # @example
+    #   stage_json(file_path)
+    def stage_json(file_path)
+      filename = File.basename(file_path)
+      file_dir = File.dirname(file_path)
```
johnduarte: FYI, the [split method](https://ruby-doc.org/core-2.6.3/File.html#method-c-split) will do this in one call.
```
file_dir, filename = File.split(file_path)
```

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,377 @@
...
+    # @return [boolean] true if yes, false if no
+    #
+    # @example
+    #   tarfile_from_target_window?(filename, service)
+    def tarfile_from_target_window?(filename, service)
+      time_str_with_extension = filename.gsub(/^#{service}-/, "")
+      time_str = time_str_with_extension.gsub(/.tar.gz$/, "")
+      time_obj = DateTime.strptime(time_str, "%Y.%m.%d.%H.%M.%S")
+      epoch = time_obj.to_time.to_i
+      return true if (epoch - @archive_interval) > @start_epoch && epoch < @end_epoch
```
johnduarte: Can you assign the results of these comparisons to variables and 
```
return this || that
```
1. I don't understand the distinction between this condition and the next. So, names might help be understand.
2. It would reduce the return to one call.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,377 @@
...
+    # Tars up the resulting metrics files for easy transport off the system
+    #
+    # @author Randell Pelak
+    #
+    # @return [void]
+    #
+    # @example
+    #   tar_metrics_files
+    def tar_metrics_files
+      v = (@verbose ? "v" : "")
```
johnduarte: What is the point of this ternary call?  You use the `@verbose` variable below.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
RandellP: That raise caught a bug that tar files where getting passed to it.  Fixing that bug also sped it up a lot.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
RandellP: Turns out these were a distinction without a difference after the refactor.  Previously I cared if the tar file was completely inside the time window, as I would just take the whole tar file as was.  But since I am now opening them up and checking the files inside no matter if it was completely in the window or partially in the window I don't need the difference.  So I can just use the last if by itself.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
RandellP: The tar command uses the v variable.  This was to avoid duplicating the rest of the command in an if verbose block.  I feel like there is a better way to do it, but I couldn't think of one.

RandellP:
@johnduarte pushed an update based on your feedback...

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
johnduarte: Oh, I see it now. Thanks for the explanation.  I completely missed the variable embedded in the flag below.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,370 @@
...
+    # Determines if a json file is from the target window
+    #
+    # @author Randell Pelak
+    #
+    # @param [string] file_path json file_path
+    #
+    # @return [boolean] true if yes, false if no
+    #
+    # @example
+    #   json_file_from_target_window(file_path)
```
johnduarte: missing '?' in method call.

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,370 @@
...
+    #
+    # @author Randell Pelak
+    #
+    # @param [string] file_path absolute path to tar file
+    # @param [string] service name of the service the file comes from
+    #
+    # @return [void]
+    #
+    # @example
+    #   process_tar_file(file_path, service)
```
johnduarte: wrong method call in `@example`

Do we need these `@example`'s if they are only duplicating the method signature itself?  I don't see what value they add.

Boltdir/site-modules/metrics/spec/metrics_spec_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+
+SimpleCov.start do
+  add_filter "/spec/"
+  add_group "metrics", "files"
+end
+
+puts "yo"
+# files
+Dir["./Boltdir/site-modules/metrics/files/*.rb"].each { |file| require file }
+Dir["./Boltdir/site-modules/metrics/files/*.rb"].each { |file| puts "#{file}" }
```
johnduarte: Rubocop no likey.

s/puts "#{file}"/puts file/

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,370 @@
...
+  This script is designed to gather up the json files from the puppet_metrics_collector module
+  within a specified time frame.  Going as far as opening up the daily tar files and extracting
+  only the matching jsons.
+DESCRIPTION
+
+DEFAULTS = <<~DEFAULTS
+  The following defaults values are used if the options are not specified:
+    * start (-s, --start): Now minus 8 hours. Ex: #{START_EPOCH_DEFAULT}
+    * end (-e, --end): Now. Ex: #{END_EPOCH_DEFAULT}
+    * metrics_dir (-t, --metrics_dir): #{METRICS_DIR_DEFAULT}
```
billclaytor: This should be "-m".

Boltdir/site-modules/metrics/files/collect_metrics_files.rb
```diff
@@ -0,0 +1,370 @@
...
+
+DESCRIPTION = <<~DESCRIPTION
+  This script is designed to gather up the json files from the puppet_metrics_collector module
+  within a specified time frame.  Going as far as opening up the daily tar files and extracting
+  only the matching jsons.
+DESCRIPTION
+
+DEFAULTS = <<~DEFAULTS
+  The following defaults values are used if the options are not specified:
+    * start (-s, --start): Now minus 8 hours. Ex: #{START_EPOCH_DEFAULT}
```
billclaytor: These should be "--start_epoch" and "--end_epoch".

PR #310, opened 2019-08-09
(SLV-430) Create rake task for PR test

RandellP:
updated rakefile to have a rubocop task and to namespace the spec test task.
test:check
link:rubocop
Also added pr:check which runs both.
Then updated travis.yml to call the new PR task.


RandellP:
@billclaytor made a change to make it run both even if one fails.

PR #313, opened 2019-08-16
(SLV-434) Use code manager for PE installs

johnduarte:
This PR updates the beaker pre-suite and options files to use code manager rather than r10k when installing PE.

### DIFF
--- a/setup/helpers/perf_helper.rb
+++ b/setup/helpers/perf_helper.rb
@@ -653,6 +653,28 @@ module PerfHelper
     )
   end
 
+  def configure_code_manager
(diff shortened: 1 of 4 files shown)


billclaytor:
I ran `performance_setup` and it seemed successful, and the code manager config in the console seemed correct. However, when I attempt to deploy on the master I get the following:
```
[root@ip-10-227-1-29 ~]# puppet-code deploy production --wait
Error: Code Manager requires a token, please use `puppet access login` to generate a token
```

It looks like an additional step is needed to generate a token.

PR #314, opened 2019-08-21
(SLV-430) Move collect_metrics_files out of bolt

RandellP:
turned out that writing a plan to run the script and get the file is not possible with bolt as is.  So pivoting to make it a normal util script
This just moves it and it files, as well as adjusts spec setup to work in the new place.


billclaytor:
Does the following comment in the script still apply?
https://github.com/puppetlabs/gatling-puppet-load-test/blob/d62d5aefe6277db7879dde98b51c838a19ad46d3/util/metrics/collect_metrics_files.rb#L14
```
The long term use case is with puppet bolt. But it isn't ready yet.  Like this
```

PR #321, opened 2019-08-29
(SLV-533) Update provision_pe_xl_nodes for large

johnduarte:
This PR updates the provision_pe_xl_nodes script to make it compatible with recent updates to the reidmv-pe_xl module.

* Add --arch option to enable defining a Large or Extra Large architecture.
* Remove the "ha" parameter in the params.json
* Process the JSON node parameters based on nodes provisioned.

### DIFF
--- a/docs/abs_utils.md
+++ b/docs/abs_utils.md
@@ -19,10 +19,13 @@ It provisions the nodes used by the module and generates the [Bolt](https://gith
 EC2 hosts are provisioned for the following roles:
 
 #### Core roles
+* metrics
 * master
-* puppet_db
 * compiler_a
 * compiler_b
+
+#### Extra Large Architecture roles
+* puppet_db
 
 #### HA roles
 * master_replica
@@ -35,6 +38,7 @@ The script accepts the following options which override the default values:
         --noop                       Run in no-op mode
         --test                       Use test data rather than provisioning hosts
         --ha                         Deploy HA environment
+    -a, --ref_arch REF_ARCH          Type of reference architecture to deploy (l, xl)
     -i, --id ID                      The value for the AWS 'id' tag
     -o, --output_dir DIR             The directory where the Bolt files should be written
     -v, --pe_version VERSION         The PE version to install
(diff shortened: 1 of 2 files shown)


util/abs/provision_pe_xl_nodes.rb
```diff
@@ -8,6 +8,8 @@
 require "./setup/helpers/abs_helper.rb"
 include AbsHelper # rubocop:disable Style/MixinUsage
 
+ARCH = { large: "l", extra_large: "xl" }.freeze
```
RandellP: We often refer to it as the ref arch.  So maybe make this REF_ARCH_TYPES  and DEFAULT_REF_ARCH

johnduarte: :+1: I have updated to change the variables and option accordingly.

util/abs/provision_pe_xl_nodes.rb
```diff
@@ -37,6 +39,7 @@
 DEFAULTS = <<~DEFAULTS
 
   The following defaults values are used if the options are not specified:
+  * REF_ARCH (--ref_arch): #{DEFAULT_REF_ARCH}
```
billclaytor: Not sure if it is necessary to include the short forms here... if so we should add `-a`; if not `-i` and `-o` should be removed.

billclaytor:
Should the abs_utils doc be updated as part of this ticket or should a separate ticket be created?
https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/docs/abs_utils.md

util/abs/provision_pe_xl_nodes.rb
johnduarte: I have updated to include the `-a` short form.

johnduarte:
@billclaytor I have added an update to the abs_utils documentation.

docs/abs_utils.md
```diff
@@ -20,10 +20,12 @@ EC2 hosts are provisioned for the following roles:
 
 #### Core roles
 * master
-* puppet_db
 * compiler_a
 * compiler_b
```
RandellP: Metrics seems to be missing

PR #322, opened 2019-09-03
(SLV-355) Use CMF to get tarball of metrics

RandellP:
Added code to generate a start_epoch and end_epoch file with the epoch of the, you guessed it, start and end of the test.
Added code to use the epochs in those files as arguments to the collect_metrics_files script to generate a tarball of metrics, and scp it into the results dir.

### DIFF
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -42,7 +42,9 @@ module PerfRunHelper
     @atop_session_timestamp = start_monitoring(master, gatling_scenario, true, 30)
     create_perf_archive_root
     capture_current_tune_settings
+    generate_timestamp_file("start_epoch")
     execute_gatling_scenario(gatling_scenario, simulation_id, gatling_assertions)
+    generate_timestamp_file("end_epoch")
   end
 
   # Handles the perf results
@@ -106,6 +108,22 @@ module PerfRunHelper
 
     puts "Writing file: #{output_path}"
     File.write(output_path, settings_json)
+  end
+
+  # Write out the epoch into a file for info and use when gathering metrics
+  #
+  # @author Randell Pelak
+  #
+  # @return [void]
+  #
+  # @example:
+  #   generate_timestamp_file("start_epoch")
+  #   generate_timestamp_file("end_epoch")
+  #
+  def generate_timestamp_file(filename)
+    now = Time.now.to_i
+    output_path = "#{@archive_root}/#{filename}"
+    File.write(output_path, now)
   end
 
   # Iteratively run perf_setup while automatically scaling the number of agents
@@ -787,6 +805,15 @@ module PerfRunHelper
       perf.log_csv
       copy_archive_files
     end
+    scp_to(master, "util/metrics/collect_metrics_files.rb", "/root/collect_metrics_files.rb")
+    start_epoch = File.read("#{@archive_root}/start_epoch")
+    end_epoch = File.read("#{@archive_root}/end_epoch")
+    cmf_output = on(master,
+                    "env PATH=\"#{master['privatebindir']}:${PATH}\" \
+                    ruby /root/collect_metrics_files.rb --start_epoch #{start_epoch} --end_epoch #{end_epoch}")
+                 .output
+    filename = cmf_output.match(/\w+-\w+.tar.gz/)[0].to_s
+    scp_from(master, "/root/#{filename}", "#{@archive_root}/")
     [perf, GatlingResult.new(gatling_assertions, mean_response_time)]
   end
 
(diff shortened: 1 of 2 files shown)


RandellP:
FYI, pushed up incomplete to see if the spec tests pass.  They are failing on my machine but on a method I didn't touch, and there is no obvious connection to the code I did touch...  Also they started failing only after I ran the clean rake task...

RandellP:
So the spec test failure is only on my mac.  My testrunner doesn't have the issue, travis doesn't.  So something updated on my mac and killed it (it was working most of the way through my development but now even failed on a fresh checkout of the master branch).  So removed the do not merge and added reviewers.

tests/helpers/perf_run_helper.rb
```diff
@@ -787,6 +805,15 @@ def get_perf_result
...
     end
+    scp_to(master, "util/metrics/collect_metrics_files.rb", "/root/collect_metrics_files.rb")
+    on(master, "ruby /root/collect_metrics_files.rb")
+    start_epoch = File.read("#{@archive_root}/start_epoch")
+    end_epoch = File.read("#{@archive_root}/end_epoch")
+    cmf_output = on(master,
+                    "ruby /root/collect_metrics_files.rb --start_epoch #{start_epoch} --end_epoch #{end_epoch}")
+                 .output
+    filename = cmf_output.match(/\w+-\w+.tar.gz/)[0].to_s
+    scp_from(master, "/root/#{filename}", "#{@archive_root}/#{filename}")
```
johnduarte: I am curious as to why we are using `scp_from` here and `archive_file_from` via the `copy_archive_files` method above?

Should this patch be applied to the `copy_archive_files` method instead?  I am unclear as to the difference in context between this additional code and the code within the `if perf` block above.  I don't know where the `stop_monitoring` method comes from or what it does.

tests/helpers/perf_run_helper.rb
```diff
@@ -787,6 +805,15 @@ def get_perf_result
       perf.log_csv
       copy_archive_files
     end
+    scp_to(master, "util/metrics/collect_metrics_files.rb", "/root/collect_metrics_files.rb")
+    on(master, "ruby /root/collect_metrics_files.rb")
+    start_epoch = File.read("#{@archive_root}/start_epoch")
+    end_epoch = File.read("#{@archive_root}/end_epoch")
+    cmf_output = on(master,
+                    "ruby /root/collect_metrics_files.rb --start_epoch #{start_epoch} --end_epoch #{end_epoch}")
```
johnduarte: Is ruby in the PATH? Is it the expected ruby?  Should this be set to "env PATH=\"#{master['privatebindir']}:${PATH}\" ruby ..."?

tests/helpers/perf_run_helper.rb
```diff
@@ -787,6 +805,15 @@ def get_perf_result
       perf.log_csv
       copy_archive_files
     end
+    scp_to(master, "util/metrics/collect_metrics_files.rb", "/root/collect_metrics_files.rb")
```
billclaytor: Would it be easier to run this script via Bolt?

tests/helpers/perf_run_helper.rb
billclaytor: `stop_monitoring` is from beaker-benchmark:
https://github.com/puppetlabs/beaker-benchmark/blob/8ae906f432ad378e3b59ee16464fa7fcc9b4d9d8/lib/beaker-benchmark/helpers.rb#L81

tests/helpers/perf_run_helper.rb
RandellP: no, because bolt creates a tmp area and removes it after the task is done.  So if for any reason you wanted to run it again manually, the script would be gone.  And you would have to add an output dir arg to the script run (not a problem, just not easier).  And of course you add in the possibility that bolt changes or what not.

tests/helpers/perf_run_helper.rb
RandellP: Since we didn't install ruby on the master as far as I know.  I assume it is part of the puppet install process. I could change the path though just to be on the safe side.

tests/helpers/perf_run_helper.rb
RandellP: yeah, the whole stop monitoring thing is beaker_benchmark.  And I believe it could fail for it's own reasons... which I didn't want to block the collection of metrics.  I was actually tempted to move this code above the stop monitoring, but the metrics collection is more resilient to being done later if needed than the beaker_benchmark, so I didn't want to take the chance that this code failed and blocked beaker_benchmark from stopping atop...

tests/helpers/perf_run_helper.rb
RandellP: and you meant export not env right?
like
```
cmf_output = on(master,
                    "export PATH=\"#{master['privatebindir']}:${PATH}\";" +
                    " ruby /root/collect_metrics_files.rb --start_epoch #{start_epoch} --end_epoch #{end_epoch}")
                 .output
```

johnduarte: I meant env in order for it to be available for the life of the command not the life of the session.

RandellP: got it sorted an updated

PR #328, opened 2019-09-09
(SLV-565) Document REF_ARCH env var

johnduarte:
This commit adds the `REF_ARCH` environment variable to the list of
variables used in the NEW_README document.

### DIFF
--- a/docs/rake_tasks.md
+++ b/docs/rake_tasks.md
@@ -1,0 +1,102 @@
+# Rake Tasks
+
+This document provides further detail on the rake tasks provided as part of
+this repository.  It is intended to clarify usage and workflows for tasks whose
+complexity requires further explanation.
+
+## PE_XL
+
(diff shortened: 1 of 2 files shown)


docs/rake_tasks.md
```diff
@@ -0,0 +1,102 @@
...
+compatible and is intended only to be used for PE installations.
+
+
+### Deploy
+
+`pe_xl:deploy`
+: Clean; provision via ABS; run pe_xl plan.
+
+The `pe_xl:deploy` task is the provisioning task performed by performance rake
+tasks (i.e. `performance, `soak`, `autoscale`) when the `REF_ARCH` environment
```
billclaytor: The ending ` is missing after "`performance" which prevents "`soak`" and "`autoscale`" from rendering properly.

johnduarte: Thanks for catching this @billclaytor . I have updated

PR #335, opened 2019-09-17
(SLV-575) Add loadbalancer when Large arch

johnduarte:
This PR adds a loadbalancer node in front of the compile masters when using a Large Architecture setup.  This includes:
* Create host, and roles via provision_pe_xl
* Add pre-suite to configure loadbalancer and PE infrastructure to use it

### DIFF
--- a/setup/install_gatling/30_classification/60_classify_use_loadbalancer.rb
+++ b/setup/install_gatling/30_classification/60_classify_use_loadbalancer.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require File.expand_path("../../../setup/helpers/perf_helper", __dir__)
+Beaker::TestCase.class_eval { include PerfHelper }
+
+test_name "classify and use loadbalancer" do # rubocop:disable Metrics/BlockLength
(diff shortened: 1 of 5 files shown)


johnduarte:
This PR should not be merged until https://github.com/puppetlabs/puppetlabs-pe_perf_control_repo/pull/10 has been merged into the control repo

setup/install_gatling/30_classification/60_classify_use_loadbalancer.rb
```diff
@@ -0,0 +1,70 @@
...
+    add_loadbalancer_groups(loadbalancer, compile_master)
+  end
+
+  step "add loadbalancer to PE Infra Agent group" do
+    group_id = classifier.get_node_group_id_by_name("PE Infrastructure Agent")
+    classifier.pin_nodes(group_id, { nodes: [loadbalancer] }.to_json)
+  end
+
+  step "export haproxy balancer member resources" do
+    on(compile_master,
```
RandellP: Does this work is there is more than one host with the role compile_master?

johnduarte: Yes, beaker commands take host objects or arrays of host objects.

PR #337, opened 2019-09-23
(SLV-507) update save/load db docs

RandellP:
While attempting to complete SLV-507, got new advice on how to load dbs.
The original method worked for a DB from kearney getting loaded into Kearney, but not loading into lovejoy.

### DIFF
--- a/docs/load_save_dbs.md
+++ b/docs/load_save_dbs.md
@@ -1,8 +1,8 @@
 # Saving and loading database content
 
-In real world use the databases of a puppet install would already have a lot of data in them.  So accurate performance testing should start with data as well.  But a fresh install doesn't have any old data in the databases.  This doc describes how data can be saved from one install and loaded into another.
+In real world use the databases of a Puppet installation would already have a lot of data in them, so accurate performance testing should start with data as well.  However, a fresh installation doesn't have any old data in the databases.  This document describes how data can be saved from one installation and loaded into another.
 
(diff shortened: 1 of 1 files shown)


docs/load_save_dbs.md
```diff
@@ -2,7 +2,7 @@
...
+There are some catches though.  If you load the data, it will overwrite existing data.  Like any nodes already setup with the install or any classifications.  So it is best to do this as early as possible.  If you do it immediately after the master install you can run `puppet infrastructure configure` and it will pick the master back up.  Possibly other infrastructure components as well, I didn't test it.  Otherwise you can run the puppet agent on nodes to get them back into the system.  As for classification, I think you could simply not load the classification db and that might save those.
```
billclaytor:
```suggestion
There are some catches though. Loading the data will overwrite existing data, such as any nodes already setup with the install or any classifications, so it is best to do this as early as possible.  When loading the data immediately after the master install running `puppet infrastructure configure` will pick the master back up. Other infrastructure components my be included as well but this has not been tested. Otherwise you can run the puppet agent on nodes to get them back into the system.  As for classification, not loading the classification db might save those.
```

docs/load_save_dbs.md
```diff
@@ -2,7 +2,7 @@
 
 In real world use the databases of a puppet install would already have a lot of data in them.  So accurate performance testing should start with data as well.  But a fresh install doesn't have any old data in the databases.  This doc describes how data can be saved from one install and loaded into another.
```
billclaytor:
```suggestion
In real world use the databases of a Puppet installation would already have a lot of data in them, so accurate performance testing should start with data as well.  However, a fresh installation doesn't have any old data in the databases.  This document describes how data can be saved from one installation and loaded into another.
```

RandellP:
@billclaytor pulled in your changes, and squashed the commits to one commit.

PR #338, opened 2019-09-26
(SLV-574) Enable gatling to use loadbalancer

johnduarte:
This PR enabled gatling to target a load balancer if it is present and to set the puppetserver auth.conf files on compile masters to allow for gatling submissions to be allowed.

### DIFF
--- a/rakefile
+++ b/rakefile
@@ -588,8 +588,16 @@ namespace :pe_xl do
 
   file "#{BUILD_DIR}/pe_xl_plan_result" do
     puts "Running bolt plan"
-    system "bolt plan run pe_xl --debug --inventory #{BUILD_DIR}/nodes.yaml --params @#{BUILD_DIR}/params.json"
-    File.write("#{BUILD_DIR}/pe_xl_plan_result", $?)
+    cmd = %W[
+      bolt plan run pe_xl
+      --debug
+      --inventory #{BUILD_DIR}/nodes.yaml
+      --params @#{BUILD_DIR}/params.json
+    ].join(" ")
+    res = system cmd
+    res_str = "#{$?.to_i}\n#{res}\n"
+    File.write("#{BUILD_DIR}/pe_xl_plan_result", res_str)
+    raise "Plan for pe_xl failed: #{res_str}" if $?.to_i != 0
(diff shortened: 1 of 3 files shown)


RandellP:
I ran some tests.  Like I ran acceptance to make sure it still worked, and it did.  I tried to run acceptance on the Large ref arch, but that failed.  Though it wasn't due to any code changed here, and it has never been proven to work.  But I feel confident that it won't break anything we currently run.  So I approve.

johnduarte:
> I ran some tests. Like I ran acceptance to make sure it still worked, and it did. I tried to run acceptance on the Large ref arch, but that failed. Though it wasn't due to any code changed here, and it has never been proven to work. But I feel confident that it won't break anything we currently run. So I approve.

I am able to reproduce this failure using the acceptance test on a pe xl deploy with the following command.
```
bundle install rake pe_xl:deploy
PUPPET_GATLING_R10K_CONTROL_REPO=https://github.com/puppetlabs/puppetlabs-pe_perf_control_repo.git PUPPET_GATLING_R10K_BASEDIR=/etc/puppetlabs/code-staging/environments PUPPET_GATLING_R10K_ENVIRONMENTS=production  bundle exec beaker -h build/pe_xl/beaker.cfg  --debug --options setup/options/options_pe_xl.rb --tests tests/acceptance.rb
```

This failure is caused by `ruby` not being available on the PATH on the master host.  When beaker is used to install PE it assigns `privatebindir` to the directory used for the puppet vendored binaries.  Since beaker is not involved in installing PE in the pe_xl deploy, this value is empty, so the `ruby` command is not available to the command in the [perf_run_helper](https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/tests/helpers/perf_run_helper.rb#L847-L850) that is used to execute the `collect_metrics_files.rb` script.

If I add ruby to the path on the master host and re-run the test it passes.

johnduarte:
@RandellP I have added https://github.com/puppetlabs/gatling-puppet-load-test/pull/338/commits/6f89603a27de477e7ed5056ed62d01e3558dd68c to address the issue that you discovered.

PR #340, opened 2019-10-01
(SLV-589) Update collect_metrics_files to use replace underscores with hyphens in staging dir

billclaytor:
This update changes the name of the staging directory in the collect_metrics_files script from `puppet_metrics_collector` to `puppet-metrics-collector` for consistency. The methods that handle the metrics collector data in `perf_run_helper` and `perf_results_helper` have also been updated.

### DIFF
--- a/setup/install/00_pre_install/00_configuration_check.rb
+++ b/setup/install/00_pre_install/00_configuration_check.rb
@@ -6,9 +6,10 @@ test_name "check configuration" do
   end
 
   step "fail on presence of scale: logic: setting" do
-    if options[:scale] && options[:scale][:logic]
-      fail_test "The 'logic' setting in the 'scale' configuration section is obsolete." \
-        "Please use 'static_files' and 'dynamic_files' instead."
-    end
+    fail_msg = %w[
+      The logic setting in the scale configuration section is obsolete.
+      Please use static_files and dynamic_files instead.
+    ].join(" ")
+    fail_test fail_msg if options.dig(:scale, :logic)
   end
 end
--- a/setup/install/00_pre_install/00_configuration_check.rb
+++ b/setup/install/00_pre_install/00_configuration_check.rb
@@ -6,6 +6,9 @@ test_name "check configuration" do
   end
 
   step "fail on presence of scale: logic: setting" do
-    fail_test "The 'logic' setting in the 'scale' configuration section is obsolete. Please use 'static_files' and 'dynamic_files' instead." if options[:scale] && options[:scale][:logic] # rubocop:disable Metrics/LineLength
+    if options[:scale] && options[:scale][:logic]
+      fail_test "The 'logic' setting in the 'scale' configuration section is obsolete." \
+        "Please use 'static_files' and 'dynamic_files' instead."
+    end
   end
 end
--- a/tests/helpers/perf_results_helper.rb
+++ b/tests/helpers/perf_results_helper.rb
@@ -66,8 +66,7 @@ module PerfResultsHelper
                            "catalog",
                            "report"].freeze
 
-  # TODO: update to 'puppet-metrics-collector' (SLV-589)
-  PUPPET_METRICS_COLLECTOR_DIR_NAME = "puppet_metrics_collector"
+  PUPPET_METRICS_COLLECTOR_DIR_NAME = "puppet-metrics-collector"
 
   # Extract Gatling JSON data into a CSV file in the format used in our release test reports
   # and convert the CSV file to an HTML file for easy viewing
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -517,8 +517,8 @@ module PerfRunHelper
 
     # copy puppet-metrics-collector to scale results dir (this iteration) and parent dir (entire scale run)
     src = File.join(@archive_root, PUPPET_METRICS_COLLECTOR_DIR_NAME)
-    FileUtils.copy_entry src, "#{scale_result_dir}/#{PUPPET_METRICS_COLLECTOR_DIR_NAME}"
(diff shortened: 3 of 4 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -549,9 +522,9 @@ def copy_scale_results(scenario)
 
     # copy puppet-metrics-collector to scale results dir (this iteration) and parent dir (entire scale run)
     # TODO: rename dir to 'puppet-metrics-collector'
```
johnduarte: Delete this line

tests/helpers/perf_results_helper.rb
```diff
@@ -66,8 +66,7 @@ module PerfResultsHelper
                            "catalog",
                            "report"].freeze
 
-  # TODO: update to 'puppet-metrics-collector' (SLV-589)
-  PUPPET_METRICS_COLLECTOR_DIR_NAME = "puppet_metrics_collector"
+  PUPPET_METRICS_COLLECTOR_DIR_NAME = "puppet-metrics-collector"
```
johnduarte: Does the [spec test](https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/spec/tests/helpers/perf_results_helper_spec.rb#L399) need to be updated?

tests/helpers/perf_run_helper.rb
```diff
@@ -23,6 +23,7 @@ module PerfRunHelper
                                       else
                                         ["puppetserver"]
                                       end
+  PUPPET_METRICS_COLLECTOR_DIR_NAME = "puppet-metrics-collector"
```
johnduarte: Since PerfResultsHelper is included on line 11 and this constant is declared there, this line should not be needed.

tests/helpers/perf_run_helper.rb
```diff
@@ -549,9 +522,9 @@ def copy_scale_results(scenario)
 
     # copy puppet-metrics-collector to scale results dir (this iteration) and parent dir (entire scale run)
     # TODO: rename dir to 'puppet-metrics-collector'
-    src = "#{@archive_root}/puppet_metrics_collector"
-    FileUtils.copy_entry src, "#{scale_result_dir}/puppet_metrics_collector"
-    FileUtils.copy_entry src, "#{scale_results_parent_dir}/puppet_metrics_collector"
+    src = "#{@archive_root}/#{PUPPET_METRICS_COLLECTOR_DIR_NAME}"
+    FileUtils.copy_entry src, "#{scale_result_dir}/#{PUPPET_METRICS_COLLECTOR_DIR_NAME}"
+    FileUtils.copy_entry src, "#{scale_results_parent_dir}/#{PUPPET_METRICS_COLLECTOR_DIR_NAME}"
```
johnduarte: nitpick, it would be safer to use File.join for these, and since they are both variables now it would be a tiny bit tidier.
```
src = File.join(@archive_root, PUPPET_METRICS_COLLECTOR_DIR_NAME)
FileUtils.copy_entry src, File.join(scale_result_dir, PUPPET_METRICS_COLLECTOR_DIR_NAME)
FileUtils.copy_entry src, File.join(scale_results_parent_dir, PUPPET_METRICS_COLLECTOR_DIR_NAME)
```

setup/install/00_pre_install/00_configuration_check.rb
```diff
@@ -6,6 +6,9 @@
   end
 
   step "fail on presence of scale: logic: setting" do
-    fail_test "The 'logic' setting in the 'scale' configuration section is obsolete. Please use 'static_files' and 'dynamic_files' instead." if options[:scale] && options[:scale][:logic] # rubocop:disable Metrics/LineLength
+    if options[:scale] && options[:scale][:logic]
```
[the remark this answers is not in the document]
billclaytor: While I agree in principal, this file is only being updated to satisfy an updated rubocop rule that was already outside the scope of this ticket. Although the suggested updated would likely work just fine, my testing is already complete for the submitted code, and I wouldn't want to merge without re-testing, so I'm not sure that the ROI would be worth it in this case. However, if the team feels that this is a worthwhile change I am happy to implement it and re-test.

tests/helpers/perf_run_helper.rb
billclaytor: I had thought about it but still wasn't sure if we might want to copy the entire thing for some reason. I've now removed it since we can always go back and re-add it.

setup/install/00_pre_install/00_configuration_check.rb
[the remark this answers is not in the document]
johnduarte: You are perfectly correct @billclaytor . Thanks. :+1:

tests/helpers/perf_results_helper.rb
billclaytor: No, because I use `puppet_metrics_dir_name = PerfResultsHelper::PUPPET_METRICS_COLLECTOR_DIR_NAME` in the spec.

tests/helpers/perf_run_helper.rb
billclaytor: Good call; I will try removing it and re-test.

tests/helpers/perf_run_helper.rb
billclaytor: Nice; updated.

billclaytor:
Added DO NOT MERGE label for re-test after updates; will remove after testing.

setup/install/00_pre_install/00_configuration_check.rb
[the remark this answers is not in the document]
billclaytor: Since other changes should also be re-tested I added this as well. Thanks for the suggestion!

billclaytor:
Testing completed successfully; removed the DO NOT MERGE label.

PR #341, opened 2019-10-07
(SLV-512) Update rakefile to validate environment variables before executing performance tasks

billclaytor:
This update adds the `validate:performance_env_vars` as a pre-requisite for the `performance` and `performance_without_provision` tasks, and the `validate:beaker_install_type` as a pre-requisite for the `performance_provision_with_abs` task.

The `validate:validate_performance_env_vars` task currently invokes the `validate:baseline_pe_version` task which verifies the `BASELINE_PE_VER` environment variable if specified. Additional validations can be added here.

The `validate:beaker_install_type` task extracts the existing verification of `BEAKER_INSTALL_TYPE` and also verifies `BEAKER_PE_VER` and `BEAKER_PE_DIR` for 'pe', `PACKAGE_BUILD_VERSION` and `PUPPET_AGENT_VERSION` for 'foss'.

To avoid duplication the code to perform the BigQuery query has been extracted to `perf_results_helper.rb`; spec tests have been added.

### DIFF
--- a/rakefile
+++ b/rakefile
@@ -12,7 +12,7 @@ REF_ARCH_STD, REF_ARCH_LARGE = ["S","L"]
 task :default => :performance
 
 desc 'Provision systems using ABS (or use already set ABS_RESOURCE_HOSTS) and then execute beaker performance setup and tests'
-rototiller_task :performance do
+rototiller_task :performance => "validate:validate_performance_env_vars" do
   ENV['REF_ARCH'] ||= REF_ARCH_STD
   if ENV['PUPPET_GATLING_REPORTS_ONLY'] == 'true' && (!ENV['PUPPET_GATLING_REPORTS_TARGET'] || ENV['PUPPET_GATLING_REPORTS_TARGET'] == '')
     abort 'A valid result folder (i.e. PerfTestLarge-1524848074511) must be specified for PUPPET_GATLING_REPORTS_TARGET if PUPPET_GATLING_REPORTS_ONLY=true'
@@ -21,7 +21,7 @@ rototiller_task :performance do
   if ENV['REF_ARCH'] == REF_ARCH_LARGE
     Rake::Task["pe_xl:deploy"].invoke
   else
-    Rake::Task["performance_provision_with_abs"].execute unless ENV['ABS_RESOURCE_HOSTS'] ||
+    Rake::Task["performance_provision_with_abs"].invoke unless ENV['ABS_RESOURCE_HOSTS'] ||
       (ENV['BEAKER_HOSTS'] &&
           ENV['BEAKER_HOSTS'] != 'config/beaker_hosts/pe-perf-test.cfg' &&
           ENV['BEAKER_HOSTS'] != 'config/beaker_hosts/foss-perf-test.cfg' &&
@@ -43,7 +43,7 @@ rototiller_task :opsworks_performance do
   raise "The test must be Opsworks.rb, or '' if you plan to execute tests later" unless
     ["", "tests/OpsWorks.rb"].include? ENV["BEAKER_TESTS"]
   configure_opsworks
-  Rake::Task["performance"].execute
+  Rake::Task["performance"].invoke
 end
 
 def configure_opsworks
@@ -74,9 +74,9 @@ def configure_opsworks
 end
 
 desc 'Provision systems for the performance task with ABS - included in the performance task'
-rototiller_task :performance_provision_with_abs do |t|
-  raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
-      ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE'] == 'foss'
+rototiller_task :performance_provision_with_abs  => "validate:verify_beaker_install_type" do |t|
+  # raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
+  #     ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE'] == 'foss'
   t.add_env({:name => 'BEAKER_PE_DIR', :message => 'The PE download directory, example: http://enterprise.delivery.puppetlabs.net/archives/releases/2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
   t.add_env({:name => 'BEAKER_PE_VER', :message => 'The PE version to install on hosts, example: 2017.2.4'}) if ENV['BEAKER_INSTALL_TYPE'] == 'pe'
   t.add_env({:name => 'ABS_OS', :default => 'centos-7-x86-64-west', :message => 'The OS to provision with ABS: centos-7-x86-64-west (for AWS), or centos-7-x86_64 (for vmpooler)'})
@@ -102,7 +102,7 @@ rototiller_task :setup_recording_agent do
   ENV['ABS_RESOURCE_HOSTS'] = File.open("last_abs_resource_hosts.log").readlines[0] if File.exist? "last_abs_resource_hosts.log"
   ENV['BEAKER_PRE_SUITE'] = "setup/install_gatling/00_pre_install/05_initialize_helpers.rb,setup/install_gatling/setup_metrics_as_agent.rb"
   ENV['BEAKER_TESTS'] = ""
-  Rake::Task["performance_without_provision"].execute
+  Rake::Task["performance_without_provision"].invoke
 end
 
 desc 'Run Performance tests using previously provisioned hosts'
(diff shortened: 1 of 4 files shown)


rakefile
```diff
@@ -74,9 +74,9 @@ def configure_opsworks
 end
 
 desc 'Provision systems for the performance task with ABS - included in the performance task'
-rototiller_task :performance_provision_with_abs do |t|
-  raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
-      ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE'] == 'foss'
+rototiller_task :performance_provision_with_abs  => "validate:verify_beaker_install_type" do |t|
+  # raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
+  #     ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE'] == 'foss'
```
johnduarte: Delete commented lines that have been pulled into validate:verify_beaker_install_type

rakefile
```diff
@@ -612,4 +612,44 @@ namespace :pe_xl do
...
+end
+
+namespace :validate do
+  desc "Validate the environment variables"
+  task :validate_performance_env_vars do
+    Rake::Task["bigquery:verify_baseline_version"].invoke
+  end
+
+  desc "Verifies that the BEAKER_INSTALL_TYPE environment variable and dependencies have been specified"
+  task :verify_beaker_install_type do
```
johnduarte: From a ux perspective, I would prefer that we standardize on "validate" or "verify" and use that prefix as the namespace.  The prefix should then not be used in the namespaced task name.  This would enable me to execute the tasks relative to the namespace.

E.G.
```
bundle exec rake validate:performance_env_vars
bundle exec rake validate:beaker_install_type
bundle exec rake validate:baseline_version
```

rakefile
```diff
@@ -74,9 +74,9 @@ def configure_opsworks
 end
 
 desc 'Provision systems for the performance task with ABS - included in the performance task'
-rototiller_task :performance_provision_with_abs do |t|
-  raise "The BEAKER_INSTALL_TYPE env var must be set to either 'pe' or 'foss' to continue" unless
-      ENV['BEAKER_INSTALL_TYPE'] == 'pe' || ENV['BEAKER_INSTALL_TYPE'] == 'foss'
+rototiller_task :performance_provision_with_abs  => "validate:verify_beaker_install_type" do |t|
```
RandellP: I believe you said this had to happen here to get in front of pe_xl doing the install.  Since it is not intuitive to think this needs to come before provisioning add a comment explaining that.

rakefile
```diff
@@ -102,19 +102,19 @@ rototiller_task :setup_recording_agent do
...
   ENV['BEAKER_HOSTS'] ||= "log/latest/hosts_preserved.yml"
   ENV['ABS_RESOURCE_HOSTS'] = File.open("last_abs_resource_hosts.log").readlines[0] if File.exist? "last_abs_resource_hosts.log"
   ENV['BEAKER_PRE_SUITE'] = ""
-  Rake::Task["performance_without_provision"].execute
+  Rake::Task["performance_without_provision"].invoke
 end
 
 desc 'Execute beaker performance setup and tests against existing hosts specified by ABS_RESOURCE_HOSTS env var, intended to be used in CI - use task "performance" for local/dev execution'
-rototiller_task :performance_without_provision do |t|
+rototiller_task :performance_without_provision => "validate:validate_performance_env_vars" do |t|
```
RandellP: If I run the performance task it depends on validate_performance_env_vars and then calls this task... will it do teh validation twice?

rakefile
```diff
@@ -612,4 +612,44 @@ namespace :pe_xl do
...
+  task :baseline_versions do
+    versions = baseline_versions
+    puts "Baseline versions:"
+    puts versions
+  end
+
+  desc "Verify that the specified baseline PE version exists in BigQuery"
+  task :verify_baseline_version do
+    # if BASELINE_PE_VER is specified verify that it exists before proceeding
+    verify_baseline_version ENV["BASELINE_PE_VER"] if ENV["BASELINE_PE_VER"]
```
RandellP: Seems like this shouldn't have an if...  cause if verify_baseline_version is called, it is implied that there is a version to verify.  The if should be on the thing that calls this task.  This may seem small, and it kinda is... but imagine a task that required a baseline_pe_ver.  It would have to do an if env var set check and this verify... where as if the if were not here, it could just call this verify and get a fail if unset or if not verified.  That makes the task more flexible and true to it's name.

tests/helpers/perf_results_helper.rb
```diff
@@ -757,5 +760,73 @@ def extract_puppetserver_metrics_from_json(file)
...
+    versions
+  end
+
+  # Verify that the specified baseline PE version exists in BigQuery
+  #
+  # @author Bill Claytor
+  #
+  # @param [String] baseline_pe_ver The baseline PE version to verify
+  #
+  # @raise [StandardError] If the specified baseline PE version is not found
```
RandellP: I feel like this shouldn't raise an error.  It should be called valid_baseline_version?  and return true or false.  The caller should decide what to do with the information.  Generally the caller will have better information to include in the error as well... like the source of the version.  So it can give a better error.

tests/helpers/perf_results_helper.rb
```diff
@@ -757,5 +760,73 @@ def extract_puppetserver_metrics_from_json(file)
...
+  #
+  # @example
+  #   versions_array = baseline_versions
+  #
+  def baseline_versions
+    sql = "SELECT DISTINCT pe_build_number FROM `perf-metrics.perf_metrics.atop_metrics` " \
+          "WHERE pe_build_number IS NOT NULL " \
+          "ORDER BY pe_build_number"
+    data = query_bigquery(sql)
+    raise "Error: cannot find result that matches query: #{sql}" if data.empty?
```
RandellP: Should probably say more about where the query was sent... like the big query project name, and that it was big query.

spec/tests/helpers/perf_results_helper_spec.rb
```diff
@@ -508,5 +510,98 @@
...
+        test_result = [pe_build_number: "2142.2.1"]
+        test_expected_versions = ["2142.2.1"]
+
+        expect(subject).to receive(:query_bigquery).and_return(test_result)
+        expect(subject.baseline_versions).to eq(test_expected_versions)
+      end
+    end
+  end
+
+  # TODO: implement
```
RandellP: looks implemented... so comment can go

spec/tests/helpers/perf_results_helper_spec.rb
```diff
@@ -508,5 +510,98 @@
...
+      it "raises an error" do
+        ENV["GOOGLE_APPLICATION_CREDENTIALS"] = nil
+        expect { subject.query_bigquery(TEST_BIGQUERY_SQL) }
+          .to raise_error(/GOOGLE_APPLICATION_CREDENTIALS/)
+      end
+    end
+
+    context "when the GOOGLE_APPLICATION_CREDENTIALS environment variable is empty" do
+      it "raises an error" do
+        ENV["GOOGLE_APPLICATION_CREDENTIALS"] = nil
```
RandellP: This test is the same as the one above...  I think you meant to replace nil with "" or something.

rakefile
billclaytor: I've added the following comment: `# validate the BEAKER_INSTALL_TYPE (and dependent env vars) to prevent provisioning if incorrectly specified`

rakefile
billclaytor: Thanks, removed

rakefile
billclaytor: Good catch; I updated `performance` to call `performance_without_provision` using `invoke` which should prevent the validation from being performed twice.

rakefile
billclaytor: I was trying to differentiate between validation vs. verification but I agree that it is confusing from a UX perspective. I've renamed based on your suggestion; thanks.

rakefile
```diff
@@ -612,4 +612,44 @@ namespace :pe_xl do
...
+namespace :bigquery do
+  desc "List the baseline versions in BigQuery"
+  task :baseline_versions do
+    versions = baseline_versions
+    puts "Baseline versions:"
+    puts versions
+  end
+
+  desc "Verify that the specified baseline PE version exists in BigQuery"
+  task :verify_baseline_version do
```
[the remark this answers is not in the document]
billclaytor: Updated

rakefile
billclaytor: I've updated it to check BASELINE_PE_VER before calling the task.

spec/tests/helpers/perf_results_helper_spec.rb
billclaytor: Exactly; updated :)

spec/tests/helpers/perf_results_helper_spec.rb
billclaytor: Good catch; updated

tests/helpers/perf_results_helper.rb
billclaytor: Good call; updated error messages.

tests/helpers/perf_results_helper.rb
billclaytor: In this case I'm naming it 'verify...' and raising the error here so that it can be easily tested, and if the caller wants to trap it they can do that. However, at this point there is only one caller, which is the `validate:baseline_pe_version` rake task, which we don't currently have a good way to test. I could add a `valid_baseline_version?` method and have this call that, but I was holding off on that until there was a need to check the baseline version without raising an error.

RandellP: Since the caller in a rake task I think it is fine this way

PR #344, opened 2019-10-17
(SLV-280) Copy system logs from suts

johnduarte:
This commit adds a helper to copy puppet system logs from suts to the
test runner.

### DIFF
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -674,6 +674,21 @@ module PerfRunHelper
     flunk("One or more assertions failed") unless assertion_exceptions.empty?
   end
 
+  def copy_system_logs(host)
+    job_name = if ENV["JOB_NAME"]
+                 ENV["JOB_NAME"]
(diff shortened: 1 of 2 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -786,6 +801,9 @@ def get_perf_result
       # TODO: tar archive file before copying to avoid timeouts with soak results
```
billclaytor: Since you're making an update here would you mind removing this TODO that I previously added now that SLV-560 is done?

billclaytor: Since you're making an update here would you mind removing this TODO that I previously added now that SLV-560 is done?

johnduarte: You bet. Done.

> Since you're making an update here would you mind removing this TODO that I previously added now that SLV-560 is done?

You bet. Done.

billclaytor: Thanks!

PR #345, opened 2019-10-18
(SLV-652) Update abs_helper to retry request on invalid response code

billclaytor:
This update modifies the 'get_abs_resource_host' method to retry the request up to a max number of attempts (currently set to 3) before failing.

### DIFF
--- a/setup/helpers/abs_helper.rb
+++ b/setup/helpers/abs_helper.rb
@@ -34,6 +34,8 @@ module AbsHelper
   MAX_THREADS = 8
   ABS_ID = "slv"
 
+  ABS_MAX_REQUEST_ATTEMPTS = 3
+
   # Checks whether the user has a valid token and if so initializes the instance variables
   #
   # @author Bill Claytor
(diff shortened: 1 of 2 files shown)


setup/helpers/abs_helper.rb
```diff
@@ -202,14 +204,32 @@ def get_abs_resource_hosts(hosts_to_request)
...
+
     puts "Host_to_request: #{host_to_request}"
+    puts
 
     request_body = get_awsdirect_request_body(host_to_request)
-    response = perform_awsdirect_request(uri, request_body)
+    result = nil
+
+    (1..ABS_MAX_REQUEST_ATTEMPTS).each do |ct|
+      response = perform_awsdirect_request(uri, request_body)
```
johnduarte: Should there be a delay between retries?

billclaytor: Good call; I will add this setting.

billclaytor: Updated.

PR #347, opened 2019-10-24
(SLV-655) Add the pe_xl build directory to the perf and scale results

billclaytor:
This update adds the build dir to the perf and scale results (if it exists). This allows us to identify the nodes for the Large and XL ref archs by role when processing the puppet-metrics-collector data.
- update get_perf_result to copy build dir to perf results
- update copy_scale_results to copy build dir to scale results

### DIFF
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -524,6 +524,9 @@ module PerfRunHelper
     FileUtils.copy_file "#{@archive_root}/start_epoch", "#{scale_result_dir}/start_epoch"
     FileUtils.copy_file "#{@archive_root}/end_epoch", "#{scale_result_dir}/end_epoch"
 
+    # copy the pe_xl build directory to the results if it exists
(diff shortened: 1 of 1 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -524,6 +524,9 @@ def copy_scale_results(scenario)
     FileUtils.copy_file "#{@archive_root}/start_epoch", "#{scale_result_dir}/start_epoch"
     FileUtils.copy_file "#{@archive_root}/end_epoch", "#{scale_result_dir}/end_epoch"
 
+    # copy the pe_xl build directory to the results if it exists
+    FileUtils.cp_r "build", "#{scale_results_parent_dir}/build" if Dir.exist?("build")
```
RandellP: build dir should probably get a module wide variable to hold it's name... like PERF_RESULTS_DIR
Small thing...

billclaytor:
I have updated SLV-655 based on the above suggestion.

PR #348, opened 2019-10-28
(SLV-566) Update provision_hosts_for_roles to update abs resource log

billclaytor:
With this update all hosts provisioned via `provision_hosts_for_roles` are written to the last_abs_resource_hosts.log file. This allows the `performance_deprovision_with_abs` rake task to be used to de-provision hosts that have been provisioned via the `util/abs/provision_pe_xl_nodes.rb` script which is used to provision hosts for the Large and XL reference architectures:
* `provision_hosts_for_role` has been updated to return the host hash rather than the hostname. It converts the keys in the `abs_resource_hosts` hash to symbols and adds the role to the hash. This provides compatibility with the `util/abs/provision_pe_xl_nodes.rb` script which expects the keys to be symbols and the role to be included.
* `provision_hosts_for_roles` has been updated to call `update_last_abs_resource_hosts` with the hosts array (converted to JSON). 
* The spec tests have been updated accordingly.

### DIFF
--- a/setup/helpers/abs_helper.rb
+++ b/setup/helpers/abs_helper.rb
@@ -754,26 +754,31 @@ module AbsHelper
   #
   # @raise [RuntimeError] If provisioning is not successful
   #
-  # @return [String] The hostname of the provisioned host
+  # @return [Hash] The provisioned host
   #
   # @example
-  #   hostname = provision_host_for_role(role)
+  #   host = provision_host_for_role(role)
   #     or
-  #   hostname = provision_host_for_role(role, "c5.xlarge", "40")
+  #   host = provision_host_for_role(role, "c5.xlarge", "40")
   #
-  # TODO: spec test(s)
   def provision_host_for_role(role, size = AWS_SIZE, volume_size = AWS_VOLUME_SIZE)
     host_to_provision = get_host_to_provision(role, size, volume_size)
     abs_resource_hosts = get_abs_resource_hosts(host_to_provision)
     raise "Unable to provision host via ABS" unless abs_resource_hosts
 
+    # parse the JSON
     hosts = parse_abs_resource_hosts(abs_resource_hosts)
-    hostname = hosts[0]["hostname"]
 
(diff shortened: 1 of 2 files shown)


setup/helpers/abs_helper.rb
```diff
@@ -754,26 +754,31 @@ def parse_abs_resource_hosts(abs_resource_hosts)
...
     host_to_provision = get_host_to_provision(role, size, volume_size)
     abs_resource_hosts = get_abs_resource_hosts(host_to_provision)
     raise "Unable to provision host via ABS" unless abs_resource_hosts
 
+    # parse the JSON
     hosts = parse_abs_resource_hosts(abs_resource_hosts)
-    hostname = hosts[0]["hostname"]
 
-    puts "Successfully provisioned host - role: #{role}, hostname: #{hostname}"
+    # convert keys from strings to symbols (provision_pe_xl_nodes.rb expects symbols)
```
johnduarte: Is this a bug in provision_pe_xl_nodes.rb?
It seems like we should standardize on what an "abs_host" hash looks like.  So, either provision_pe_xl_nodes.rb should be updated to use string keys _or_ `parse_abs_resource_hosts` should be updated to use symbol keys.

spec/setup/helpers/abs_helper_spec.rb
```diff
@@ -1292,23 +1304,41 @@ class AbsHelperClass
       end
 
       it "uses the default values and returns the hosts array" do
-        skip "FIXME: The roles array can be built in any order, the test should be fixed to accomodate"
```
johnduarte: High Five

setup/helpers/abs_helper.rb
billclaytor: Ah; I meant to create a ticket to standardize on using symbols as hash keys throughout the module. [SLV-676](https://tickets.puppetlabs.com/browse/SLV-676) has been created.

billclaytor:
Pushed a small typo fix to try and get travis to run the CI - Branch check which didn't run for the last update.

PR #350, opened 2019-11-07
(SLV-520) Create CLI for validate2baseline

johnduarte:
This PR adds the gplt CLI command and the `validate2baseline` to
provide the ability to validate performance results data against
baseline data.

The subcommand requires the following options be provided:
* baseline - Baseline data to compare to (e.g. 2019.1.0)
* results_dir - Path to gatling results to extract data from

The following option is optional and defaults to "apples to apples".
This is required for the big query lookup of the desired baseline data.
* test_type - Test type used to create baseline data


lib/gplt.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "thor"
+require "./tests/helpers/perf_run_helper"
+
+require "pry"
```
johnduarte: delete this

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -397,44 +399,55 @@ class PerfRunHelperClass
   end
 
   describe ".baseline_assert" do
+    # rubocop:disable Metrics/LineLength
+    let(:gatling_assertions) do
```
johnduarte: Refactor the "actual_value" keys to use the baseline_result values.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -397,44 +399,55 @@ class PerfRunHelperClass
...
-        allow(atop_result).to receive(:processes)
-        allow(gatling_result).to receive(:avg_response_time).and_return(baseline_result[:avg_response_time])
-        baseline_result.keys.each do |key|
-          allow(atop_result).to receive(key.to_sym).and_return(baseline_result[key])
-        end
-        # TODO: change back when per process cpu asserts are turned back on
-        # changed to 8 while per process cpu asserts are disabled.
-        # expect(subject).to receive(:assert).with(any_args).exactly(12).times.and_return(nil)
-        expect(subject).to receive(:assert).with(any_args).exactly(8).times.and_return(nil)
+        expect(subject).to receive(:assert).with(any_args).exactly(1).times.and_return(nil)
```
johnduarte: Do we care how many times this is called?

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -397,44 +399,55 @@ class PerfRunHelperClass
...
-        baseline_result.keys.each do |key|
-          allow(atop_result).to receive(key.to_sym).and_return(baseline_result[key])
-        end
-        # TODO: change back when per process cpu asserts are turned back on
-        # changed to 8 while per process cpu asserts are disabled.
-        # expect(subject).to receive(:assert).with(any_args).exactly(12).times.and_raise(Minitest::Assertion)
-        # subject.send(:baseline_assert, atop_result, gatling_result)
-        # expect(subject.assertion_exceptions.count()).to eql(12)
-        expect(subject).to receive(:assert).with(any_args).exactly(8).times.and_raise(Minitest::Assertion)
+        expect(subject).to receive(:assert).with(any_args).exactly(1).times.and_raise(Minitest::Assertion)
```
johnduarte: Do we care now many time this is called?

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -467,4 +480,135 @@ class PerfRunHelperClass
       subject.copy_system_logs(myhost)
     end
   end
+  describe "#validate_results_to_baseline" do
```
johnduarte: :pick: the spec tests in this file use both ".mymethod" and "#mymethod" syntax. Should we standardize? If so, which wins?

tests/helpers/perf_run_helper.rb
```diff
@@ -29,6 +35,14 @@ module PerfRunHelper
   METRIC_RESULTS_DIR = "/root/gatling-puppet-load-test/simulation-runner/results"
   CURRENT_TUNE_SETTINGS = "current_tune_settings.json"
 
+  def logger
```
johnduarte: Explain why this is here!
Two contexts:
* Run with beaker (uses inherited beaker logger)
* Run without beaker (creates logger)

tests/helpers/perf_run_helper.rb
```diff
@@ -29,6 +35,14 @@ module PerfRunHelper
...
+    @logger ||= Beaker::Logger.new
+  end
+
+  def current_test_name
```
johnduarte: Explain why this is here!
Two contexts:
* Run without beaker (uses given test_type from CLI)
* Run with beaker (should inherited [beaker method](https://github.com/puppetlabs/beaker/blob/a01b5732a7d6cec146bd18d08795c64aafe1d2f1/lib/beaker/dsl/helpers/test_helpers.rb#L12), but cannot get to it).  This is a hack to re-implement the method to solve the problem.  This may be solved by calling "extend" rather than "include" on the beaker dsl at the end of this file.

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +942,241 @@ def get_process_hash(perf_result_processes)
...
+    find_file(dir, "atop_log_#{runtype}*.csv")
+  end
+
+  # @param  [String] dir  Path where results are stored
+  # @param  [String] runtype  Test type used to generate results
+  #
+  # @return [PerformanceResult]
+  def atop_results_from_dir(dir, runtype)
+    # This is a dirty, dirty reconstitution of the data from a CSV file
+    # created by [beaker-benchmark(https://github.com/puppetlabs/beaker-benchmark/blob/master/lib/beaker-benchmark/helpers.rb#L219-L231).
```
johnduarte: Fix link syntax

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +942,241 @@ def get_process_hash(perf_result_processes)
...
+  # @return [String]  Path to atop log file
+  def find_atop_log_from_dir(dir, runtype)
+    find_file(dir, "atop_log_#{runtype}*.csv")
+  end
+
+  # @param  [String] dir  Path where results are stored
+  # @param  [String] runtype  Test type used to generate results
+  #
+  # @return [PerformanceResult]
+  def atop_results_from_dir(dir, runtype)
```
johnduarte: This should be broken up into multiple methods.  Add the following
* `gbl_h, proc_h = global_results_from_atop_csv(csv_string)`
* `res = clean_perf_result_attrs(perf_result_obj)`

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +942,241 @@ def get_process_hash(perf_result_processes)
...
+
+  # @param  [String] dir  Path where results are stored
+  #
+  # @return [Array<Hash>]  Array of gatling assertions hashes
+  def gatling_assertions_from_dir(dir)
+    file = find_gatling_assertions_from_dir(dir)
+    json = JSON.parse(read_file(file))
+    gatling_assertions = []
+    json["assertions"].each do |assertion|
+      gatling_assertions << { "expected_values" => assertion["expectedValues"], "message" => assertion["message"],
```
johnduarte: :pick: put each k,v pair on its own line.

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
+  # failing elements with the variance calculation appended to each key.
+  #
+  # @param  [Hash] data  Data to validate { key: [v0, v2], ... }
+  #
+  # @return [Hash]  Hash of failing performance keys associated with there values and variance
+  def find_failing_variances(data)
+    deltas = data.transform_values { |v| v[1].to_f / v[0] }
+    failures = deltas.select { |_k, v| (1 - v).abs > MAX_BASELINE_VARIANCE }
+    # exception for PROC_ORCH_REL_MEM
+    if failures.include? PROC_ORCH_REL_MEM
```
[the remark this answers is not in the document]
RandellP: also, since we are refactoring this code.  It would be better if the override was like 5% (and could be negative) and it meant add/subtract 5% to the baseline then compare.  The difference is that a result that is 15% below the baseline should get flagged...

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
-                   "was not within #{max_baseline_variance}% " \
-                   "of the baseline '#{value}'")
+  # rubocop: enable Naming/AccessorMethodName
+
+  # Calculate the variances on the data set provided and return a hash of the
+  # failing elements with the variance calculation appended to each key.
+  #
+  # @param  [Hash] data  Data to validate { key: [v0, v2], ... }
+  #
+  # @return [Hash]  Hash of failing performance keys associated with there values and variance
```
RandellP: and example of this hash might help... I don't follow what it is... and the line that makes it is complex enough I would have to fire up irb to find out what it does.

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
+  # @param  [Hash] data  Data to validate { key: [v0, v2], ... }
+  #
+  # @return [Boolean]
+  def validate_baseline_delta(data)
+    failures = find_failing_variances(data)
+    return true if failures.empty?
+
+    failures.each do |k, v|
+      variance = ((v.last - 1) * 100).round(2)
+      logger.error("Result '#{k}' is outside tolerances: baseline: #{v.first}; result: #{v[1]}; variance: #{variance}%")
```
RandellP: mixing v.first and v[1] seems odd.  woudl have expected v[0] and v[1] for consistency.

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
+  #
+  # @return [String]  Path to assertions file
+  def find_gatling_assertions_from_dir(dir)
+    find_file(dir, "assertions.json")
+  end
+
+  # @param  [String] file  File to read from
+  #
+  # @return [String]  Contents of file
+  def read_file(file)
```
RandellP: why?

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
+  # @param  [String] dir  Path where results are stored
+  #
+  # @return [GatlingResult]
+  def gatling_result_from_dir(dir)
+    gatling_assertions = gatling_assertions_from_dir(dir)
+    mean_response_time = mean_response_time_from_dir(dir)
+    GatlingResult.new(gatling_assertions, mean_response_time)
+  end
+
+  def find_gatling_stats_from_dir(dir)
```
RandellP: the method name is only 6 chars shorter than the code it runs... why make a method for that?  Seems like more often we define a constant with the file name at the top or something.  Not saying that is the right method... but if you think this is the right way, I would like to discuss it in the next team meeting so we can all agree to move toward doing this in the future... or not... I am a fan of consistency

lib/gplt.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      The --test_type is optional and defaults to "#{default_test_type}"
+
+      A passing state will echo "PASS" to STDOUT and exit with 0.
+      A failing state will echo "FAIL" to STDOUT and exit with 1.
+    LONGDESC
+
+    def validate2baseline
+      if CLI.validate_results_to_baseline(options[:results_dir], options[:baseline], options[:test_type])
+        puts "PASS"
```
RandellP: unless I am missing something... it doesn't actually trigger the code to do the comparison...

spec/tests/helpers/perf_run_helper_spec.rb
billclaytor: Per http://www.betterspecs.org/#describe:

> ...use the Ruby documentation convention of . (or ::) when referring to a class method's name and # when referring to an instance method's name.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# vim: set foldmethod=marker
```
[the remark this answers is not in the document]
johnduarte: As a vim user, the data boilerplate for the spec tests was driving me crazy.  I added the foldmethod setting in order for me to include folding markers around this code in order to collapse it while working on logical components of the tests.

tests/helpers/perf_run_helper.rb
johnduarte: I used this as a simple wrapper in an effort to make mocking it in the spec tests easier.  It may not be necessary.  I will investigate.

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
johnduarte: I will disagree when it comes to defining this value as an adder to the normal variance (i.e. 5% rather than 15%).  To me this would be more confusing and would muddy the waters if we chose to define something with a stricter than normal variance.

All of the variance comparisons occur in either direction.  Any variance either more or less will fail the variance logic.

I don't quite understand what you have in mind for the dynamic looping for `MAX_BASELINE_VARIANCE_PROC_ORCH_REL_MEM`.  Can you provide more detail to your thoughts on this?

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
+
+    json = JSON.parse(read_file(file))
+    json.fetch("meanResponseTime").fetch("total")
+  end
+
+  # @param  [Hash{Symbol=>String}] baseline  A hash of performance results for basline
+  # @param  [Object] results  Results data
+  #
+  # @return [Hash{Symbol=>Array}]  Merged results performance data in arrays associated with keys
+  def baseline_to_results_delta(baseline, *results)
```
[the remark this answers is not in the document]
johnduarte: Point well taken.  I have added more detail to the yard docs for this method to clarify.

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
+  # rubocop: enable Naming/AccessorMethodName
+
+  # Calculate the variances on the data set provided and return a hash of the
+  # failing elements with the variance calculation appended to each key.
+  #
+  # @param  [Hash] data  Data to validate { key: [v0, v2], ... }
+  #
+  # @return [Hash]  Hash of failing performance keys associated with there values and variance
+  def find_failing_variances(data)
+    deltas = data.transform_values { |v| v[1].to_f / v[0] }
```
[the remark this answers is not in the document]
johnduarte: Would the term ratio be better?

tests/helpers/perf_run_helper.rb
johnduarte: Point well taken.  I have added more detail to the yard docs for this method to clarify.

tests/helpers/perf_run_helper.rb
johnduarte: fixed.

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
RandellP: what you say is true.  Really there are 2 types of overrides than.  One to change the overall variance from the default. This is really a configuration type setting, so not really an override.  The other is to shift the whole range based on a root change to the perf.  What we did for orch was the first one, increase the variance range... because it was the easiest thing we could do at the time.  But it is technically wrong.  Say a metrics was 10, and we accepted plus or minus 1 (10%).  9 would be okay.  But they added something new in the latest release, so that 12 was the nominal value.  What we did to accommodate that was change the plus or minus to 3... allowing for 13 (10+3)... but now also allowing for 7 (10-3).  Which we shouldn't have done.  If seven was worth flagging before it should still be worth flagging now.  The correct way would be to apply a +2 to the base and keep the variance the same.  Your refactor makes this not so hard to do... and I was suggesting we do it... or accommodate doing it with some minor changes...

tests/helpers/perf_run_helper.rb
johnduarte: This is the result of refactoring multiple regex lookups.  My solution was to create a generic `find_file` method and then use it to abstract methods for finding desired files by there intention.  I am not married to this implementation, but it does encapsulate these lookups.  If the file used to store gatling stats changes in the future, only the pattern within this method would need to be updated.
This was the outcome of driving the code from the spec tests.  These methods made it much easier for me to control the data that I wanted to pull in to test the larger validation methods.

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
RandellP: as for dynamic, which figures into my above novel...  when calculating the delta, you could add in a base adjustment if a variable like BASE_ADJUSTMENT_<key>  where key is like PROC_ORCH_REL_MEM which I think is also the key to the data hash that is passed in.

lib/gplt.rb
johnduarte: I think you are missing something. :-)
You should be able to test this for yourself on the command line:
```
$ bundle exec  bin/gplt validate2baseline --baseline=2019.1.0 --results-dir=$MYRESULTS
Getting mean response time from /home/john/Downloads/results/perf/PERF_1571758219/ip-10-227-3-92.amz-dev
.puppet.net/root/gatling-puppet-load-test/simulation-runner/results/PerfTestLarge-1571758226846/js/globa
l_stats.json
Baseline result returned from BigQuery: [{:avg_cpu=>20, :avg_mem=>2630938, :avg_disk_write=>22063, :avg_
response_time=>644, :process_bolt_server_avg_cpu=>nil, :process_bolt_server_avg_mem=>nil, :process_ace_s
erver_avg_cpu=>nil, :process_ace_server_avg_mem=>nil, :process_puppetdb_avg_cpu=>3, :process_puppetdb_av
g_mem=>734776, :process_console_services_release_avg_cpu=>2, :process_console_services_release_avg_mem=>
549411, :process_orchestration_services_release_avg_cpu=>6, :process_orchestration_services_release_avg_
mem=>595188, :process_puppet_server_release_avg_cpu=>6, :process_puppet_server_release_avg_mem=>3706311}
]
...
Result 'avg_cpu' is outside tolerances: baseline: 20; result: 11; variance: -45.0%
Result 'avg_response_time' is outside tolerances: baseline: 644; result: 391; variance: -39.29%
Result 'process_puppetdb_avg_cpu' is outside tolerances: baseline: 3; result: 0; variance: -100.0%
Result 'process_console_services_release_avg_cpu' is outside tolerances: baseline: 2; result: 3; variance: 50.0%
Result 'process_orchestration_services_release_avg_cpu' is outside tolerances: baseline: 6; result: 4; variance: -33.33%
Result 'process_puppet_server_release_avg_cpu' is outside tolerances: baseline: 6; result: 0; variance:
-100.0%
FAIL
$ echo $?
1
```

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
RandellP: dynamic part 2... Currently you do the variance check against max variance.  Then check the <key> against a hardcoded key value, and if found redo the variance check with a hardcoded variable override.  I am saying, name the variable VARIANCE_ADJUSTMENT_<key>.  Then before doing the first variance check, look to see if a VARIANCE_ADJUSTMENT_<key> exists for the key you are considering.  And if so, use the value of that VARIANCE_ADJUSTMENT_<key> variable instead of MAX_BASELINE_VARIANCE.  By doing this... all we have to do to add a new variance override is define a new variable.  As it is right now, we have to add a new variable, and a hardcoded conditional to figure out when to recalculate and possibly delete the failure...

tests/helpers/perf_run_helper.rb
RandellP: like this one...
allow(subject).to receive(:read_file).with("atop_file").and_return(atop_file_content)
you can do this with
allow(File).to receive(:read).with("atop_file").and_return(atop_file_content)
Then you won't need the wrapper.

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
RandellP: created https://tickets.puppetlabs.com/browse/SLV-699 for the baseline adjust feature.

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
RandellP: yes :)

lib/gplt.rb
RandellP: looks like thor does it under the hood.

RandellP: ok i see it.  You extend perf_runner and thus it brings validate_results_to_baseline into the class...
Any reason why you didn't require it instead and then call into it?

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
johnduarte: Using the defined constants to dynamically assign variance maximum thresholds has been implemented and added to this PR.

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +945,253 @@ def get_process_hash(perf_result_processes)
...
-      assert_later((assert_value - value) / value * 100 <= max_baseline_variance,
-                   "The value of #{key} '#{assert_value}' " \
-                   "was not within #{max_baseline_variance}% " \
-                   "of the baseline '#{value}'")
+  # rubocop: enable Naming/AccessorMethodName
+
+  # Calculate the variances on the data set provided and return a hash of the
+  # failing elements with the variance calculation appended to each key.
+  #
+  # @param  [Hash] data  Data to validate { key: [v0, v2], ... }
```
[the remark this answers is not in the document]
johnduarte: I have updated the yard docs for this method to clarify its expectations.

tests/helpers/perf_run_helper.rb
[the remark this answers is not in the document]
johnduarte: All occurrences of "delta" have been replaced with more appropriate terms.

tests/helpers/perf_run_helper.rb
johnduarte: Wrapper deleted.  Thanks for the example.

lib/gplt.rb
```diff
@@ -0,0 +1,51 @@
...
+module GPLT
+  class Error < StandardError; end
+
+  # Class to define CLI subcommands
+  class CLI < Thor
+    extend PerfRunHelper
+
+    default_test_type = "apples to apples"
+
+    option :test_type, default: default_test_type
```
billclaytor: Are these options of the `gplt` command or the `validate2baseline` sub-command? If the former, does this mean that the options for all sub-commands need to be options of the primary command? Based on the discussion on this being the main CLI for GPLT it would seem best to limit the scope for options to each sub-command to avoid collisions / cluttering the primary 'namespace' but I'm not sure if that is possible.

billclaytor: All of the other standalone scripts for GPLT feature short versions of the options. It would be nice if this followed that pattern.

They also feature a `-h` option to display the help. It seems weird to have a totally different usage pattern for this script:
```
be bin/gplt help validate2baseline
```
vs. (for examlple)
```
be ruby util/abs/provision_pe_xl_nodes.rb -h
```

billclaytor:
Perhaps I'm doing it wrong, but I can't seem to get it to work:

This works:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> be bin/gplt help
Commands:
  gplt help [COMMAND]                                                   # Describe available commands or one specific command
  gplt validate2baseline --baseline=BASELINE --results-dir=RESULTS_DIR  # validate perf results against baseline data
```

So does this:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> be bin/gplt help validate2baseline
Usage:
  gplt validate2baseline --baseline=BASELINE --results-dir=RESULTS_DIR

Options:
  [--test-type=TEST_TYPE]    
                             # Default: apples to apples
  --baseline=BASELINE        
  --results-dir=RESULTS_DIR  

Description:
  `validate2baseline` validates performance results data against baseline data.

  You must specify a --baseline for the baseline data and a --results_dir containing the performance results to compare for validation.

  The --test_type is optional and defaults to "apples to apples"

  A passing state will echo "PASS" to STDOUT and exit with 0. A failing state will echo "FAIL" to STDOUT and exit with 1.
```

But this does not:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> be bin/gplt validate2baseline --baseline=2018.1.0 --results_dir=/Users/bill.claytor/tmp/680p/gatling-puppet-load-test/results/perf/PERF_1572446968
bundler: failed to load command: bin/gplt (bin/gplt)
NameError: uninitialized constant PerfRunHelper::System
Did you mean?  SystemExit
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1107:in `find_file'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1043:in `find_atop_log_from_dir'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1076:in `atop_results_from_dir'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1198:in `validate_results_to_baseline'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/lib/gplt.rb:42:in `validate2baseline'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
  /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
  bin/gplt:7:in `<top (required)>'
```

johnduarte:
> 
> But this does not:
> 
> ```
> bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> be bin/gplt validate2baseline --baseline=2018.1.0 --results_dir=/Users/bill.claytor/tmp/680p/gatling-puppet-load-test/results/perf/PERF_1572446968
> bundler: failed to load command: bin/gplt (bin/gplt)
> NameError: uninitialized constant PerfRunHelper::System
> Did you mean?  SystemExit
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1107:in `find_file'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1043:in `find_atop_log_from_dir'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1076:in `atop_results_from_dir'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1198:in `validate_results_to_baseline'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/lib/gplt.rb:42:in `validate2baseline'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
>   /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/vendor/bundle/ruby/2.5.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
>   bin/gplt:7:in `<top (required)>'
> ```

Thanks @billclaytor . This looks like a bug with the library scoping.  I will investigate.

lib/gplt.rb
RandellP: I am a big fan of -h for help as well...   It looks like the top level with no subcommand kinda prints out the help if you do -h, but only because that is the default for anything it doesn't understand... so that one is fine.  But setting up 
`be bin/gplt validate2baseline -h`
To also give the help would be really nice.  Right now it give a pretty lame message 
`No value provided for required options '--baseline', '--results-dir'`
Which doesn't even tell you how to get the help you are looking for.

johnduarte:
@billclaytor fix pushed for the StandardError exception bug you found.

johnduarte:
Short options flags added.  Help flags added.

util/validate2baseline
```diff
@@ -0,0 +1,61 @@
...
+  mandatory = %i[test_type baseline results_dir]
+  missing = mandatory.select { |o| options[o].nil? }
+  raise OptionParser::MissingArgument, missing.join(", ") unless missing.empty?
+rescue OptionParser::InvalidOption, OptionParser::MissingArgument
+  puts $ERROR_INFO.to_s
+  puts optparse
+  exit
+end
+
+c = GPLT::CLI.new
```
billclaytor: Now that this is a standalone script what is the benefit of going through `CLI` rather than just calling `validate_results_to_baseline` directly? Not saying there isn't a benefit, just want to understand it :)

util/validate2baseline
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env ruby
```
billclaytor: Should this be in a sub-directory of `util` to keep things organized (perhaps `results`)?

tests/helpers/perf_run_helper.rb
```diff
@@ -944,75 +947,259 @@ def get_process_hash(perf_result_processes)
...
-
-      max_baseline_variance = \
-        if key.to_s == "process_orchestration_services_release_avg_mem"
-          15
-        else
-          10
-        end
-      next unless value.is_a? Integer
+    data = baseline_to_results_data(baseline, atop_result, gatling_result)
+    assert_later(validate_baseline_data(data), "Things are really fucked up now")
```
billclaytor: I'm not offended by this language, but others may be, so we may want to keep it clean:
```suggestion
    assert_later(validate_baseline_data(data), "Things are really mucked up now")
```

billclaytor:
I'm not able to get this to work. I'm not sure if I'm supposed to have created the results with this updated code as well, but the description doesn't say that, so I'm using the same results dir that I used for the `results.md` doc for SLV-680.

My results dir exists:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> ls /Users/bill.claytor/tmp/680p/gatling-puppet-load-test/results/perf/PERF_1572446968
20191030T144928Z-20191030T184941Z.tar.gz        ip-10-227-0-248.amz-dev.puppet.net              puppetserver.csv
current_tune_settings.json                      puppet-metrics-collector                        puppetserver.csv.html
end_epoch                                       puppetserver.average.csv                        start_epoch
ip-10-227-0-214.amz-dev.puppet.net              puppetserver.average.csv.html
```
Yet it is saying "The file does not exist":
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> be ruby util/validate2baseline validate2baseline --r /Users/bill.claytor/tmp/680p/gatling-puppet-load-test/results/perf/PERF_1572446968 --b 2018.1.0
Traceback (most recent call last):
        5: from util/validate2baseline:61:in `<main>'
        4: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/lib/cli.rb:17:in `validate2baseline'
        3: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1198:in `validate_results_to_baseline'
        2: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1076:in `atop_results_from_dir'
        1: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1043:in `find_atop_log_from_dir'
/Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1107:in `find_file': The file does not exist (StandardError)
```

Even if I'm doing it wrong, the error doesn't tell me what file it is trying to find or where it expects to find it within the results dir I specified.

util/validate2baseline
RandellP: I have the same question. :)

util/validate2baseline
```diff
@@ -0,0 +1,61 @@
...
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
+
+require "optparse"
+require "cli"
+
+DEFAULT_TEST_TYPE = "apples to apples"
+DESCRIPTION = <<-DESC
+  `validate2baseline` validates performance results data against baseline data.
```
RandellP: There should be an update to one of our docs about how this script fits into our workflow, and such...

johnduarte: I have no idea how this fits into our workflow.  I am happy to ticket such documentation, but I have no context for what workflows/efforts this script are meant to support.  The ticket for this PR and all of its related tickets are simply flagged as tech-debt.  @RandellP do you have insight into what epic(s) such documentation should be added to?

RandellP: Well it is for use comparing to a baseline after a run is done.  Like say you forgot to set the baseline.  Or you want to compare to an additional.  But at the very list the base usage should be covered in something like report_utils...  https://github.com/puppetlabs/gatling-puppet-load-test/blob/master/docs/report_utils.md  
The long term goal is to use this in CI.  So we only have to characterize a build once, then have separate jobs that compare that build to multiple pipelines.

util/validate2baseline
johnduarte: A class instance is required for the helpers in the PerfRunHelper module to work as expected.  Normally this happens in the context of a Beaker instance, but here a class wrapper needs to be created to satisfy that requirement.

util/validate2baseline
johnduarte: I am fine with nesting it in a `results` directory.

IMHO, this type of organization doesn't add any value.  When I want to run a command I don't want to type out a multilayered path to said command.  If the `util` directory is where the executables live, then that is as much as I want to type.  This is out of sheer laziness on my part.

```
bundle exec util/validate2baseline
```
vs
```
bundle exec util/results/validate2baseline
```

Having said that, if it makes more sense to @billclaytor and @RandellP than I have no problem with it.

tests/helpers/perf_run_helper.rb
johnduarte: I have pushed a commit to change this message to something more appropriate to the context in which it gets emitted.

johnduarte:
> I'm not able to get this to work. I'm not sure if I'm supposed to have created the results with this updated code as well, but the description doesn't say that, so I'm using the same results dir that I used for the `results.md` doc for SLV-680.
> 
> My results dir exists:
> 
> ```
> bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> ls /Users/bill.claytor/tmp/680p/gatling-puppet-load-test/results/perf/PERF_1572446968
> 20191030T144928Z-20191030T184941Z.tar.gz        ip-10-227-0-248.amz-dev.puppet.net              puppetserver.csv
> current_tune_settings.json                      puppet-metrics-collector                        puppetserver.csv.html
> end_epoch                                       puppetserver.average.csv                        start_epoch
> ip-10-227-0-214.amz-dev.puppet.net              puppetserver.average.csv.html
> ```
> 
> Yet it is saying "The file does not exist":
> 
> ```
> bill.claytor:~/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test> be ruby util/validate2baseline validate2baseline --r /Users/bill.claytor/tmp/680p/gatling-puppet-load-test/results/perf/PERF_1572446968 --b 2018.1.0
> Traceback (most recent call last):
>         5: from util/validate2baseline:61:in `<main>'
>         4: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/lib/cli.rb:17:in `validate2baseline'
>         3: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1198:in `validate_results_to_baseline'
>         2: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1076:in `atop_results_from_dir'
>         1: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1043:in `find_atop_log_from_dir'
> /Users/bill.claytor/RubymineProjects/_forks/johnduarte/~/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1107:in `find_file': The file does not exist (StandardError)
> ```
> 
> Even if I'm doing it wrong, the error doesn't tell me what file it is trying to find or where it expects to find it within the results dir I specified.

@billclaytor I have updated the exception error message to emit the file pattern and directory for which it is looking.  So, you should get a better error now.

From a flat directory list I cannot tell you which file is missing.  The code looks for the following files recursively in given results dir and will raise an exception if any of them are not found.
* atop_log_*_json.csv # where the wild card matches the test type given (a2a by default)
* assertions.json
* global_stats.json

util/validate2baseline
billclaytor: I'm thinking of it more from an organizational perspective than an execution perspective (which admittedly may be the wrong perspective). It seems that if all of the scripts were in the `util` directory it would be very cluttered and potentially difficult to find the script you're interested in, especially if a directory listing scrolls offscreen in tmux and I have to do a `control-B / [` and scroll up to find the script I want. Of course this script would always be at or near the bottom of that list ;)

Just trying to keep things standardized and organized.

johnduarte: :+1: I have moved the file into `util/results`

billclaytor: Thanks!

util/validate2baseline
johnduarte: [SLV-701](https://tickets.puppetlabs.com/browse/SLV-701) created for this work.

billclaytor:
Running with the updated code I get an error that it can't find `atop_log_applestoapples*.csv`:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test> be ruby util/results/validate2baseline validate2baseline --r results/perf/PERF_1572446968  --b 2018.1.0
Traceback (most recent call last):
        5: from util/results/validate2baseline:61:in `<main>'
        4: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/lib/cli.rb:17:in `validate2baseline'
        3: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1198:in `validate_results_to_baseline'
        2: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1076:in `atop_results_from_dir'
        1: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1043:in `find_atop_log_from_dir'
/Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1107:in `find_file': File matching pattern 'atop_log_applestoapples*.csv' not found in path results/perf/PERF_1572446968 (StandardError)
```

---

Added some output for debugging; perhaps the issue is that it is finding multiple files matching the pattern:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test> be ruby util/results/validate2baseline validate2baseline --r results/perf/PERF_1572446968  --b 2018.1.0

find_file:
  dir: results/perf/PERF_1572446968
  pat: atop_log_applestoapples*.csv

file:
results/perf/PERF_1572446968/ip-10-227-0-214.amz-dev.puppet.net/atop_log_applestoapples_json.csv
results/perf/PERF_1572446968/ip-10-227-0-214.amz-dev.puppet.net/atop_log_applestoapples_json.summary.csv
results/perf/PERF_1572446968/ip-10-227-0-214.amz-dev.puppet.net/atop_log_applestoapples_json.detail.csv

Traceback (most recent call last):
        5: from util/results/validate2baseline:61:in `<main>'
        4: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/lib/cli.rb:17:in `validate2baseline'
        3: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1209:in `validate_results_to_baseline'
        2: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1076:in `atop_results_from_dir'
        1: from /Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1043:in `find_atop_log_from_dir'
/Users/bill.claytor/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test/tests/helpers/perf_run_helper.rb:1118:in `find_file': File matching pattern 'atop_log_applestoapples*.csv' not found in path results/perf/PERF_1572446968 (StandardError)
```

The file does exist:
```
bill.claytor:~/RubymineProjects/_forks/johnduarte/_branches/SLV-520/gatling-puppet-load-test> ls results/perf/PERF_1572446968/ip-10-227-0-214.amz-dev.puppet.net
atop_log_applestoapples_json.csv                atop_log_applestoapples_json.log.tar.gz         atop_log_warmupjit_json.log.tar.gz
atop_log_applestoapples_json.detail.csv         atop_log_applestoapples_json.summary.csv        var
atop_log_applestoapples_json.detail.csv.html    atop_log_applestoapples_json.summary.csv.html
atop_log_applestoapples_json.log                atop_log_warmupjit_json.log
```

util/validate2baseline
billclaytor: I was able to avoid using `cli.rb` by updating `validate2baseline` with the following:
* require `perf_run_helper` and include `PerfRunHelper`
```
# require "cli"

require "./tests/helpers/perf_run_helper.rb"
include PerfRunHelper # rubocop:disable Style/MixinUsage
```
* call `validate_results_to_baseline` directly:
```
# c = GPLT::CLI.new
# c.validate2baseline(options)

validate_results_to_baseline(options[:results_dir], options[:baseline], options[:test_type])
```

This is how the other util scripts do it and seems to work fine unless I'm not understanding some aspect of this (which is entirely possible).

johnduarte: Without bringing in the conditional logic from `cli.rb` the interface contract for the application is broken.  It won't emit the PASS/FAIL status and it won't return the expected exit code on failure.
Stylistically, I would wrap it in class anyway to avoid having all methods in the Object rather than namespaced (thus avoiding having to squash rubocop).
Having said that, I am not against removing the `cli.rb` file and bringing that code into the `validate2baseline` file if that satisfies the criteria of being more consistent with the other `util` items.

billclaytor: Sorry for overlooking the conditional logic; I was just trying to see if I could get the perf_results_helper functionality working within the script without relying on `cli.rb`. In addition to being consistent with the other scripts, having less 'parts' makes it easier to understand (for me).

johnduarte: I have also just complicated the issues by adding a `--log-level` flag which needs to influence the object outside of the context of the `validate_results_to_baseline`.  I believe this to be assisted by current structure as well.  Feel free to take a look and see if your opinion differs.  Thanks Bill.

docs/results_utils.md
```diff
@@ -0,0 +1,34 @@
...
+* `--results_dir`: This is the path to the results data set that you wish to validate.
+
+There is one optional option:
+* `--test_type`:  This corresponds to the test type that was used to store the baseline data in Google Big Query.  This defaults to "apples to apples".
+
+Example:
+```
+$ bundle exec util/results/validate2baseline --baseline=2019.1.0 --results_dir=$HOME/Downloads/results/perf/PERF_1571758219
+...
+FAIL
```
RandellP: An example of passing output would probably be more helpful.

johnduarte: So, you would prefer the following content?
```
$ bundle exec util/results/validate2baseline --baseline=2019.1.0 --results_dir=$HOME/Downloads/results/perf/PERF_1571758219
...
PASS
$echo $?
0
```

RandellP: yeah.  that would be.  Though I was thinking of the detailed output we see in the test run.  I think we may need a ticket to make it produce the comparison output with the actual and baseline...  Cause just thinking about using this in CI... all we would get is a fail in the log... that wouldn't be very helpful.

PR #351, opened 2019-11-08
(SLV-691) Delete left over puppet logdir

johnduarte:
This commit updates the `copy_system_logs` helper to delete the system
log file tree left over by the beaker archive_file_from helper.

### DIFF
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -444,15 +444,14 @@ describe PerfRunHelperClass do # rubocop:disable Metrics/BlockLength
                            { platform: Beaker::Platform.new("centos-6.5-x86_64"),
                              role: "master" }, logger: @logger)]
     end
-    it "calls archive_file_from with given host and parent dir of puppet_logdir" do
+    it "calls scp_from to copy puppet logdir to archive root" do
       # test variables
-      build_id = gplt_timestamp = nil
-      puppet_logdir = "/foo/bar/baz"
-      archive_root = "/test/archive"
-      job_name = "foobar"
-      archive_name = "#{job_name}__#{build_id}__#{gplt_timestamp}__system_logs.tgz"
-      ENV["JOB_NAME"] = job_name
+      logdir_root = "foo"
+      puppet_logdir = File.join(logdir_root, "bar", "baz")
+      archive_root = "/tmp/test/archive"
       myhost = hosts.first
+      expected_dest = File.join(archive_root, myhost, logdir_root)
+
       # mock out @archive_root instance variable
       subject.instance_variable_set(:@archive_root, archive_root)
       # mock out beaker result
@@ -461,9 +460,9 @@ describe PerfRunHelperClass do # rubocop:disable Metrics/BlockLength
 
       allow(subject).to receive(:on).with(myhost, any_args).and_return(result)
       expect(subject).to receive(:puppet).with("config", "print", "logdir")
-      expect(subject).to receive(:archive_file_from).with(myhost,
-                                                          File.dirname(puppet_logdir),
-                                                          {}, archive_root, archive_name)
+      expect(subject).to receive(:scp_from).with(myhost,
+                                                 File.dirname(puppet_logdir),
+                                                 expected_dest)
       subject.copy_system_logs(myhost)
     end
   end
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -439,31 +439,63 @@ describe PerfRunHelperClass do # rubocop:disable Metrics/BlockLength
     end
   end
   describe "#copy_system_logs" do
-    let!(:hosts) do
-      [Beaker::Host.create("master",
-                           { platform: Beaker::Platform.new("centos-6.5-x86_64"),
-                             role: "master" }, logger: @logger)]
+    let(:host) { double.as_null_object }
+    let(:archive_root) { "/tmp/test/archive" }
+    let(:archive_dir) { "/var/log" }
+    let(:archive_name) { "puppet_logdir.tgz" }
+    let(:archive_path) { File.join(archive_dir, archive_name) }
+
+    context "when archiving succeeds" do
+      it "calls scp_from to copy archive path to archive root" do
+        expected_dest = File.join(archive_root, host)
+
+        # mock out @archive_root instance variable
+        subject.instance_variable_set(:@archive_root, archive_root)
+
+        allow(subject).to receive(:archive_system_logs).with(host).and_return(archive_path)
+        expect(subject).to receive(:scp_from).with(host, archive_path, expected_dest)
+
+        subject.copy_system_logs(host)
+      end
     end
-    it "calls scp_from to copy puppet logdir to archive root" do
-      # test variables
-      logdir_root = "foo"
-      puppet_logdir = File.join(logdir_root, "bar", "baz")
-      archive_root = "/tmp/test/archive"
-      myhost = hosts.first
-      expected_dest = File.join(archive_root, myhost, logdir_root)
+    context "when archiving fails" do
+      it "scp_from is not called" do
+        allow(subject).to receive(:archive_system_logs).with(host).and_return(nil)
+        expect(subject).not_to receive(:scp_from)
 
-      # mock out @archive_root instance variable
-      subject.instance_variable_set(:@archive_root, archive_root)
-      # mock out beaker result
-      result = Beaker::Result.new(myhost, "foo")
-      result.stdout = puppet_logdir
+        subject.copy_system_logs(host)
+      end
+    end
+  end
+  describe "#archive_system_logs" do
+    let(:host) { double.as_null_object }
+    let(:puppet_logdir) { "/var/log/puppetlabs/puppet" }
+    let(:archive_dir) { "/var/log" }
+    let(:archive_name) { "puppet_logdir.tgz" }
+    let(:archive_path) { File.join(archive_dir, archive_name) }
 
-      allow(subject).to receive(:on).with(myhost, any_args).and_return(result)
-      expect(subject).to receive(:puppet).with("config", "print", "logdir")
-      expect(subject).to receive(:scp_from).with(myhost,
-                                                 File.dirname(puppet_logdir),
-                                                 expected_dest)
-      subject.copy_system_logs(myhost)
+    context "when archiving succeeds" do
+      it "returns archive path" do
+        expect(subject).to receive(:archive_system_logs).with(host).and_return(archive_path)
+        subject.archive_system_logs(host)
+      end
+      it "runs tar on the host" do
+        result = Beaker::Result.new(host, "foo")
+        result.stdout = puppet_logdir
+        cmd1 = "rm -f #{archive_path}"
+        cmd2 = "cd #{archive_dir} && tar -czf #{archive_name} puppetlabs"
+        allow(subject).to receive(:puppet).and_return(puppet_logdir)
+        expect(subject).to receive(:on).with(host, puppet_logdir).ordered.and_return(result)
+        expect(subject).to receive(:on).with(host, cmd1, accept_all_exit_codes: true).ordered
+        expect(subject).to receive(:on).with(host, cmd2, accept_all_exit_codes: true).ordered
+        subject.archive_system_logs(host)
+      end
+    end
+    context "when archiving fails" do
+      it "returns nil" do
+        expect(subject).to receive(:archive_system_logs).with(host).and_return(nil)
+        subject.archive_system_logs(host)
+      end
     end
   end
 end
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -446,23 +446,28 @@ describe PerfRunHelperClass do # rubocop:disable Metrics/BlockLength
     let(:archive_path) { File.join(archive_dir, archive_name) }
 
     context "when archiving succeeds" do
+      before { subject.instance_variable_set(:@archive_root, archive_root) }
+      before { allow(subject).to receive(:archive_system_logs).with(host).and_return(archive_path) }
+      before { allow(subject).to receive(:scp_from) }
+      before { allow(FileUtils).to receive(:mkdir_p) }
       it "calls scp_from to copy archive path to archive root" do
         expected_dest = File.join(archive_root, host)
-
-        # mock out @archive_root instance variable
-        subject.instance_variable_set(:@archive_root, archive_root)
-
-        allow(subject).to receive(:archive_system_logs).with(host).and_return(archive_path)
         expect(subject).to receive(:scp_from).with(host, archive_path, expected_dest)
-
+        subject.copy_system_logs(host)
+      end
+      it "calls mkdir_p to create the local destination path" do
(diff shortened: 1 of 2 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -683,10 +683,17 @@ def copy_system_logs(host)
                  "unknown_or_dev_job"
                end
 
-    archive_name = "#{job_name}__#{ENV['BUILD_ID']}__#{@gplt_timestamp}__system_logs.tgz"
     puppet_logdir = File.dirname on(host, puppet("config", "print", "logdir")).stdout.strip
-    archive_file_from(host, puppet_logdir,
-                      {}, @archive_root, archive_name)
+    logdir_root = puppet_logdir.split(File::SEPARATOR).select { |s| s.length >= 1 }.shift
```
RandellP: I had to run this through irb to find out what it does...

tests/helpers/perf_run_helper.rb
```diff
@@ -683,10 +683,17 @@ def copy_system_logs(host)
...
+
+    dest = File.join(@archive_root, host, File.dirname(puppet_logdir))
+    tar_root = File.join(@archive_root, host, logdir_root)
```
RandellP: so... but why.  you are grabbing the log dir, which may I assume be in different places on different OS's... and you are naming the tar file after the top level dir...  seems like you should just call it .../puppet_log_dir.tgz

tests/helpers/perf_run_helper.rb
johnduarte: I have added a comment to try and demystify it. :)

tests/helpers/perf_run_helper.rb
johnduarte: This is the tree that is copied over from the SUT, so this is required to delete the tree afterwards.  Your point is well taken that the name of the archive should not use this value.  I have updated it below to give it the name `puppet_logdir.tgz`.  Thanks.

tests/helpers/perf_run_helper.rb
RandellP: it's better...  but I might only understand it because I now know what it does...
Maybe
# Get top-level dir name.  ex: /var/logs/puppet => var
That would be undeniably clear. :)

tests/helpers/perf_run_helper.rb
RandellP: yeah on the name change... that was my main thing.  I actually missed the use in the rm_rf.  Which gives me a new point... can we/should we tar it up on the SUT instead.  That would remove the need to clean up.  Another PRO would be less data moving (faster) meaning less of a chance of it doing what we saw with the soak tests and the atop log (failed due to the size of the data in transit).  The CON is the SUT has to have the space for the tar file to be created...

billclaytor:
The contents of the `.tgz` file includes the full directory tree of the results dir. It seems like the directory tree should just include the `log` dir:
```
bill.claytor:~/Downloads/PERF_1573500768/ip-10-227-1-19.amz-dev.puppet.net> tree results
results
└── perf
    └── PERF_1573500768
        └── ip-10-227-1-19.amz-dev.puppet.net
            └── var
                └── log
                    └── puppetlabs
```

tests/helpers/perf_run_helper.rb
johnduarte: @RandellP this PR has been retooled to archive the directory on the SUT and then copy the archive only.  This is basically a complete rewrite, so this PR need a complete re-review.

johnduarte:
> The contents of the `.tgz` file includes the full directory tree of the results dir. It seems like the directory tree should just include the `log` dir:
> 
> ```
> bill.claytor:~/Downloads/PERF_1573500768/ip-10-227-1-19.amz-dev.puppet.net> tree results
> results
> └── perf
>     └── PERF_1573500768
>         └── ip-10-227-1-19.amz-dev.puppet.net
>             └── var
>                 └── log
>                     └── puppetlabs
> ```

@billclaytor this bug has been addressed.  The PR has undergone a complete rewrite, so this PR need a complete re-review.

In a test case where `@archive_root` is set to `/tmp/foobar`, the following results are obtained when running the `copy_system_logs` helper...
```
$ tree /tmp/foobar
/tmp/foobar
├── aerobic-record.delivery.puppetlabs.net
│   └── puppet_logdir.tgz
└── western-ileum.delivery.puppetlabs.net
    └── puppet_logdir.tgz
$ tar -tzf /tmp/foobar/aerobic-record.delivery.puppetlabs.net/puppet_logdir.tgz
puppetlabs/
puppetlabs/installer/
puppetlabs/installer/2019-11-12T18.23.28+0000.install.log
puppetlabs/puppet/
puppetlabs/pxp-agent/
puppetlabs/pxp-agent/pxp-agent.log
puppetlabs/puppetserver/
...
```

tests/helpers/perf_run_helper.rb
johnduarte: This has been refactored out of existence.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,63 @@ class PerfRunHelperClass
...
-      [Beaker::Host.create("master",
-                           { platform: Beaker::Platform.new("centos-6.5-x86_64"),
-                             role: "master" }, logger: @logger)]
+    let(:host) { double.as_null_object }
+    let(:archive_root) { "/tmp/test/archive" }
+    let(:archive_dir) { "/var/log" }
+    let(:archive_name) { "puppet_logdir.tgz" }
+    let(:archive_path) { File.join(archive_dir, archive_name) }
+
+    context "when archiving succeeds" do
```
billclaytor: Running this actually creates the `/tmp/test/archive/#[Double (anonymous)]` on my laptop.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,63 @@ class PerfRunHelperClass
...
-      allow(subject).to receive(:on).with(myhost, any_args).and_return(result)
-      expect(subject).to receive(:puppet).with("config", "print", "logdir")
-      expect(subject).to receive(:archive_file_from).with(myhost,
-                                                          File.dirname(puppet_logdir),
-                                                          {}, archive_root, archive_name)
-      subject.copy_system_logs(myhost)
+    context "when archiving fails" do
+      it "scp_from is not called" do
+        allow(subject).to receive(:archive_system_logs).with(host).and_return(nil)
+        expect(subject).not_to receive(:scp_from)
```
billclaytor: See comment above; this example should probably expect not to receive the `FileUtils.mkdir_p` as well.

spec/tests/helpers/perf_run_helper_spec.rb
johnduarte: Thanks @billclaytor I have added a test for the mkdir_p expectation.

spec/tests/helpers/perf_run_helper_spec.rb
johnduarte: Thanks @billclaytor I have added a test for the anti mkdir_p expectation.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
...
-      result.stdout = puppet_logdir
-
-      allow(subject).to receive(:on).with(myhost, any_args).and_return(result)
-      expect(subject).to receive(:puppet).with("config", "print", "logdir")
-      expect(subject).to receive(:archive_file_from).with(myhost,
-                                                          File.dirname(puppet_logdir),
-                                                          {}, archive_root, archive_name)
-      subject.copy_system_logs(myhost)
+  end
+  describe "#archive_system_logs" do
```
billclaytor:
```suggestion
  
  describe "#archive_system_logs" do
```

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
     end
   end
   describe "#copy_system_logs" do
```
billclaytor:
```suggestion
  
  describe "#copy_system_logs" do
```

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
...
+    let(:archive_dir) { "/var/log" }
+    let(:archive_name) { "puppet_logdir.tgz" }
+    let(:archive_path) { File.join(archive_dir, archive_name) }
+
+    context "when archiving succeeds" do
+      before { subject.instance_variable_set(:@archive_root, archive_root) }
+      before { allow(subject).to receive(:archive_system_logs).with(host).and_return(archive_path) }
+      before { allow(subject).to receive(:scp_from) }
+      before { allow(FileUtils).to receive(:mkdir_p) }
+      it "calls scp_from to copy archive path to archive root" do
```
billclaytor:
```suggestion
      
      it "calls scp_from to copy archive path to archive root" do
```

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
...
+        expected_dest = File.join(archive_root, host)
+        expect(subject).to receive(:scp_from).with(host, archive_path, expected_dest)
+        subject.copy_system_logs(host)
+      end
+      it "calls mkdir_p to create the local destination path" do
```
billclaytor:
```suggestion
      
      it "calls mkdir_p to create the local destination path" do
```

billclaytor:
```suggestion
      
      it "calls mkdir_p to create the local destination path" do
```

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
...
+        expect(FileUtils).to receive(:mkdir_p)
+        subject.copy_system_logs(host)
+      end
+    end
+    context "when archiving fails" do
```
billclaytor:
```suggestion
    
    context "when archiving fails" do
```

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
...
+      before { allow(subject).to receive(:archive_system_logs).with(host).and_return(nil) }
+      it "scp_from is not called" do
+        expect(subject).not_to receive(:scp_from)
+        subject.copy_system_logs(host)
+      end
+      it "mkdir_p is not called" do
```
billclaytor:
```suggestion
      
      it "does not call mkdir_p" do
```

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -439,32 +439,68 @@ class PerfRunHelperClass
...
+        subject.copy_system_logs(host)
+      end
+      it "calls mkdir_p to create the local destination path" do
+        expect(FileUtils).to receive(:mkdir_p)
+        subject.copy_system_logs(host)
+      end
+    end
+    context "when archiving fails" do
+      before { allow(subject).to receive(:archive_system_logs).with(host).and_return(nil) }
+      it "scp_from is not called" do
```
billclaytor:
```suggestion
      
      it "does not call scp_from" do
```

spec/tests/helpers/perf_run_helper_spec.rb
johnduarte: This suggestion reduced the indentation of the describe line by one space.  Please add the space back to the line.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -438,33 +438,76 @@ class PerfRunHelperClass
...
-      allow(subject).to receive(:on).with(myhost, any_args).and_return(result)
-      expect(subject).to receive(:puppet).with("config", "print", "logdir")
-      expect(subject).to receive(:archive_file_from).with(myhost,
-                                                          File.dirname(puppet_logdir),
-                                                          {}, archive_root, archive_name)
-      subject.copy_system_logs(myhost)
+    context "when archiving fails" do
+      it "returns nil" do
+        expect(subject).to receive(:archive_system_logs).with(host).and_return(nil)
+        subject.archive_system_logs(host)
```
RandellP: Same as above.  I think this mocks out the method under test.
I think this should be
expect(subject.archive_system_logs(host)).to eq(nil)  
or something like it.  nil might be a special case...

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -438,33 +438,76 @@ class PerfRunHelperClass
...
+      before { allow(FileUtils).to receive(:mkdir_p) }
+
+      it "calls scp_from to copy archive path to archive root" do
+        expected_dest = File.join(archive_root, host)
+        expect(subject).to receive(:scp_from).with(host, archive_path, expected_dest)
+        subject.copy_system_logs(host)
+      end
+
+      it "calls mkdir_p to create the local destination path" do
+        expect(FileUtils).to receive(:mkdir_p)
```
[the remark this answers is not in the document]
johnduarte: You are quite correct.  Update pushed to resolve this omission.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -438,33 +438,76 @@ class PerfRunHelperClass
...
+    let(:archive_path) { File.join(archive_dir, archive_name) }
+
+    context "when archiving succeeds" do
+      before { subject.instance_variable_set(:@archive_root, archive_root) }
+      before { allow(subject).to receive(:archive_system_logs).with(host).and_return(archive_path) }
+      before { allow(subject).to receive(:scp_from) }
+      before { allow(FileUtils).to receive(:mkdir_p) }
+
+      it "calls scp_from to copy archive path to archive root" do
+        expected_dest = File.join(archive_root, host)
```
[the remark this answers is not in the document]
johnduarte: The double object resolves to its `to_s` value.  So, the actual value passed around is "/tmp/test/archive/#[Double (anonymous)]".

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -438,33 +438,76 @@ class PerfRunHelperClass
...
+  describe "#archive_system_logs" do
+    let(:host) { double.as_null_object }
+    let(:puppet_logdir) { "/var/log/puppetlabs/puppet" }
+    let(:archive_dir) { "/var/log" }
+    let(:archive_name) { "puppet_logdir.tgz" }
+    let(:archive_path) { File.join(archive_dir, archive_name) }
+
+    context "when archiving succeeds" do
+      it "returns archive path" do
+        expect(subject).to receive(:archive_system_logs).with(host).and_return(archive_path)
```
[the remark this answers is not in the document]
johnduarte: You are absolutely correct. Thanks @RandellP .  Update pushed to resolve this.

spec/tests/helpers/perf_run_helper_spec.rb
johnduarte: Indeed.  Update pushed to resolve this.

spec/tests/helpers/perf_run_helper_spec.rb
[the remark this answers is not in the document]
RandellP: interesting... didn't know that. :)

PR #354, opened 2019-11-11
(SLV-671) make current_settings.json per host

RandellP:
created hosts dir in results, and a subdir by host role.  Then in there a subdir by host ip (since there may be more then one).
In that dir placed the current_settings.json for each host

### DIFF
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -27,7 +27,7 @@ module PerfRunHelper
   SCALE_RESULTS_DIR = "results/scale"
 
   METRIC_RESULTS_DIR = "/root/gatling-puppet-load-test/simulation-runner/results"
-  CURRENT_TUNE_SETTINGS = "current_tune_settings.json"
+  CURRENT_TUNE_SETTINGS_FILENAME = "current_tune_settings.json"
 
   # Performs the following steps:
   # - set timestamps
@@ -85,6 +85,7 @@ module PerfRunHelper
   end
 
   # Sets the results timestamp and creates the 'archive root' directory
+  # as well as host subdirs
   #
   # @author Bill Claytor
   #
@@ -96,10 +97,40 @@ module PerfRunHelper
   def create_perf_archive_root
     @gplt_timestamp = Time.now.getutc.to_i
     @archive_root = "#{PERF_RESULTS_DIR}/PERF_#{@gplt_timestamp}"
+    # Separate mkdir's for spec test clarity
     FileUtils.mkdir_p @archive_root
+    FileUtils.mkdir_p "#{@archive_root}/hosts"
+    hosts.each do |host|
+      role = get_most_relevant_role(host)
+      role_dir = "#{@archive_root}/hosts/#{role}"
+      FileUtils.mkdir_p("#{role_dir}/#{host.hostname}")
+    end
+  end
+
+  # Pics out the most relevant role from the hosts roles
+  #
+  # @author Randell Pelak
+  #
+  # @param [Object] host The beaker host object
+  #
+  # @return [string] role
+  #
+  # @example:
+  #   get_most_relevant_role(host)
+  #
+  def get_most_relevant_role(host)
+    roles = host[:roles]
+    return "master" if roles.include?("master")
+    return "metric" if roles.include?("metric")
+    return "compiler" if roles.include?("compile_master")
+    return "database" if roles.include?("database")
+    return "loadbalancer" if roles.include?("loadbalancer")
+
+    return "unknown"
   end
 
   # Get the current values for the settings that can be tuned with pe_tune
+  # places them in a file in each hosts subdir in archive root
   #
   # @author Bill Claytor
   #
@@ -109,12 +140,13 @@ module PerfRunHelper
   #   capture_current_tune_settings
   #
   def capture_current_tune_settings
-    puts "Checking current tune settings..."
-    settings_json = run_script_on(master, "./util/tune/current_settings.rb").output
-    output_path = "#{@archive_root}/#{CURRENT_TUNE_SETTINGS}"
-
-    puts "Writing file: #{output_path}"
-    File.write(output_path, settings_json)
+    puts "Generating current tune settings files for each host..."
+    hosts.each do |host|
+      settings_json = run_script_on(host, "./util/tune/current_settings.rb").output
+      role = get_most_relevant_role(host)
+      output_path = "#{@archive_root}/hosts/#{role}/#{host.hostname}/#{CURRENT_TUNE_SETTINGS_FILENAME}"
+      File.write(output_path, settings_json)
+    end
   end
 
   # Write out the epoch into a file for info and use when gathering metrics
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -15,6 +15,26 @@ TEST_ARCHIVE_ROOT = "#{TEST_DIR}/PERF_#{TEST_TIMESTAMP}"
 TEST_JSON = "{'parameter': 'value'}"
 
 describe PerfRunHelperClass do # rubocop:disable Metrics/BlockLength
+  let(:hosts) do
+    [Beaker::Host.create("ip_master",
+                         { platform: Beaker::Platform.new("centos-6.5-x86_64"),
+                           role: "master" }, logger: @logger),
+     Beaker::Host.create("ip_metric",
+                         { platform: Beaker::Platform.new("centos-6.5-x86_64"),
+                           role: "metric" }, logger: @logger),
+     Beaker::Host.create("ip_compile_master_A",
+                         { platform: Beaker::Platform.new("centos-6.5-x86_64"),
+                           role: "compile_master" }, logger: @logger),
+     Beaker::Host.create("ip_compile_master_B",
+                         { platform: Beaker::Platform.new("centos-6.5-x86_64"),
+                           role: "compile_master" }, logger: @logger),
+     Beaker::Host.create("ip_database",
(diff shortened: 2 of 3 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -93,13 +94,42 @@ def perf_teardown
...
+  #
+  def get_most_relevant_role(host)
+    roles = host[:roles]
+    return "master" if roles.include?("master")
+    return "metric" if roles.include?("metric")
+    return "compiler" if roles.include?("compile_master")
+    return "database" if roles.include?("database")
+    return "loadbalancer" if roles.include?("loadbalancer")
+
+    return "unknown"
```
johnduarte: This seems like a weird value to return.  If the given role set for a host is not one of the ones in the list that we care about I would expect that we would return on of the roles actually assigned to the host rather than unknown.  For example, when we get to true XL we will have a 'postgres' role.  It would be helpful to have the host tree reflect that value rather than 'unknown'.

I understand that there is no way to determine a suitable candidate out of an arbitrary list of roles. Perhaps they could be concatenated?

tests/helpers/perf_run_helper.rb
```diff
@@ -93,13 +94,42 @@ def perf_teardown
   # @example:
   #   create_perf_archive_root
   #
-  def create_perf_archive_root
+  def create_perf_archive_area
```
johnduarte: :-1: for this breaking change.
The intent of this method has not changed.  The `@archive_root` directory is still created by this method it just does more work.  I do not think there is enough change to warrant altering this method name.

tests/helpers/perf_run_helper.rb
RandellP: Unknown signals that the host isn't one of the currently "relevant" roles.  If I returned the first role on the list when none matched, than it wouldn't be clear to the user that the host doesn't have what is classified as a relevant role (and that likely the code needs updating).   Also if I just grab the first one... it could potentially be different from run to run... order of the roles is not guaranteed. I could change it to none since the host didn't have any of the relevant roles, but I didn't think that looked as good in the directory tree and unknown.

tests/helpers/perf_run_helper.rb
RandellP: Since the method is only used internal to this file, it didn't feel like a breaking change.  And while the old name is still true, if you read the code for the first time with no background history, you would think the name is wrong.  That is what I was trying to fix... we can discuss in the parking lot

tests/helpers/perf_run_helper.rb
johnduarte: My suggestion was to concatenate the roles not to pick the first one from the list.  As a user, how do I know what roles are considered "relevant" by the code paths that this method may be used in?

RandellP: Good point.  How about instead of making a really long directory name (some have like 5 or 6 roles), instead I keep the unknown, and add a file in the host specific dir outlining what the roles were, and a auto generated message about what are currently considered relevant.   kinda like a debug log.  in fact I can call it role_debug.log... or could call it why_unknown.log... thoughts?  @billclaytor any thoughts on this?

johnduarte: :-1: on another file.  This information is captured in the beaker configuration.

It sounds like we don't have a great solution to this case.  So, I am fine with using "unknown" for now and ticketing a future spike to try and resolve it.

tests/helpers/perf_run_helper.rb
```diff
@@ -97,9 +98,38 @@ def create_perf_archive_root
     @gplt_timestamp = Time.now.getutc.to_i
     @archive_root = "#{PERF_RESULTS_DIR}/PERF_#{@gplt_timestamp}"
     FileUtils.mkdir_p @archive_root
```
billclaytor: It seems like the line:
```
FileUtils.mkdir_p @archive_root
```
is no longer necessary because:
```
FileUtils.mkdir_p "#{@archive_root}/hosts"
```
will create it as well.

tests/helpers/perf_run_helper.rb
billclaytor: What if there are multiple hosts that wind up with 'unknown'?

tests/helpers/perf_run_helper.rb
```diff
@@ -109,12 +139,13 @@ def create_perf_archive_root
...
-    settings_json = run_script_on(master, "./util/tune/current_settings.rb").output
-    output_path = "#{@archive_root}/#{CURRENT_TUNE_SETTINGS}"
-
-    puts "Writing file: #{output_path}"
-    File.write(output_path, settings_json)
+    puts "Generating current tune settings files for each host..."
+    hosts.each do |host|
+      settings_json = run_script_on(host, "./util/tune/current_settings.rb").output
+      role = get_most_relevant_role(host)
+      output_path = "#{@archive_root}/hosts/#{role}/#{host.hostname}/#{CURRENT_TUNE_SETTINGS}"
```
billclaytor: I know I added `CURRENT_TUNE_SETTINGS` but maybe it should be updated to `CURRENT_TUNE_SETTINGS_FILENAME` to clarify?

tests/helpers/perf_run_helper.rb
RandellP: under the role there is a dir that is the hostname... so they would each get there own.  I hated doing that, but compilers will have multiple, so we needed another dir level anyway.

tests/helpers/perf_run_helper.rb
RandellP: yeah, I tried merging them... but the resulting changes to the spec tests made them less clear.  Like I couldn't have a spec test to ensure the archive root alone was created, which is a critical dep for other things... Because the call would be for hosts not just archive root... So kinda have to choose, clearer spec test names and such, or one less call the mkdir_p.  Thoughts?  I could really go either way on it.

tests/helpers/perf_run_helper.rb
RandellP: updated

billclaytor:
I get the following errors:
Standard:
```
ip-10-227-1-17.amz-dev.puppet.net (perf-test-metrics) 14:01:51$ /tmp/current_settings.rb
  /tmp/current_settings.rb:220:in `exist?': no implicit conversion of nil into String (TypeError)
        from /tmp/current_settings.rb:220:in `get_conf_parameter'
        from /tmp/current_settings.rb:250:in `get_postgres_parameter'
        from /tmp/current_settings.rb:283:in `database_shared_buffers'
        from /tmp/current_settings.rb:385:in `current_settings'
        from /tmp/current_settings.rb:410:in `<main>'

ip-10-227-1-17.amz-dev.puppet.net (perf-test-metrics) executed in 0.24 seconds
Exited: 1
Beaker::Host::CommandFailure: Host 'ip-10-227-1-17.amz-dev.puppet.net' exited with 1 running:
 /tmp/current_settings.rb
Last 10 lines of output were:
        /tmp/current_settings.rb:220:in `exist?': no implicit conversion of nil into String (TypeError)
                from /tmp/current_settings.rb:220:in `get_conf_parameter'
                from /tmp/current_settings.rb:250:in `get_postgres_parameter'
                from /tmp/current_settings.rb:283:in `database_shared_buffers'
                from /tmp/current_settings.rb:385:in `current_settings'
                from /tmp/current_settings.rb:410:in `<main>'
```

Large:
```
Generating current tune settings files for each host...
localhost $ scp ./util/tune/current_settings.rb ip-10-227-1-93.amz-dev.puppet.net:/tmp/current_settings.rb {:ignore => }

ip-10-227-1-93.amz-dev.puppet.net 14:18:36$ /tmp/current_settings.rb
  /tmp/current_settings.rb:220:in `exist?': no implicit conversion of nil into String (TypeError)
        from /tmp/current_settings.rb:220:in `get_conf_parameter'
        from /tmp/current_settings.rb:250:in `get_postgres_parameter'
        from /tmp/current_settings.rb:283:in `database_shared_buffers'
        from /tmp/current_settings.rb:385:in `current_settings'
        from /tmp/current_settings.rb:410:in `<main>'

ip-10-227-1-93.amz-dev.puppet.net executed in 0.13 seconds
Exited: 1
Beaker::Host::CommandFailure: Host 'ip-10-227-1-93.amz-dev.puppet.net' exited with 1 running:
 /tmp/current_settings.rb
Last 10 lines of output were:
        /tmp/current_settings.rb:220:in `exist?': no implicit conversion of nil into String (TypeError)
                from /tmp/current_settings.rb:220:in `get_conf_parameter'
                from /tmp/current_settings.rb:250:in `get_postgres_parameter'
                from /tmp/current_settings.rb:283:in `database_shared_buffers'
                from /tmp/current_settings.rb:385:in `current_settings'
                from /tmp/current_settings.rb:410:in `<main>'
```

tests/helpers/perf_run_helper.rb
billclaytor: That makes sense, thanks. Could you please add a comment to the code explaining that so someone reading / refactoring in the future doesn't get this same idea?

RandellP:
@billclaytor Fixed the bug in current_settings that was causing the issue on hosts that didn't have postgres installed.  And made the rest of the changes you requested.  Take another look please.

tests/helpers/perf_run_helper.rb
RandellP: added

RandellP:
@billclaytor  can you check the responses to your feedback.
@jarretlavallee  now that you are back home... Do you want to be part of these reviews or should we review and merge with just SLV members?

jarretlavallee:
@RandellP thanks for including me. I would be happy to be a part of the reviews. I haven't used this repo yet, so I will take a look today.

RandellP:
whoops @jarretlavallee  I put that on the wrong PR... you don't want to look at this repo... :)  it's a mess.  That was supposed to be for the puppet metrics collector pr...

PR #357, opened 2019-11-15
(SLV-702) handle duplicate params in conf files

RandellP:
Conf files can have the same parameter defined multiple times.  If they are the last one wins.  The code change here sets us up to get the last one instead of the current one.
That change made it also a good time to remove the exclusion mechanism since it would have to have been refactored to continue to function.
Also Fixed the reporting on the max jruby instances to be NA if there was no puppetserver conf file instead of reporting the calculated default jrubies.

### DIFF
--- a/util/tune/current_settings.rb
+++ b/util/tune/current_settings.rb
@@ -27,7 +27,10 @@ DEFAULT_EXCLUSION = ""
 DEFAULT_WORKING_DIRECTORY = "/root/tmp"
 DEFAULT_OUTPUT_FILE = "current_tune_settings.json"
 
-# Returns the value for the 'max-active-instances' parameter if found, otherwise the default value
+# Returns the value for the 'max-active-instances' parameter if found in the conf file.
+# If the conf file doesn't exist returns NA as puppetserver must not be installed on this host.
+# Otherwise calculates the default based on the documentation linked above which was current at the
+# time this was written.  It could have changed, but probably not.
 #
 # See the following documentation:
 #   https://puppet.com/docs/pe/2019.1/config_puppetserver.html#tune-the-maximum-number-of-jruby-instances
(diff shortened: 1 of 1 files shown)


util/tune/current_settings.rb
```diff
@@ -40,10 +43,10 @@
 #   value = puppetserver_jruby_max_active_instances
 #
 def puppetserver_jruby_max_active_instances
-  if File.exist? PE_PUPPET_SERVER_CONF
-    line = File.open(PE_PUPPET_SERVER_CONF).grep(/max-active-instances/)[0]
-    value = line.scan(/\d+/).first unless line.nil?
-  end
+  return "NA" unless File.exist? PE_PUPPET_SERVER_CONF
```
johnduarte: NA is defined as a constant at the beginning of this script.  The constant should be used here.

util/tune/current_settings.rb
```diff
@@ -216,15 +220,12 @@ def puppetdb_java_args
 # TODO: update to eliminate the use of exclusions
 #   See https://tickets.puppetlabs.com/browse/SLV-531
 #
-def get_conf_parameter(file, parameter, exclusions = DEFAULT_EXCLUSION)
+def get_conf_parameter(file, parameter)
```
johnduarte: Remove `exclusions` from the YARD doc for this method.

util/tune/current_settings.rb
```diff
@@ -216,15 +220,12 @@ def puppetdb_java_args
...
   value = if File.exist? file
-
-            # this pattern extracts the value from the result with the '\K'
-            match_pattern = /#{parameter} = \K[^\s]+/
-
-            # eliminate comments, search for the parameter, apply exclusions, extract the value
-            File.open(file).grep_v(/^#/).grep(/#{parameter}/).grep_v(exclusions)[0].match(match_pattern)
-
+            conf = File.read(file)
```
johnduarte: This will throw an exception if the file doesn't exist.  Is that the expected behavior, or should the error be eaten to allow the `NA` value to be returned by this method?

util/tune/current_settings.rb
```diff
@@ -40,10 +43,10 @@
...
 #
 def puppetserver_jruby_max_active_instances
-  if File.exist? PE_PUPPET_SERVER_CONF
-    line = File.open(PE_PUPPET_SERVER_CONF).grep(/max-active-instances/)[0]
-    value = line.scan(/\d+/).first unless line.nil?
-  end
+  return "NA" unless File.exist? PE_PUPPET_SERVER_CONF
+
+  line = File.open(PE_PUPPET_SERVER_CONF).grep(/max-active-instances/)[0]
+  value = line.scan(/\d+/).first unless line.nil?
```
johnduarte: Should this be changed to `value = line.scan(/\d+/).last unless line.nil?` ??  The commit comment says that the last value in a duplicated list should win.  I may not understand what this line does.

util/tune/current_settings.rb
RandellP: It's inside an if file.exists? block, so the file must exist.

util/tune/current_settings.rb
RandellP: yeah, I just changed the indent of this.  I didn't really look at this part.  But it is totally wrong.  It doesn't ignore comments...  so I will have to rework it.

util/tune/current_settings.rb
RandellP: done

RandellP:
@johnduarte ready for re-review

RandellP:
@billclaytor looks like I need a re-review from you as well... I think you approved previously, but if you did it cleared it... probably due to the changes made since then.

util/tune/current_settings.rb
```diff
@@ -211,62 +215,44 @@ def puppetdb_java_args
...
+  return NA unless File.exist? file
 
-          else
-            NA
-          end
-  value
+  conf = File.read(file)
+  match_pattern = /^\s*#{parameter} = (\S+)/
+  conf_scan = conf.scan(match_pattern)
+  conf_scan.last ? conf_scan.last[0] : NA
```
johnduarte: :pick: could you change the index number to `first` to prevent the mixing of `first/last` and index references?

RandellP: updated

util/tune/current_settings.rb
```diff
@@ -40,29 +43,29 @@
...
-    else
-      value = NA
-    end
-
+  return NA unless File.exist? PE_PUPPET_SERVER_CONF
+
+  conf = File.read(PE_PUPPET_SERVER_CONF)
+  match_pattern = /^\s*max-active-instances: (\d+)/
+  conf_scan = conf.scan(match_pattern)
+  value = conf_scan.last ? conf_scan.last[0] : nil
```
billclaytor: Does John's request for line 225 (https://github.com/puppetlabs/gatling-puppet-load-test/pull/357#discussion_r348170034) apply here as well?

util/tune/current_settings.rb
```diff
@@ -40,29 +43,29 @@
...
+  return value unless value.nil?
```
billclaytor: Should the code below be extracted into a separate method to calculate the value and potentially eliminate this extra exit point?

Something like this (not tested):
```
def puppetserver_jruby_max_active_instances
  return NA unless File.exist? PE_PUPPET_SERVER_CONF

  conf = File.read(PE_PUPPET_SERVER_CONF)
  match_pattern = /^\s*max-active-instances: (\d+)/
  conf_scan = conf.scan(match_pattern)
  value = conf_scan.last ? conf_scan.last[0] : calculate_jruby_max_active_instances
  
  value
end

def calculate_jruby_max_active_instances
  # TODO: eliminate bash
  command = "facter processorcount"
  output = `#{command}`

  if output
    # See the docs link in the description above
    # The default used in PE is the number of CPUs - 1, expressed as $::processorcount - 1.
    # One instance is the minimum value and four instances is the maximum value.
    #
    # TODO: Determine if / where the default value for the 'jruby_max_active_instances' is set
    #   See https://tickets.puppetlabs.com/browse/SLV-530
    num_cores = output.to_i
    value = [[(num_cores - 1), MIN_DEFAULT_JRUBIES].max, MAX_DEFAULT_JRUBIES].min.to_s
  else
    value = NA
  end

  value
end
```

billclaytor: I have a test running on the current version, so if you like the suggestion above I can drop that in and re-test in the same environment.

PR #370, opened 2019-11-23
(SLV-656) Add save_average_compile_time helper

johnduarte:
This commit adds the `save_average_compile_time` helper to to the
PerfRunHelper module.  This helper queries puppetdb for the
"config_retrieval" and "total" values associated with metrics data from
the reports endpoint.  This query is constrained to a time range bound
by the start_epoch and end_epoch instance values associated with the
code at runtime.

Spec tests for this helper are added as well.

### DIFF
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -834,7 +834,7 @@ current_tune_settings.json"
       end
     end
   end
-  describe "#save_average_compile_time" do
+  describe "#save_average_transaction_time" do
     let(:host) do
       Beaker::Host.create("localhost", { role: "database" }, logger: @logger)
     end
@@ -857,8 +857,9 @@ current_tune_settings.json"
       curl.gsub(/\s+/, " ")
     end
     context "when puppet query returns json data with config_retreival times" do
-      it "writes average compile time to the given file" do
-        compile_time = 41.534695086999933
+      it "writes average transaction time to the given file" do
+        config_retrieval_time = 41.534695086999933
+        expected_file_content = { avg_transaction_time: config_retrieval_time, unit: "seconds" }.to_json
         result.stdout = <<~JSON_DATA # {{{
           [
             {
@@ -866,7 +867,7 @@ current_tune_settings.json"
                 "data": [
                   {
                     "name": "config_retrieval",
-                    "value": #{compile_time},
+                    "value": #{config_retrieval_time},
                     "category": "time"
                   }
                 ]
@@ -876,20 +877,22 @@ current_tune_settings.json"
         JSON_DATA
         # }}}
         expect(subject).to receive(:on).with(host, curl).and_return(result)
-        subject.save_average_compile_time(file.path, host)
+        subject.save_average_transaction_time(file.path, host)
 
         file.open
-        expect(file.read.chomp).to eq(compile_time.ceil.to_s)
+        expect(file.read.chomp).to eq(expected_file_content)
       end
     end
     context "when puppet query returns no data" do
-      it "writes nothing to the given file" do
+      it "writes error to the given file" do
+        err_msg = "JSON::ParserError A JSON text must at least contain two octets!"
+        expected_file_content = { avg_transaction_time: nil, unit: "seconds", error: err_msg }.to_json
         result.stdout = ""
         expect(subject).to receive(:on).with(host, curl).and_return(result)
-        subject.save_average_compile_time(file.path, host)
+        subject.save_average_transaction_time(file.path, host)
 
         file.open
-        expect(file.read.chomp).to eq("")
+        expect(file.read.chomp).to eq(expected_file_content)
       end
     end
   end
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# vim: set foldmethod=marker
+# vim: foldmethod=marker
 
 require "spec_helper"
 require "minitest/assertions"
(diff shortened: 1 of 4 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -911,9 +913,53 @@ def get_perf_result
...
+      data.delete_if { |report| report["metrics"]["data"].empty? }
+      config_retrieval_times = data.map do |report|
+        report["metrics"]["data"].select do |md|
+          md["category"] == "time" && (md["name"] == "config_retrieval" || md["name"] == "total")
+        end.first.fetch("value")
+      end
+      avg_config_retrieval_time = config_retrieval_times.reduce(0.0, :+) / config_retrieval_times.size
+      result = avg_config_retrieval_time.ceil
+    rescue StandardError # rubocop:disable Lint/HandleExceptions
+      # empty file will be written
```
RandellP: could we write the error to the file instead of empty?

tests/helpers/perf_run_helper.rb
```diff
@@ -911,9 +913,53 @@ def get_perf_result
...
+      bkr_res = on(puppetdb_host, curl)
+      data = JSON.parse(bkr_res.stdout.chomp)
+      data.delete_if { |report| report["metrics"]["data"].empty? }
+      config_retrieval_times = data.map do |report|
+        report["metrics"]["data"].select do |md|
+          md["category"] == "time" && (md["name"] == "config_retrieval" || md["name"] == "total")
+        end.first.fetch("value")
+      end
+      avg_config_retrieval_time = config_retrieval_times.reduce(0.0, :+) / config_retrieval_times.size
+      result = avg_config_retrieval_time.ceil
```
RandellP: We probably need more than an integer...  if the unit is seconds we would usually only get about 6...  So rounding could be significant.  what kind of numbers were you getting?

johnduarte: The ceiling was pulled from Kishel's method https://github.com/tkishel/pe_tune/blob/f8270a784d2a6a96f1cd483ef95745dcf0a7e01e/lib/puppet_x/puppetlabs/tune/query.rb#L76-L102

I have no objection to using a float.

tests/helpers/perf_run_helper.rb
johnduarte: We could, however my bias is that doing so breaks the interface contract for the file (contains average compile time).  The absence of a value indicates that something when sideways.  I would counter propose that we emit the error via the logger instead to assist in determining why things went sideways.  Thoughts?

tests/helpers/perf_run_helper.rb
RandellP: yeah, Tom probably doesn't need as much precision as we probably will.  Though in the end AWS may end up making the extra precision pointless.  But we won't know until we start looking at it to see.

tests/helpers/perf_run_helper.rb
RandellP: My only concern with that is so many errors that get logged are lost in the mass of output that we get.  And I am not sure an error breaks the contract more then a nil value...  All in all I would rather have some kind of error file that errors are logged to instead of the mass of output we get from the run...

johnduarte: I have pushed an update to write a JSON payload to the file.  If an error is encountered, it will be written as part of the JSON data structure.

tests/helpers/perf_run_helper.rb
johnduarte: I have updated this to use a float value.  The naming conventions have been changed.  In my research, I discovered that the report query times are not specifically catalog compile times.  The time was discovered to be in seconds, so this has been added to the payload written to disk as well.

Ref:
https://puppet.com/docs/puppet/latest/format_report.html
https://github.com/puppetlabs/puppet/blob/master/api/schemas/report.json#L254-L275
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction.rb#L50
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/face/catalog.rb#L137-L139
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util.rb#L519-L525

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -793,5 +793,67 @@ class PerfRunHelperClass
...
     end
   end
+  describe "#save_average_transaction_time" do
+    let(:host) do
+      Beaker::Host.create("localhost", { role: "database" }, logger: @logger)
+    end
+    before { subject.instance_variable_set(:@start_epoch, 0) }
+    before { subject.instance_variable_set(:@end_epoch, 946_684_799) }
+    let(:file) { Tempfile.new(described_class.to_s) }
+    let(:result) { Beaker::Result.new(host, "") }
```
RandellP: should this be in a before or something to ensure it is new for each test?

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -793,5 +793,67 @@ class PerfRunHelperClass
       end
     end
   end
+  describe "#save_average_transaction_time" do
+    let(:host) do
+      Beaker::Host.create("localhost", { role: "database" }, logger: @logger)
+    end
+    before { subject.instance_variable_set(:@start_epoch, 0) }
+    before { subject.instance_variable_set(:@end_epoch, 946_684_799) }
+    let(:file) { Tempfile.new(described_class.to_s) }
```
RandellP: Looks like the tests are sharing this file.  So I think that it is possible for one test to read the file the other has written.  That seems bad.

johnduarte: This appears to be lazily evaluated prior to each test run, so the file is unique each time.

```
Run options: include {:locations=>{"./spec/tests/helpers/perf_run_helper_spec.rb"=>[796]}}

From: /home/john/Documents/src/jrd-gatling-puppet-load-test/spec/tests/helpers/perf_run_helper_spec.rb @ line 842
:
    841:         binding.pry
 => 842:         file.open
> file
=> #<File:/tmp/PerfRunHelperClass20191126-17557-hqstb2>

From: /home/john/Documents/src/jrd-gatling-puppet-load-test/spec/tests/helpers/perf_run_helper_spec.rb @ line 855
:
    854:         binding.pry
 => 855:         file.open
)> file
=> #<File:/tmp/PerfRunHelperClass20191126-17557-1l9hk7b>
```

spec/tests/helpers/perf_run_helper_spec.rb
johnduarte: This is just to get a Beaker::Result object.  Each test explicitly defines the `stdout` for it, so I don't think it is an issue.

PR #372, opened 2019-11-25
(SLV-708) Skip better performance for memory

johnduarte:
This commit updates the logic for squelching failures that are positive.
These should only be throttled for memory values, in the sense that less
is better than more.  Other relationships do not require suppression at
this time.  This logic should be removed when SLV-716 is resolved.

### DIFF
--- a/spec/tests/helpers/perf_run_helper_spec.rb
+++ b/spec/tests/helpers/perf_run_helper_spec.rb
@@ -763,12 +763,38 @@ current_tune_settings.json"
                                               "pass_thing2" => [100, 102])).to eq(false)
       end
     end
+    context "when any delta is > MAX_BASELINE_VARIANCE other direction" do
+      it "returns false" do
+        expect(subject.validate_baseline_data("fail_thing"  => [2, 1],
+                                              "pass_thing2" => [100, 102])).to eq(false)
+      end
+    end
     context "when MAX_BASELINE_VARIANCE < 'orchestration_service memory delta' > MAX_BASELINE_VARIANCE_ORCH_REL_MEM" do
       it "returns true" do
         expect(
           subject.validate_baseline_data(
             "pass_thing1"                                    => [1.00, 0.98],
             "process_orchestration_services_release_avg_mem" => [100, 112]
+          )
+        ).to eq(true)
+      end
+    end
+    context "when 'orchestration_service memory delta' < MAX_BASELINE_VARIANCE_ORCH_REL_MEM AND worse" do
+      it "returns false" do
+        expect(
+          subject.validate_baseline_data(
+            "pass_thing1"                                    => [1.00, 0.98],
+            "process_orchestration_services_release_avg_mem" => [100, 150]
+          )
+        ).to eq(false)
+      end
+    end
+    context "when 'orchestration_service memory delta' < MAX_BASELINE_VARIANCE_ORCH_REL_MEM AND better" do
+      it "returns true" do
+        expect(
+          subject.validate_baseline_data(
+            "pass_thing1"                                    => [1.00, 0.98],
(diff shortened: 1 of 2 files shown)


spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -773,6 +779,26 @@ class PerfRunHelperClass
         ).to eq(true)
       end
     end
+    context "when 'orchestration_service memory delta' < MAX_BASELINE_VARIANCE_ORCH_REL_MEM AND worse" do
+      it "returns false" do
+        expect(
+          subject.validate_baseline_data(
+            "pass_thing1"                                    => [1.00, 0.98],
```
RandellP: I can't find MAX_BASELINE_VARIANCE_ORCH_REL_MEM defined anywhere...
I think it changed to MAX_VARIANCE_OVERRIDE_PROCESS_ORCHESTRATION_SERVICES_RELEASE_AVG_MEM
And the hardcoded numbers should probably be calculated based on it so that if it changes the tests don't have to change too.

johnduarte: :+1: Spec tests updated to use the modules constants to calculate the variances. Fixed the incorrect reference.

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -746,31 +746,72 @@ class PerfRunHelperClass
...
+      it "returns false" do
+        baselines = [1.00, 100]
+        data = { pass_lower: [baselines[0],
+                              baselines[0] - baselines[0] * PerfRunHelper::MAX_BASELINE_VARIANCE * 0.9],
+                 process_orchestration_services_release_avg_mem: [baselines[1],
+                                                                  baselines[1] + baselines[1] * PerfRunHelper::MAX_VARIANCE_OVERRIDE_PROCESS_ORCHESTRATION_SERVICES_RELEASE_AVG_MEM * 1.1] } # rubocop:disable  Metrics/LineLength
+        expect(subject.validate_baseline_data(data)).to eq(false)
+      end
+    end
+    context "when 'orchestration_service memory variance' greateer than max override AND better" do
```
RandellP: typo: greateer  => greater

spec/tests/helpers/perf_run_helper_spec.rb
```diff
@@ -746,31 +746,72 @@ class PerfRunHelperClass
...
+        expect(subject.validate_baseline_data(data)).to eq(true)
+      end
+    end
+    context "when 'orchestration_service memory variance' greater than max override AND worse" do
+      it "returns false" do
+        baselines = [1.00, 100]
+        data = { pass_lower: [baselines[0],
+                              baselines[0] - baselines[0] * PerfRunHelper::MAX_BASELINE_VARIANCE * 0.9],
+                 process_orchestration_services_release_avg_mem: [baselines[1],
+                                                                  baselines[1] + baselines[1] * PerfRunHelper::MAX_VARIANCE_OVERRIDE_PROCESS_ORCHESTRATION_SERVICES_RELEASE_AVG_MEM * 1.1] } # rubocop:disable  Metrics/LineLength
```
billclaytor: Since `PerfRunHelper::MAX_VARIANCE_OVERRIDE_PROCESS_ORCHESTRATION_SERVICES_RELEASE_AVG_MEM` is being used twice and requires disabling the LineLength cop would it make sense to assign this to a shorter-named variable here to improve readability and satisfy the cop?

johnduarte: :-1: that seems like the tail wagging the dog to me.  Such a variable would obfuscate the constant.

PR #373, opened 2019-11-25
(SLV-717) Increate 'num_repetitions' to 2 in Scale.json scenario

billclaytor:
This update increases the number of repetitions per scale test iteration from 1 to 2.

### DIFF
--- a/tests/helpers/perf_run_helper.rb
+++ b/tests/helpers/perf_run_helper.rb
@@ -461,25 +461,40 @@ module PerfRunHelper
     file = File.read("#{scenarios_dir}/#{gatling_scenario}")
     json = JSON.parse(file)
 
-    # allow the base instances to be set via environment variable
+    # allow the settings to be set via environment variables
+    # TODO: refactor to streamline handling of multiple environment variables / settings
+    env_scale_class = ENV["PUPPET_SCALE_CLASS"]
     env_base_instances = ENV["PUPPET_GATLING_SCALE_BASE_INSTANCES"]
+    env_num_repetitions = ENV["PUPPET_GATLING_SCALE_NUM_REPETITIONS"]
     json_base_instances = json["nodes"][0]["num_instances"]
+    json_num_repetitions = json["nodes"][0]["num_repetitions"]
 
-    # TODO: refactor
     if !env_base_instances.nil?
       puts "Using environment specified base instances: #{env_base_instances}"
       @scale_base_instances = Integer(env_base_instances)
-
-      # update json
-      desc = "'role::by_size_small' role from perf control repo, #{@scale_base_instances} agents, 1 iteration"
-      json["run_description"] = desc
-      json["nodes"][0]["num_instances"] = @scale_base_instances
     else
       puts "Using JSON specified base instances: #{json_base_instances}"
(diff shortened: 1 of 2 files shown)


tests/helpers/perf_run_helper.rb
```diff
@@ -461,25 +461,40 @@ def generate_scale_scenarios(gatling_scenario)
     file = File.read("#{scenarios_dir}/#{gatling_scenario}")
     json = JSON.parse(file)
 
-    # allow the base instances to be set via environment variable
+    # allow the settings to be set via environment variables
+    # TODO: refactor to streamline handling of multiple environment variables / settings
```
johnduarte: What is the use case for this?  Is a ticket needed?

tests/helpers/perf_run_helper.rb
```diff
@@ -461,25 +461,40 @@ def generate_scale_scenarios(gatling_scenario)
...
+      @scale_num_repetitions = json_num_repetitions
+    end
+
+    # instances will be incremented per iteration
     instances = @scale_base_instances
+    desc = "'#{env_scale_class}' role from perf control repo, " \
+           "#{instances} agents, #{@scale_num_repetitions} repetition(s)"
+
+    # update JSON (this was previously conditional)
+    # TODO: refactor to update before each iteration vs. after
```
johnduarte: Why not do this now?

tests/helpers/perf_run_helper.rb
```diff
@@ -461,25 +461,40 @@ def generate_scale_scenarios(gatling_scenario)
...
-      json["run_description"] = desc
-      json["nodes"][0]["num_instances"] = @scale_base_instances
     else
       puts "Using JSON specified base instances: #{json_base_instances}"
       @scale_base_instances = json_base_instances
     end
 
+    if !env_num_repetitions.nil?
+      puts "Using environment specified number of repetitions: #{env_num_repetitions}"
+      @scale_num_repetitions = Integer(env_num_repetitions)
```
johnduarte: Why is this an instance variable?

tests/helpers/perf_run_helper.rb
billclaytor: I was really just noting that it seems like this could be streamlined now that multiple environment variables are being handled here in a similar way. Given that we have the agent simulator on the horizon it doesn't seem to make sense to write (and groom) a ticket for work that may never be done. If we reach a point where we are sticking with GPLT and looking for opportunities for improvement we can generate tickets from the TODOs.

tests/helpers/perf_run_helper.rb
billclaytor: This is how the existing code works (not sure why I coded it this way in the initial implementation) and this update seems to be outside the scope of this ticket. I brought this ticket into the sprint to get the functionality implemented ASAP to avoid blocking scale testing and wanted to ensure it works as intended first. I noted it as a TODO as an opportunity for potential future improvement.

tests/helpers/perf_run_helper.rb
billclaytor: Good question... I've updated `scale_setup` to make use of it :)

PR #386, opened 2020-01-08
(SLV-666) Add import_metrics_files.rb script

billclaytor:
This update adds the `import_metrics_files.rb` script which implements the design specified in the "[Importing puppet-metrics-collector data](https://docs.google.com/document/d/1X9sKBsMHEhd2FSpnALBm1lfUOyATQkGDEkq1G95nxek/edit?usp=sharing)" document.

The script was used to import the A2A test results on the Large ref arch for SLV-667:
Kearney:
http://10.234.4.46:3000/d/VJsAGxMZk/archive-puppetserver-performance?orgId=1&from=1578074136228&to=1578089465634&var-server=slv-667-k_1578074520_ip-10-227-0-164.amz-dev.puppet.net&var-server=slv-667-k_1578074520_ip-10-227-1-0.amz-dev.puppet.net&var-server=slv-667-k_1578074520_ip-10-227-2-221.amz-dev.puppet.net

Lovejoy:
http://10.234.4.46:3000/d/VJsAGxMZk/archive-puppetserver-performance?orgId=1&var-server=slv-667-l_1578078239_ip-10-227-3-132.amz-dev.puppet.net&var-server=slv-667-l_1578078239_ip-10-227-0-116.amz-dev.puppet.net&var-server=slv-667-l_1578078239_ip-10-227-2-123.amz-dev.puppet.net&from=1578077696376&to=1578093240680


util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,165 @@
...
+  # @attr [string] output_dir The directory to place the tarfile in.
+  # @attr [string] tar_file_name Name of the tar file to create with the metrics
+  # @attr [boolean] verbose Verbose output
+  # @attr [integer] poll_interval The "collection_frequency" of the metrics collector puppet module
+  # @attr [integer] archive_interval The frequency of the *_metrics_dir cron job for the metrics
+  #   collector module.  This is how often it tars up the json files.
+  #
+  class ImportMetricsFiles
+    # TODO: make this optional?
+    JSON2GRAPHITE_PATH = "~/git/puppet-metrics-viewer/json2graphite.rb"
```
johnduarte: TechDebt alert: This is very brittle.  Explicitly expecting that this would be installed in a particular location outside of the context of the current repo is very presumptuous.  (Yes, I know that the gplt setup module does this currently.).

Should this repo be added as a submodule to gplt?

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,165 @@
...
+  # @attr [integer] end_epoch Epoch seconds for the end of the collection window
+  # @attr [string] metrics_dir The "output directory" of the metrics collector puppet module
+  # @attr [string] output_dir The directory to place the tarfile in.
+  # @attr [string] tar_file_name Name of the tar file to create with the metrics
+  # @attr [boolean] verbose Verbose output
+  # @attr [integer] poll_interval The "collection_frequency" of the metrics collector puppet module
+  # @attr [integer] archive_interval The frequency of the *_metrics_dir cron job for the metrics
+  #   collector module.  This is how often it tars up the json files.
+  #
+  class ImportMetricsFiles
```
johnduarte: There are no spec tests for this class in this pr.

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,165 @@
...
+    # @return [void]
+    #
+    # @example
+    #   import_metrics_files
+    #
+    def import_metrics_files_for_host_dir(host_dir)
+      hostname = File.basename host_dir
+      server_tag = "#{@prefix}_#{@id}_#{hostname}"
+      pattern = "'#{host_dir}/*.json'"
+      cmd = "ruby #{JSON2GRAPHITE_PATH} --pattern #{pattern}" \
```
johnduarte: Which ruby is this going to use (system or defined ruby for env)?

util/metrics/import_metrics_files.rb
billclaytor: Sorry; I meant to add a note that spec tests would be added once the implementation was approved. I wanted to avoid having to update them if there were significant updates to the implementation based on feedback.

util/metrics/import_metrics_files.rb
billclaytor: I would prefer to make this an optional parameter with the default being the default location used by the runner setup plan since the primary use case is for a user to run this script on their test runner.

util/metrics/import_metrics_files.rb
billclaytor: My assumption was that it would be run by a user on their test runner using the default ruby for that system. Is there an expectation that it would be run in a different environment where the specific ruby to use would need to be specified? If so, how should that be handled?

RandellP: i just run the script without "ruby" before it.  Which I think does the same as this does

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+
+      `#{cmd}`
+    end
+  end
+end
+
+if $PROGRAM_NAME == __FILE__
+
+  DEFAULT_JSON2GRAPHITE_PATH = File.expand_path "~/git/puppet-metrics-viewer/json2graphite.rb"
+
```
RandellP: Should we allow the prefix to be nil?  make it optional?

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+  DESCRIPTION = <<~DESCRIPTION
+    This script imports metrics collector data for the Standard and Large reference architectures.
```
RandellP: This works for XL as well.  Really it should work for any setup at this point.  The difference from what we had before was that we couldn't handle multi node data... but the comment doesn't need to consider history.  So maybe just drop the "for the Standard and Large reference architectures"

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+    It checks each service subdirectory of the puppet-metrics-collector directory and calls the json2graphite.rb script once for each host subdirectory found within.
+    The server tags are constructed using the following pattern:
+    prefix_id_hostname
```
RandellP: same as above \<prefix>\_\<id>\_\<hostname>

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+
+    For example:
+    be ruby util/metrics/import_metrics.rb -r results/scale/PERF_SCALE_12345 -p slv-649
+
+    This will result in the following server tags:
+    slv-649_12345_127.0.0.1
```
RandellP: doesn't get 127.0.0.1 anymore right... so should make this match a normal hostname

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+    slv-649_12345_ip-10-227-0-22.amz-dev.puppet.net
+    slv-649_12345_ip-10-227-2-173.amz-dev.puppet.net
+
+  DESCRIPTION
+
+  DEFAULTS = <<~DEFAULTS
+
+    The following defaults values are used if the options are not specified:
+    * JSON2GRAPHITE_PATH (-j, --json2graphite): #{DEFAULT_JSON2GRAPHITE_PATH}
```
RandellP: will need update for the default results dir and the nil prefix if added.

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+    end
+  end.parse!
+
+  raise "A results directory must be specified with the -r or --results_dir option" if options[:results_dir].nil?
+  raise "A prefix must be specified with the -p or --prefix option" if options[:prefix].nil?
+
+  raise "Specified directory does not exist: #{options[:results_dir]}" unless File.directory? options[:results_dir]
+
+  json2graphite_path = options[:json2graphite] || DEFAULT_JSON2GRAPHITE_PATH
+  raise "The json2graphite.rb script was not found: #{json2graphite_path}" unless File.exist? json2graphite_path
```
RandellP: if you are going to call it with ruby, maybe should check if ruby is on the path here to avoid getting a sea of errors for each attempt to run the script with ruby.

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+require "optparse"
+
+# This script is intended to be run on a puppet infrastructure node that had the puppet_metrics_collector
+# module installed and running.  Given a time window it will collect the json files generated by the
+# puppet_metrics_collector module and tar them into the tar file.  This includes jsons that are in tar
+# files already.  Those will be extracted and included if they are in the time window.
+
+# General namespace for metrics module
+module Metrics
+  # Main class for collecting metrics files into a tar file
```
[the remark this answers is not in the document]
billclaytor: Oops; good catch! Updated.

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+    # @return [void]
+    #
+    # @example
+    #   import_metrics_files
+    #
+    def import_metrics_files
+      pmc_dir = "#{@results_dir}/puppet-metrics-collector"
+      raise "Directory not found: #{pmc_dir}" unless File.directory? pmc_dir
+
+      puts "Importing puppet_metrics_collector files:"
```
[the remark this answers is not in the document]
billclaytor: Updated

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+
+      service_dirs = Dir.glob("#{pmc_dir}/*").select { |f| File.directory? f }
+
+      puts " service directories:"
+      puts service_dirs
+      puts
+
+      # each service directory
+      service_dirs.each do |service_dir|
+        puts "Checking directory: #{service_dir}"
```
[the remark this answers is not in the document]
billclaytor: It is still going to output the info from the json2graphite script so you would really only see these as they're scrolling by or if you scroll up to review the output. And if you do that these server to break that up so you can see the order of execution. Is there a benefit to hiding this info and making the user specify an option to see it?

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+        puts
+
+        host_dirs.each do |host_dir|
+          import_metrics_files_for_host_dir(host_dir)
+        end
+      end
+    end
+
+    # Calls the json2graphite.rb script for the specified host directory
+    # with a server tag using the following pattern: prefix_id_hostname
```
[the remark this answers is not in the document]
billclaytor: Updated

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+      puts
+
+      # each service directory
+      service_dirs.each do |service_dir|
+        puts "Checking directory: #{service_dir}"
+
+        # each host directory
+        host_dirs = Dir.glob("#{service_dir}/*").select { |f| File.directory? f }
+
+        puts " host directories: "
```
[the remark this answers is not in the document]
billclaytor: Including the array in the string results in the array being printed rather than each element being printed on a separate line. I was not able to quickly find a way around this, so hopefully the following is sufficient:
```
puts " host directories:"; puts host_dirs; puts
```
I did have to disable the rubocop rule for this to be allowed.

util/metrics/import_metrics_files.rb
billclaytor: Updated to "This script imports data captured by puppet-metrics-collector."

util/metrics/import_metrics_files.rb
billclaytor: Updated

util/metrics/import_metrics_files.rb
billclaytor: Updated

util/metrics/import_metrics_files.rb
billclaytor: Updated to remove `ruby` from cmd.

util/metrics/import_metrics_files.rb
billclaytor: Updated to remove `ruby` from cmd.

util/metrics/import_metrics_files.rb
billclaytor: I'm not sure how a default results dir or nil prefix would work. There is a ticket to add an option to optionally suppress the ID portion of the server tag (SLV-685) in case you just want to use the prefix.

util/metrics/import_metrics_files.rb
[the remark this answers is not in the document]
RandellP: It just about doing the right thing, even if the subscript isn't. :) but I can agree since the subscript is already spamming it probably doesn't matter.

util/metrics/import_metrics_files.rb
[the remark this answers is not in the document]
RandellP: odd that the original didn't work, but could also do 
puts " host directories: \n" + host_dirs + "\n\n"
rubocop might like that better.  But doesn't matter much as long as it is only one line. :)

util/metrics/import_metrics_files.rb
RandellP: Should be the same mech as the json...path.  Just set it to ".".  Could get fancy and call Dir.pwd.  Then let the user input override the default.
nil for the prefix could just be setting it to "" or to nil.  And a one line change to not use it when building the tag if nil or "".

util/metrics/import_metrics_files.rb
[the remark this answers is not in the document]
RandellP: just realized you were getting output were each dir got it's own line.  That's rather space heavy...  This would be better...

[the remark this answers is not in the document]
billclaytor: Having each dir get a separate line is intentional. As above, since the json2graphite script outputs the import of each JSON file to a separate line, the only way these are even seen is if the user scrolls up through the output. If they're doing that, I think having each dir on a separate line makes this easier to read than a comma separated list that might wrap depending upon screen width.

util/metrics/import_metrics_files.rb
billclaytor: This seems error-prone and potentially confusing to users. Having the path to the json2graphite script set as a default makes sense since that is the primary use case. I think it is better to be explicit with the results dir to ensure that the user is actually running it against a results dir and that it is the one they mean to run against. The primary use case I see is the user running this script from the root of the GPLT repo and specifying the desired results dir (which is the use case described in the doc):
```
be ruby util/metrics/import_metrics_files.rb -r results/scale/PERF_SCALE_12345 -p slv-666
```

I doubt most users would navigate to the results dir and then try to run the script from there (which would require specifying the full path to the script or a relative path with "`../../` etc...".

Not specifying a prefix would result in a server tag which just contains the ID which does not seem useful (the prefix is what provides context to the run). When combined with the option to suppress the ID portion, having no server tag specified would likely result in an error from the json2graphite script.

RandellP: Having a default for a directory input be the current working dir is really common.  So I don't see much chance of it being confusing.  As for running in the wrong dir, they might run it in the gplt root, but that should bail out pretty fast since there is no puppet-metrics-collector dir there.  And I can't imagine a user being in one perf dir and wanting to run it on another.  But even if they did... it doesn't do any damage, just puts in additional data.  So the cost of the mistake is really low.
I personally run from the perf dir as I usually an looking at other files in there before I upload to influx...  For my first run in tmux session I cut and paste the command path from a helper file, after that I just use the command history and run the command again, just from a different dir... like this:
/home/randell/git/puppet-metrics-viewer/json2graphite.rb --pattern puppet-metrics-collector/**/*.json --convert-to influxdb --netcat localhost --influx-db puppet_metrics --server-tag new_8300_3
I just edit the last arg and off I go. Simple and fast.

On the prefix.  It seems bad to force a person to input a value that they may not have anything meaningful for.  Like say I don't have a meaningful prefix... why force me to add one.  But this is mostly a semantic.  As I personally will always use the prefix...  But it makes me think of all those "unknown_*" files in gplt root... it forced an input when there wasn't a meaningful one... so unknown was given.
That said, you will have to watch out for the case of no prefix and suppressed ID anyway, as the user can input "" as the prefix.  Thought actually I don't think you get an error, it just adds more data to that tag... and I think overwrites existing data if the timestamps line up...  that or ignores if there is already data of the same timestamp...  so maybe you don't have to care about it really.

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,181 @@
...
+      hostname = File.basename host_dir
+      server_tag = "#{@prefix}_#{@id}_#{hostname}"
+      pattern = "'#{host_dir}/*.json'"
+      cmd = "ruby #{@json2graphite_path} --pattern #{pattern}" \
+        " --convert-to influxdb --netcat localhost --influx-db puppet_metrics --server-tag #{server_tag}"
+      puts "Importing puppet-metrics-collector files for host: #{hostname}"
+      puts " cmd: #{cmd}"
+      puts
+
+      `#{cmd}`
```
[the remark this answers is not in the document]
RandellP: bump, just commenting in case this one got missed...

util/metrics/import_metrics_files.rb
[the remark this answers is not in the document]
RandellP: yeah, it's a small detail, since this should usually be one dir, except for compilers which might be a few.  As you can tell I have a thing against spammy output, as it usually means things just never get seen.  But as we already discussed, this output is so spammy already due to json2graphite no one is every going to read the output unless they are looking for something specific.  So doesn't matter as much here.

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,247 @@
...
+# General namespace for metrics module
+module Metrics
+  # Main class for importing puppet-metrics-collector files via the json2graphite script
+  #
+  # @author Bill Claytor
+  #
+  # @attr [string] results_dir The GPLT results directory to process
+  # @attr [string] prefix The prefix to use when building the server tag
+  # @attr [string] json2graphite_path The path to the json2graphite.rb script
+  # @attr [string] id The ID to use when building the server tag (the last segment of the results folder name)
```
RandellP: need to add hostname to this list

util/metrics/import_metrics_files.rb
```diff
@@ -0,0 +1,247 @@
...
+
+      service_dirs.each do |service_dir|
+        import_metrics_files_for_service_dir(service_dir)
+      end
+    end
+
+    # Returns true if the specified dir is a GPLT results dir, otherwise false
+    #
+    # @author Bill Claytor
+    #
```
RandellP: missing param comment.

util/metrics/import_metrics_files.rb
```diff
@@ -63,6 +65,19 @@ def import_metrics_files
...
+    # @return [Boolean] true if the specified dir is a GPLT results dir, otherwise false
+    #
+    # @example
+    #   result = valid_results_dir?(results_dir)
+    #
+    def valid_results_dir?(results_dir)
+      results_dir.include?("PERF_")
```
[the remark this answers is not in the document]
billclaytor: Updated to use the following since `results_dir` is the full path and it also needs to work for scale results:
```
File.basename(results_dir).match?(/^PERF_(SCALE_)?\d+$/)
```

util/metrics/import_metrics_files.rb
billclaytor: Updated

util/metrics/import_metrics_files.rb
billclaytor: Updated

util/metrics/import_metrics_files.rb
billclaytor: Updated the spec tests

util/metrics/import_metrics_files.rb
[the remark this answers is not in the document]
billclaytor: Good point! Updated the script to handle a non-zero exit code and added a spec test.

PR #391, opened 2020-01-15
(SLV-747) retry abs provision on exception

RandellP:
current retry loop only retries on bad return data.  But according to Brandon High Net::HTTP will actually raise an exception on most non-200/300 responses from the endpoint, and thus it won't do the desired retry
Added code to retry on the specific error we are seeing.

### DIFF
--- a/setup/helpers/abs_helper.rb
+++ b/setup/helpers/abs_helper.rb
@@ -214,7 +214,16 @@ module AbsHelper
     result = nil
 
     (1..ABS_MAX_REQUEST_ATTEMPTS).each do |ct|
-      response = perform_awsdirect_request(uri, request_body)
+      begin
+        response = perform_awsdirect_request(uri, request_body)
+      rescue => e # rubocop:disable Style/RescueStandardError
+        # ABS raises some exceptions that we believe it is safe to rety on
+        # Bandon High thinks this is one.
+        # QENG-7458 is what this is supposed to fix.
+        raise unless e.message =~ "Received following error trying to provision request"
+
+        puts "Caught special exception due to abs response so retry can proceed"
+      end
       if valid_abs_response?(response)
         result = parse_awsdirect_response_body(response.body)
         break


setup/helpers/abs_helper.rb
```diff
@@ -214,7 +214,16 @@ def get_abs_resource_host(host_to_request)
...
 
     (1..ABS_MAX_REQUEST_ATTEMPTS).each do |ct|
-      response = perform_awsdirect_request(uri, request_body)
+      begin
+        response = perform_awsdirect_request(uri, request_body)
+      rescue => e # rubocop:disable Style/RescueStandardError
+        # ABS raises some exceptions that we believe it is safe to rety on
+        # Bandon High thinks this is one.
+        # QENG-7458 is what this is supposed to fix.
+        raise unless e.message =~ "Received following error trying to provision request"
```
johnduarte: Since the HTTP error code is included in the title and body of the response from Net::HTTP would it be better to search for the expected error code (404) as per the [comment in QENG-7458](https://tickets.puppetlabs.com/browse/QENG-7458?focusedCommentId=712038&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-712038)?
```
raise unless e.message =~ "Error 404"
```

highb: You really don't want to catch all exceptions. 😄 https://thoughtbot.com/blog/rescue-standarderror-not-exception

To kill both birds with one stone, you can rescue 404 exceptions specifically: `rescue Net::HTTPNotFound => e`

highb: Well, you'd still want to catch the specific error text from ABS, but at least you could be confident that it's coming from a 404 and not, say, hitting Ctrl-C or making a syntax error.

RandellP: So... when i looked at the output... I didn't see an exception name... so I went with general.  Digging deeper I am seeing this code... https://github.com/puppetlabs/gatling-puppet-load-test/blob/99e20c3a012f1e2655484a417703e5d2ee0b3884/setup/helpers/abs_helper.rb#L489
That is in the output noted in the ticket.  I don't think we are getting an unhandled exception here...  and based on that, the retry code "should" be retrying...   @johnduarte @billclaytor  it was something one of you said in a standup I think that cause me to update the ticket about this issue... maybe I assumed this was the issue but the failure to provision was something else?

RandellP: Bill confirmed we saw this error recently in CI.  It could have happened 3 times in a row and defeated our attempt to get around it with retry though...

highb: Are there some logs we could look at to verify that?

RandellP: Not sure how to find them.  In theory grepping the console output files from our old CI runs... but I don't know how to do that (or if you even can).  And they may have been deleted by now depending on how fast they get removed.

RandellP:
@highb @johnduarte  I am going to close this PR as I am like 98% sure an exception is not being thrown based on the output I pasted into the QENG ticket.  I will update the ticket.

PR #405, opened 2020-01-27
(SLV-738) Ensure puppet agent disabled on infra

johnduarte:
This commit adds a pre-suite step after the final puppet run to ensure
that the puppet agent service is disabled on all hosts prior to any
gatling tests.

### DIFF
--- a/setup/install_gatling/99_final/99_final_puppet_run.rb
+++ b/setup/install_gatling/99_final/99_final_puppet_run.rb
@@ -8,8 +8,9 @@ test_name "final puppet run" do
     # If deployed via install_pe this is part of the install.
     #   Ref: https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/install/pe_utils.rb
     # This is added to ensure that the puppet agent service is
-    # disabled prior to a gatling run regardless of how the hosts
-    # were deployed.
-    stop_agent_on(hosts)
+    # disabled on hosts with a role other than agent prior to a
+    # gatling run regardless of how the hosts were deployed.
+    infra_hosts = hosts.reject { |host| (host[:roles] == ["agent"]) }
+    stop_agent_on(infra_hosts)
   end
 end
--- a/setup/install_gatling/99_final/99_final_puppet_run.rb
+++ b/setup/install_gatling/99_final/99_final_puppet_run.rb
@@ -4,4 +4,12 @@ test_name "final puppet run" do
   step "Run puppet until all changes are applied" do
     run_agent_until_no_change(hosts)
   end
+  step "Ensure puppet agent disabled" do
+    # If deployed via install_pe this is part of the install.
+    #   Ref: https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/install/pe_utils.rb
+    # This is added to ensure that the puppet agent service is
+    # disabled prior to a gatling run regardless of how the hosts
+    # were deployed.
+    stop_agent_on(hosts)
+  end
 end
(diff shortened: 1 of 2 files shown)


setup/install_gatling/99_final/99_final_puppet_run.rb
```diff
@@ -4,4 +4,12 @@
   step "Run puppet until all changes are applied" do
     run_agent_until_no_change(hosts)
   end
+  step "Ensure puppet agent disabled" do
+    # If deployed via install_pe this is part of the install.
+    #   Ref: https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/install/pe_utils.rb
+    # This is added to ensure that the puppet agent service is
+    # disabled prior to a gatling run regardless of how the hosts
+    # were deployed.
+    stop_agent_on(hosts)
```
RandellP: So what if we create an actual agent?  We have a rake task for that.  Should this be more selective?  Not even sure there is a good way to do that... so just asking the question really. :)

johnduarte: Answer: Confine the step to hosts that do not include any of the PE Infra beaker roles.
```
pe_infra_roles = ["master", "database", "dashboard", "compile_master", "metric"]
confine :to, {}, hosts.select { |host| ! (host & pe_infra_roles).any? }
```

johnduarte: Super beaker

RandellP: looks good.  go for it.

johnduarte: This has been added and verified.  I had the inclusion logic inverted in my comment above.  This has been corrected in the code.

setup/install_gatling/99_final/99_final_puppet_run.rb
```diff
@@ -4,4 +4,14 @@
...
   end
+  step "Ensure puppet agent disabled" do
+    # If deployed via install_pe this is part of the install.
+    #   Ref: https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/install/pe_utils.rb
+    # This is added to ensure that the puppet agent service is
+    # disabled prior to a gatling run regardless of how the hosts
+    # were deployed.
+    pe_infra_roles = %w[master database dashboard loadbalancer compile_master metric]
+    confine :to, {}, (hosts.select { |host| (host.host_hash[:roles] & pe_infra_roles).any? })
+    stop_agent_on(hosts)
```
RandellP: This doesn't seem right.  hosts is from beaker right...  if the confine changed hosts that is bad.  and if it didn't, then you are still running stop_agent_on all hosts...

RandellP: Also, I tried to replicate just the select statement, and failed...
2.5.1 :054 > foo
 => {"one"=>{"val"=>"1"}, "two"=>{"val"=>"2"}, "three"=>{"val"=>"3"}}
2.5.1 :055 > nums
 => ["1", "3"]
2.5.1 :056 > foo.select { |obj| (foo[obj]["val"] & nums)}
Traceback (most recent call last):
        4: from /Users/randell/.rvm/rubies/ruby-2.5.1/bin/irb:11:in `<main>'
        3: from (irb):56
        2: from (irb):56:in `select'
        1: from (irb):56:in `block in irb_binding'
NoMethodError (undefined method `&' for "1":String)

johnduarte: The confine facility restricts the hosts object in the context of the test in which it is run.
https://github.com/puppetlabs/beaker/blob/master/docs/how_to/confine.md

RandellP: And lastly.  It's taken me half an hour to try and prove this line could work.  That probably means it is too complex.  Could be I am just weak on ruby, but you're stuck with me. :)

RandellP: alright.  turns out this is a beaker thing that I am just not familiar with.  And I was able to prove to myself that it should work.

johnduarte: Updated to stop agent on any host that is not only an agent.

PR #412, opened 2020-02-04
(SLV-728) Update the scale summary script to include failing iterations with standard deviation

billclaytor:
The `util/report_utils/build_scale_csv_summary.rb` has been updated to include failing iterations with standard deviation for each column:
- replace 'average' with 'mean'
- calculate and include the standard deviation for each column
- include a table with the failing iteration

### DIFF
--- a/util/report_utils/build_scale_csv_summary.rb
+++ b/util/report_utils/build_scale_csv_summary.rb
@@ -204,7 +204,6 @@ end
 #
 def update_summary_csv(summary_csv_path, line)
   puts "updating summary csv: #{summary_csv_path}"
-  puts "results: #{line}"
 
   # the results row is a string rather than an array so handle as file rather than csv
   File.open(summary_csv_path, "a+") do |f|
--- a/util/report_utils/build_scale_csv_summary.rb
+++ b/util/report_utils/build_scale_csv_summary.rb
@@ -3,10 +3,6 @@
 require "../../tests/helpers/perf_results_helper.rb"
 include PerfResultsHelper # rubocop:disable Style/MixinUsage
 
-raise Exception, "you must provide a results directory" unless ARGV[0]
-
-parent_dir = ARGV[0]
-
 TEMPLATE_PATH = "./templates/scale_csv_summary_template.html"
 
 SUMMARY_CSV_HEADINGS = ["result",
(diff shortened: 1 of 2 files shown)


util/report_utils/build_scale_csv_summary.rb
```diff
@@ -24,8 +20,134 @@
...
+# note: the template can be specified via the TEMPLATE_PATH environment variable
+#
+# @author Bill Claytor
+#
+# @param [String] parent_dir The parent directory containing the scale test result directories
+#
+# @return [void]
+#
+# @example
+#   build_report(parent_dir)
```
johnduarte: does not match method name.

util/report_utils/build_scale_csv_summary.rb
```diff
@@ -24,8 +20,134 @@
...
+# @author Bill Claytor
+#
+# @param [String] csv_path The scale test results CSV file path
+#
+# @return [String] The CSV data converted to an HTML table
+#
+# @example
+#    table = process_results_csv(csv_path)
+#
+def process_results_csv(csv_path)
```
johnduarte: nitpick: You have one `process_*` method that has a return value (process_result_csv) and one that has no return value (process_result_dirs).  This UX is inconsistent.  Consider altering the naming convention for these methods.

util/report_utils/build_scale_csv_summary.rb
billclaytor: Updated

util/report_utils/build_scale_csv_summary.rb
billclaytor: Renamed `process_result_csv` to `extract_table_from_results_csv`.

util/report_utils/build_scale_csv_summary.rb
```diff
@@ -24,8 +20,134 @@
...
+# @return [void]
+#
+# @example
+#   build_report(parent_dir)
+#
+def build_scale_csv_summary(parent_dir)
+  puts "building scale test csv summary report for parent dir: #{parent_dir}"
+  puts
+
+  parent_name = File.basename(parent_dir)
```
billclaytor: This can result in awkward report names; updated to use a generic name.

